PR #20, opened 2018-02-13
Added SupervisedTimeSeriesKMeans(TimeSeriesKMeans)

jscheithe:
Hi Romain,

I wrote this Wrapper for TimeSeriesKMeans for the (very simple) supervised case of TimeSeriesKMeans and wanted to share.

I moved some of its assignments to SupervisedTimeSeriesKMeans.fit(X, y) to make the Estimator more compliant to the scikit-learn guidelines (it "almost" passes sklearn.utils.estimator_checks.check_estimator()).

I didn't want to write tests yet, as you seem to migrate to pytest currently.

Let me know what you think!

### DIFF
--- a/tslearn/clustering.py
+++ b/tslearn/clustering.py
@@ -571,6 +571,85 @@ class TimeSeriesKMeans(BaseEstimator, ClusterMixin, TimeSeriesCentroidBasedClust
         X_ = to_time_series_dataset(X)
         return self._assign(X_, update_class_attributes=False)
 
+class SupervisedTimeSeriesKMeans(TimeSeriesKMeans):
+    """Supervised K-means clustering for time-series data.
+
+    Parameters
+    ----------
+    metric : {"euclidean", "dtw", "softdtw"} (default: "euclidean")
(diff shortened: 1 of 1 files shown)


rtavenar:
Hi, do you have a reference for this method? I had never heard of it before.

jscheithe:
Actually, I don't..

My naive understanding of what supervised k-means is supposed to mean is this:

When we have class labels available, we don't have to look for "good" centroids, that yield "good" clusters. We can simply take the clusters that the class labels give us and calculate the centroids as the mean/barycenter/.. of each class.
Prediction is done as usual.

Maybe some reference that this works (better than just forgetting about the labels) would be good, I agree.

rtavenar:
Then, if this does not correspond to some published method, I would suggest we do not merge this pull request but, if you are interested, you could have a pull request to [tslearn-addons](https://github.com/rtavenar/tslearn_addons) as the requirements for integration there are less strict. Let me know if this option suits you.

Best,
Romain

jscheithe:
Sure, that sounds good! I just wanted to make sure you can use my code if you want to.

Should i create an empty file there with my class that imports from `tslearn.clustering`?

rtavenar:
Yep!

PR #25, opened 2018-03-06
Bundle the license file in distributed .tar.gz

rth:
This PR makes sure that the LICENSE file is included in .tar.gz distributed on PyPi. Approach based on this [SO question](https://stackoverflow.com/questions/9977889/how-to-include-license-file-in-setup-py-script), the alternative would be to write a `MANIFEST.in` file.

The corresponding issues was raised while packaging tslearn in conda-forge (see https://github.com/conda-forge/staged-recipes/pull/5275).

**Note:** I tested that this does include the `LICENSE` in the produced .tar.gz when running `python setup.py sdist`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ setup(
     include_dirs=[numpy.get_include()],
     packages=['tslearn'],
     package_data={"tslearn": [".cached_datasets/Trace.npz"]},
+    data_files = [("", ["LICENSE"])],
     install_requires=['numpy', 'scipy', 'scikit-learn', 'Cython'],
     ext_modules=ext,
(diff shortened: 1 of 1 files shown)


rth:
Travis CI is unhappy, bit it also currently fails on master ...

rtavenar:
Yes, the Travis CI fail is not related to this PR, definitely. Thanks for this PR, and for making tslearn available to conda users !

rth:
FYI, tslearn can now be installed from conda-forge with,
```
conda install -c conda-forge tslearn
```
this does not require Cython or a C compiler, which should make it easier to install tslearn, in particular, on Windows.

Issue #48, opened 2018-06-15
TimeSeriesScalerMinMax need a transform method


StevenLOL:
Hi,

Like MinMaxScaler in scikitlearn , a transform method is needed for TimeSeriesScalerMinMax.

rtavenar:
Hi,

In practice, the reason why we have not implemented this feature in `tslearn` is exactly so that users do not misunderstand what `TimeSeriesScalerMinMax` does: it does not scale each feature (across the dataset) to lie in [0, 1]. It scales each time series so that its features lie in [0, 1].

jamesmyatt:
The classes in preprocessing are not scikit-learn compatible [transformers](https://scikit-learn.org/stable/data_transforms.html) since they don't implement `fit` and `transform` as separate methods.

rtavenar:
@GillesVandewiele is there a reason why, in the preprocessing module, some objects have both fit and transform methods even if fit does nothing and others only have fit_transform ?

GillesVandewiele:
I'm not entirely sure, but my tests may skip those classes at this moment (as they do not inherit from `Estimator`).

rtavenar:
This is done for all transformers in preprocessing now. Waiting for Travis, if things pass, I'll close this issue.

PR #56, opened 2018-07-30
Add a simple method to inverse the transformation.

benbenbang:
Similar to all the transformation in scikit-learn, we also need a inverse-transform in tslearn to help us to know the original time series, especially after the prediction.

It will be similar to sklearn to do:
``` python
tssm = TimeSeriesScalerMinMax()
X_train_tf = tssm.fit_transform(X_train)

# After the prediction
preds = tssm.inverse_transform(preds)
```

### DIFF
--- a/tslearn/preprocessing.py
+++ b/tslearn/preprocessing.py
@@ -82,6 +82,7 @@ class TimeSeriesScalerMinMax(TransformerMixin):
     def __init__(self, min=0., max=1.):
         self.min_ = min
         self.max_ = max
+        self.params = []
 
     def fit_transform(self, X, **kwargs):
         """Fit to data, then transform it.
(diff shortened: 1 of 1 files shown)


rtavenar:
Hi @benbenbang ,

I guess the only case where that would make sense is when the task is prediction of next value, right? I'm afraid this would lead to misuse by users not aware of what these transformers do (per-time series rescaling, not dataset-wise rescaling). Would you see another way to do this without this risk of misuse?

Best,
Romain

benbenbang:
Hi @rtavenar, pardon, I didn't get what you mean. It I understand well, the code in preprocessing is for an `array-like` data format which is neither dataframe or matrix. So when people create a new object to do the transformation, it should be only correspond to the time series they applied to. So what do you mean by dataset-wise rescaling? Would you please show me some concrete example?
Thank you.

GillesVandewiele:
In sklearn, the normalization happens dataset-wise. E.g. for the MinMaxScaler, they would calculate the minimum and maximum column-wise on the entire dataset (take min and max for each possible variable). In tslearn, the scaling happens per time series, this would mean that you would have to store the min and max for each timeseries in order to apply the inverse transform, and that the order of the time series that get passed to both `transform` and `inverse_transform` may not change.

rtavenar:
NB: related to #48 btw

benbenbang:
Got it, I will check if there's a better way to do so. Thanks.

GillesVandewiele:
@rtavenar is there by the way a reason why we should do timeseries-wise scaling instead of dataset-wise? What's the issue with calculating a minimum and maximum based on the entire dataset and scaling all time series based on those values?

jamesmyatt:
The classes in preprocessing are not scikit-learn compatible [transformers](https://scikit-learn.org/stable/data_transforms.html) since they don't implement `fit` and `transform` as separate methods. Hence there's no reason to expect an `inverse_transform` method.

Issue #58, opened 2018-08-09
KShape first guess guidance / constrain


gandroz:
For the KShape algorithm (or k-means) it could be interesting to guide the algo to converge with the user knowledge. Imagine the user knows that there are three clusters and has at least one sample of each class, it could be interesting to use these samples as a first barycenter during the learning phase.

rtavenar:
I never closed that issue, but this is actually implemented in tslearn at the moment (v0.2)

Issue #52, opened 2018-07-13
`TimeSeriesScalerMeanVariance` got warning of dividing zero when the time series is 0


DongminWu:
Hi, 

Thanks for your brilliant works of building this library!

I am trying to use `tslearn.preprocessing.TimeSeriesScalerMeanVariance` for standardization, but it failed when one of my data samples is pure `0`. Here is the example:

```
t = np.array([[0,0,0],[1,2,3]])
TimeSeriesScalerMeanVariance().fit_transform(t)
```

the output for the above code is:

```
/lib/python3.6/site-packages/tslearn/preprocessing.py:153: RuntimeWarning: invalid value encountered in true_divide
  X_[i, :, d] = (X_[i, :, d] - cur_mean) * self.std_ / cur_std + self.mu_
array([[[        nan],
        [        nan],
        [        nan]],

       [[-1.22474487],
        [ 0.        ],
        [ 1.22474487]]])
```

The issue is the standard deviation of a `[0,0,0]` is 0, it there any good methods for working around?

rtavenar:
Hi and thanks for this bug report.

I guess for this, the best idea would be to take inspiration from `sklearn` (cf [here](https://github.com/scikit-learn/scikit-learn/blob/a24c8b464d094d2c468a16ea9f8bf8d42d949f84/sklearn/preprocessing/data.py#L61))

This means setting variance to 1. (arbitrarily) as soon as it is zero.

I'll try to do it asap. Do not hesitate to propose a PR on this if you have time before I do :)

rtavenar:
OK, It was pretty straight-forward, so I fixed it directly.

The bug should not exist anymore starting from version `tslearn-0.1.18.4`

GillesVandewiele:
Just when I was about to contribute a bit!! You're too fast @rtavenar ..

I ported the code to numpy though, I expect it to be faster.

```
X_ = to_time_series_dataset(X)
means, stds = np.mean(X_, axis=1), np.std(X_, axis=1)
stds[stds == 0] = 1
return (X_ - means) * self.std_ / stds + self.mu_
```

GillesVandewiele:
Here's a few stats:

`TimeSeriesScalerMeanVariance(mu=0., std=1.).fit_transform([np.random.rand(10000000)])` takes 0.6294596195220947 seconds for the numpy implementation and  0.646794319152832 with the original implementation. 

For vectors of length 100000000 the numpy implementation is even slower... Strange stuff...

[...]

PR #59, opened 2018-08-09
Add initial guess as centroid

gandroz:
According to the issue #58 , here a proposal to improve clustering (only for the KShape method for now) by letting the user choose an initial guess as centroids. This guess is a numpy array of int which are the indices of the samples to be used as centroids instead of a random vector.

### DIFF
--- a/tslearn/clustering.py
+++ b/tslearn/clustering.py
@@ -701,7 +701,7 @@ class KShape(BaseEstimator, ClusterMixin, TimeSeriesCentroidBasedClusteringMixin
         if mean_shift > 0:
             mu_k[mean_shift:] = vec[:-mean_shift, -1].reshape((sz - mean_shift, 1))
         elif mean_shift < 0:
-            mu_k[:mean_shift] = vec[-mean_shift:, -1].reshape((sz + mean_shift, 1)) 
+            mu_k[:mean_shift] = vec[-mean_shift:, -1].reshape((sz + mean_shift, 1))
         else:
             mu_k = vec[:, -1].reshape((sz, 1))
 
@@ -732,12 +732,15 @@ class KShape(BaseEstimator, ClusterMixin, TimeSeriesCentroidBasedClusteringMixin
 
     def _fit_one_init(self, X, rs):
         n_samples, sz, d = X.shape
-        self.labels_ = rs.randint(self.n_clusters, size=n_samples)
         if hasattr(self.init, '__array__'):
             self.cluster_centers_ = self.init.copy()
         else:
             self.cluster_centers_ = rs.randn(self.n_clusters, sz, d)
         self._norms_centroids = numpy.linalg.norm(self.cluster_centers_, axis=(1, 2))
+        if hasattr(self.init, '__array__'):
+            self._assign(X)
+        else:
+            self.labels_ = rs.randint(self.n_clusters, size=n_samples)
         old_inertia = numpy.inf
 
         for it in range(self.max_iter):
--- a/tslearn/clustering.py
+++ b/tslearn/clustering.py
@@ -470,11 +470,14 @@ class TimeSeriesKMeans(BaseEstimator, ClusterMixin, TimeSeriesCentroidBasedClust
             metric_params = {}
         self.gamma_sdtw = metric_params.get("gamma_sdtw", 1.)
 
-    def _fit_one_init(self, X, x_squared_norms, rs):
+    def _fit_one_init(self, X, x_squared_norms, initial_centroids, rs):
         n_ts, _, d = X.shape
         sz = min([ts_size(ts) for ts in X])
-        self.cluster_centers_ = _k_init(X[:, :sz, :].reshape((n_ts, -1)),
-                                        self.n_clusters, x_squared_norms, rs).reshape((-1, sz, d))
+        if initial_centroids is not None:
+            self.cluster_centers_ = initial_centroids.copy()
+        else:
+            self.cluster_centers_ = _k_init(X[:, :sz, :].reshape((n_ts, -1)),
+                                            self.n_clusters, x_squared_norms, rs).reshape((-1, sz, d))
         old_inertia = numpy.inf
 
         for it in range(self.max_iter):
@@ -526,18 +529,27 @@ class TimeSeriesKMeans(BaseEstimator, ClusterMixin, TimeSeriesCentroidBasedClust
             else:
                 self.cluster_centers_[k] = EuclideanBarycenter().fit(X[self.labels_ == k])
 
-    def fit(self, X, y=None):
+    def fit(self, X, initial_centroids=None, y=None):
         """Compute k-means clustering.
 
         Parameters
         ----------
         X : array-like of shape=(n_ts, sz, d)
             Time series dataset.
+        initial_centroids: array-like of shape=(n_clusters, sz, d) of initial centroids
+            Time series dataset.
         """
         X_ = to_time_series_dataset(X)
         rs = check_random_state(self.random_state)
         x_squared_norms = cdist(X_.reshape((X_.shape[0], -1)), numpy.zeros((1, X_.shape[1] * X_.shape[2])),
                                 metric="sqeuclidean").reshape((1, -1))
+
+        if initial_centroids is not None:
+            assert initial_centroids.shape[-1] == 1, "kMeans is supposed to work on monomodal data, " \
+                                                     "provided data has dimension {}".format(X_.shape[-1])
+            assert initial_centroids.shape[0] == self.n_clusters, "Initial guess index array must contain {} samples," \
+                                                                  " {} given".format(self.n_clusters,
+                                                                                     initial_centroids.shape[0])
 
         best_correct_centroids = None
         min_inertia = numpy.inf
@@ -548,7 +560,7 @@ class TimeSeriesKMeans(BaseEstimator, ClusterMixin, TimeSeriesCentroidBasedClust
                 if self.verbose and self.n_init > 1:
                     print("Init %d" % (n_successful + 1))
                 n_attempts += 1
-                self._fit_one_init(X_, x_squared_norms, rs)
+                self._fit_one_init(X_, x_squared_norms, initial_centroids, rs)
                 if self.inertia_ < min_inertia:
                     best_correct_centroids = self.cluster_centers_.copy()
                     min_inertia = self.inertia_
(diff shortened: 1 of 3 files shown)


GillesVandewiele:
Hello, great work! Small suggestion: wouldn't it be better to let the user provide initial centroids, instead of indices? That way, you are not bound to have the initial centroids set to time series that are WITHIN your dataset. The suggested approach works well for your specific use case, where you know a sample of each cluster, but imagine if you would know more than 1 sample, then you could calculate a centroid of these samples and provide it to the algorithm.

gandroz:
You're right, it is just a "quick win" I implemented on my side. Your suggestion is interesting and that was my first though, but I did not want to deal with time series of different lengths for example, and normalization would have been a bit more difficult as we have to normalize the initial guess the same manner the X_train is normalized, but not too difficult.

I could check that on my side if you want ?

GillesVandewiele:
Does the algorithm currently support time series of variable length? I do not really know the KShape algorithm that well... Does not seem to give an error when I try it in my Python shell with a simple example.

Indeed, I did not think of the normalization, but it should be easy to fix if we decouple the line on https://github.com/rtavenar/tslearn/pull/59/commits/f4db8cdcc33256cf2c5d676e36adde97861dd66d#diff-79bececde0dc56f7672e4cedbaa36832R750 such that we have the `TimeSeriesScalerMeanVariance` object, then we can access the required data to normalize the provided initial centroid as wel.

GillesVandewiele:
And on another note: maybe a few asserts should be added, such that a fitting exception can be raised? One that comes to mind straight away is that the number of `initial centroids` must be equal to `n_clusters`. But maybe other assertions concerning the length of these centroids can be made as well?

gandroz:
Normalization is not a big deal indeed. I'll fix that. And yes, KShape support time series of variable length as it uses the method `to_time_serries_dataset` and not `to_time_series`.
There is already an assertion on the length of the initial guess at [line 753 ](https://github.com/rtavenar/tslearn/commit/f4db8cdcc33256cf2c5d676e36adde97861dd66d#diff-79bececde0dc56f7672e4cedbaa36832R754)

And a question for you, the tests failed on Travis. Is there a way I can run them before pushing ?

GillesVandewiele:
My apologies, I looked over that! It seems like the doctest of clustering.py is failing...

You can check the logs (https://travis-ci.org/rtavenar/tslearn)[here] and to run the doctest, just run `python3 -m doctest -v clustering.py` in the `tslearn` directory.

gandroz:
Hi,
I've just pushed some changes as you suggested. You can now use an initial centroids array instead of an array of indices. I tested it on my case and it works as expected. 
I also modified the preprocessing code by vectorizing it more so that it is quickier now.

GillesVandewiele:
Great work mate! Seems like the doctests are passing as well. Looks like a great addition to me!

gandroz:
you welcome, it is really a nice package btw !

GillesVandewiele:
All props to @rtavenar. I haven't contributed yet to this repository. But I completely agree, this package is great...

Hmmm Travis seems to have failed, while it was successful for python 2.7 :thinking: Do the doctests work locally for you?

gandroz:
yes it does, but I use p27. Seem like it is a build issue 

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

GillesVandewiele:
Ye but it's strange that it takes longer than 10 minutes... Other builds of python 3 mostly take around 5 minutes or something... Do you have python3 installed? Could you run a doctest with python3 possibly?

On the other hand, the build of python 3.5 just went terrible, it couldn't even install python 3.5

gandroz:
The issue is really at the build time of the docker image. I could try the doctest with python 3.5 but the test does not event go that far

GillesVandewiele:
Ye exactly, something really weird. I'm kinda clueless as well :/. In the other builds, it did seem to be executing the nosetests though.

```
$ nosetests $KERAS_IGNORE --with-doctest --with-coverage --cover-erase --cover-package=tslearn
Using TensorFlow backend.
...........
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

I think those dots correspond to successful tests.

gandroz:
The tests passed for python 3.5, but there is a strange behaviour. Two tests are said `fail` but the result is the same as expected:

>  Expected:
    array([[[ 1. ],
            [ 1.5],
            [ 2. ]]])
Got:
    array([[[1. ],
            [1.5],
            [2. ]]])

The difference is just a white space before numbers, not the values....

tslearn/preprocessing.py
```diff
@@ -131,10 +130,12 @@ class TimeSeriesScalerMeanVariance(TransformerMixin):
     def __init__(self, mu=0., std=1.):
         self.mu_ = mu
         self.std_ = std
+        self.global_mean = None
+        self.global_std = None
 
-    def fit_transform(self, X, **kwargs):
+    def fit_transform(self, X, global_stats=False, **kwargs):
```
rtavenar: Hum, `global_stats` argument is never used afterwards, is it?

If not, you should remove it, I guess.

tslearn/clustering.py
```diff
@@ -731,20 +734,27 @@ def _fit_one_init(self, X, rs):
...
         ----------
         X : array-like of shape=(n_ts, sz, d)
             Time series dataset.
+        initial_centroids: array-like of shape=(n_clusters, sz, d) of initial centroids
+            Time series dataset.
         """
 
         X_ = to_time_series_dataset(X)
         self._norms = numpy.linalg.norm(X_, axis=(1, 2))
-        X_ = TimeSeriesScalerMeanVariance(mu=0., std=1.).fit_transform(X_)
```
rtavenar: Why is this line removed?

tslearn/clustering.py
```diff
@@ -704,10 +704,13 @@ def _assign(self, X):
         _check_no_empty_cluster(self.labels_, self.n_clusters)
         self.inertia_ = _compute_inertia(dists, self.labels_)
 
-    def _fit_one_init(self, X, rs):
+    def _fit_one_init(self, X, initial_centroids, rs):
```
rtavenar: Would it be possible to rename this argument as `init` and more or less match sklearn usage (cf. https://github.com/scikit-learn/scikit-learn/blob/f0ab589f/sklearn/cluster/k_means_.py#L707)?

tslearn/preprocessing.py
```diff
@@ -146,11 +147,10 @@ def fit_transform(self, X, **kwargs):
...
         """
         X_ = to_time_series_dataset(X)
-        for i in range(X_.shape[0]):
-            for d in range(X_.shape[2]):
-                cur_mean = X_[i, :, d].mean()
-                cur_std = X_[i, :, d].std()
-                if cur_std == 0.:
-                    cur_std = 1.
-                X_[i, :, d] = (X_[i, :, d] - cur_mean) * self.std_ / cur_std + self.mu_
+        mean_t = numpy.mean(X_, axis=1)[:, numpy.newaxis, :]
```
GillesVandewiele: I tried converting the code to numpy in the past and discovered that it actually ran slower than a Pure-Python loop. Maybe this should be tested more extensively, and if it indeed runs slower, convert back to the old code

rtavenar: @GillesVandewiele : is it still slower for datasets with many time series? Because it definitely looks like a better solution than the original one in theory :)

tslearn/clustering.py
GillesVandewiele: Seems like the code has changed to 

`self._norms = numpy.linalg.norm(X_, axis=(1, 2))` and `self._norms_centroids = numpy.linalg.norm(self.cluster_centers_, axis=(1, 2))`. I would also use the Normalizers from tslearn instead of using the numpy norm function...

tslearn/preprocessing.py
GillesVandewiele: i think we actually made a reasoning mistake and thought the normalization happened dataset-wise, I think the global_stats and `self.global_mean` + `self.global_std` stem from that. Of course, this becomes unnecessary since the normalization happens timeseries-wise. We just have to use the `transform` method of the SAME normalizer on both the provided X input array as the centroids. 

Important to note in the documentation is that the centroids may not be normalized by the user himself.

tslearn/preprocessing.py
GillesVandewiele: I mentioned it in Issue #52 right before getting closed, so maybe you looked over it. My code was not exactly the same as this:

```
X_ = to_time_series_dataset(X)
means, stds = np.mean(X_, axis=1), np.std(X_, axis=1)
stds[stds == 0] = 1
return (X_ - means) * self.std_ / stds + self.mu_
```

This seemed to run slower when I did some simple tests, but maybe should be double checked. It is definitely cleaner code!

GillesVandewiele: This conversion to numpy can btw also easily be done for the MinMaxScaler!

rtavenar: Yep, but in Issue #52, you tested on a single long time series, which did not benefited from removing the loop. So hopefully with many time series that would be better!

Test:
```
time_series = numpy.random.randn(100000, 100, 10)
t0 = time.time()
TimeSeriesScalerMinMax().fit_transform(time_series)
t1 = time.time()
TimeSeriesScalerMinMaxTuned().fit_transform(time_series)
t2 = time.time()
```

With `TimeSeriesScalerMinMaxTuned` being the proposed implementation.

Gives:
```
12.474971771240234 5.250811815261841
```

So, good idea :)

rtavenar:
The good news is: tests seem to have failed because of an issue on Travis CI, now they pass for all configs!

tslearn/preprocessing.py
GillesVandewiele: You are right! Mistake from my side! I benched it myself now as well :)  

numpy_times

python_times

gandroz:
Hi sorry for the delay (i'm in vacation..). The numpy and vectorized
approach is definitely more efficient than the loop approach. It ran 10
times quickier in my tests so I think it worses the change.

Glad to ear the failed tests were due to Travis error.


Le ven. 17 août 2018 09:42, Gilles Vandewiele <notifications@github.com> a
écrit :

gandroz:
Before adding the initial guess to the `TimeSeriesKMeans`, we tweaked the `KShape` method to shift back the centroid after computation of the eigen vectors to avoid cumulative shift fo the centroid and allow further optimization of the centroids. The visual result is that all centroids are better aligned to the samples. The performance result I suspect (not fully verified) is that because the centroids are back shifted, the cross-correlation is performed on a wider vector so that the NCC measure might be more representative.

rtavenar:
I just merged it, but did not take the trick for kSHAPE into account: for me, it's unrelated to initial centroids, so I prefer to keep things separate.

PR #60, opened 2018-08-14
fix index out of bounds

apoorv007:


rtavenar:
Hi @apoorv007 ,

Thanks for your interest in tslearn. And thanks for the fix. I still have one question though: should'nt it test for `sz>0`, because `sz=0` should produce something unexpected, no?

Also, could you add an example in the doctest of the function that would not work with previous version of the code but now works as expected?

GillesVandewiele:
No, sz=0 is to be expected, since you are just iterating over the elements in that function and checking whether they np.isfinite(.) is True. So making a check `sz>0` would always result in the function returning 1 to less (if the first element would be finite).

You could actually rewrite the function to 
```
ts_ = = to_time_series(ts)
return sum(np.any(np.isfinite(ts_), axis=1))
```

rtavenar:
I added doctests to be explicit about what I would expect from `ts_size` when provided a series of all NaNs. Waiting for the checks to run before merging.

PR #62, opened 2018-08-17
fix index out of bounds II (based on #60)

GillesVandewiele:
based on PR #60 

* ts_size([[numpy.nan, numpy.nan], [3, 4]]) would have returned 2 in the past, returns 1 now.
* converted code to numpy, 1 single line now (not sure whether it works faster)
* did not find an example where an IndexOutOfBoundsError occurred though (which was the original reason behind the creation of this PR)...

### DIFF
--- a/tslearn/utils.py
+++ b/tslearn/utils.py
@@ -320,6 +320,8 @@ def ts_size(ts):
     3
     >>> ts_size([[1, 2], [2, 3], [3, 4], [numpy.nan, 2], [numpy.nan, numpy.nan]])
     4
+    >>> ts_size([[numpy.nan, numpy.nan], [3, 4]])
+    1
     """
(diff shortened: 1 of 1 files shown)


rtavenar:
Well, the point is it does something different from what `ts_size` was intended to.

From the docstring:

> Final timesteps that have NaN values for all dimensions will be removed from the count.

The idea was only to remove NaNs that occurred at the end of time series because these ones are probably here only because the time series is part of a varying-length time series dataset.

Typically, one would expect that `ts[:ts_size(ts)]` would give `ts` without the possibly padded NaNs.

GillesVandewiele:
Ok, I must have misinterpreted the docs then. My apologies, I shall close the PR :)

PR #75, opened 2018-10-17
Fixed loading of some UCR datasets by additional string parsing

MarcCoru:
I noticed that some datasets did not load correctly. These datasets return `None` instead of numpy arrays. Reasons were mostly inconsistcies between dataset names (defined in the first column of `http://www.timeseriesclassification.com/singleTrainTest.csv`) and filenames in the downloaded ZIP files. 

Two changes fixed the failed loading attempts:

(1) Fixed loading of some _UCR UEA datasets_ `Wafer`, `Yoga`, `UWaveGestureLibrary{X,Y,Z}`, `Fish`, `Mallat` by adding a string parsing function `harmonize_dataset_filenames` that ensures that datafiles follow a common naming convention (e.g.,` <datasetname>_TRAIN.txt`).  

(2) Added typo-fix in `http://www.timeseriesclassification.com/singleTrainTest.csv` after download by in-file string replacement with the new `inplace_change`-function.
This corrects "Fatal" to "Fetal" in the `NonInvasiveFetalECGThorax{1,2}` datasets and changes the download url to the correct `http://www.timeseriesclassification.com/Downloads/NonInvasiveFetalECGThorax1.zip`.

Minor changes:

* Removed `self._filenames`-dictionary: The changes (1) and (2) made this dictionary obsolete (it originally recorded filename anomalies).

* Added doctests for datasets that previously failed (`Fish` and `NonInvasiveFetalECGThorax1`)

### DIFF
--- a/tslearn/datasets.py
+++ b/tslearn/datasets.py
@@ -22,6 +22,32 @@ from tslearn.utils import to_time_series_dataset
 
 __author__ = 'Romain Tavenard romain.tavenard[at]univ-rennes2.fr'
 
+def harmonize_dataset_filenames(folder, verbose=False):
+    """
+    Checks if tilename corresponds to the dataset name format and deletes .DS_Store files if exists
+
+        <UCR_cache>/
+            <datasetname>/
+                <datasetname>_TRAIN.txt
+                <datasetname>_TEST.txt
+                ...
+
+    :param folder: unzipped dataset folder
+    :return: None
+    """
+    dataset_name = os.path.basename(folder)
+
+    # delete hidden .DS_Store directory if exists
+    if os.path.exists(os.path.join(folder, ".DS_Store")):
+        os.remove(os.path.join(folder, ".DS_Store"))
+
+    for f in os.listdir(folder):
+        if not f.startswith(dataset_name):
+            new_f = dataset_name + "_" + f.split("_")[-1]
+
+            os.rename(os.path.join(folder,f), os.path.join(folder,new_f))
+            if verbose: print("fixed " + os.path.join(folder, f) + "->" + os.path.join(folder, new_f))
+
 
 def extract_from_zip_url(url, target_dir=None, verbose=False):
     """Download a zip file from its URL and unzip it.
@@ -48,6 +74,8 @@ def extract_from_zip_url(url, target_dir=None, verbose=False):
         if not os.path.exists(target_dir):
             os.makedirs(target_dir)
         zipfile.ZipFile(local_zip_fname, "r").extractall(path=target_dir)
+        datasetfolder = os.path.join(target_dir, os.path.splitext(fname)[0])
+        harmonize_dataset_filenames(datasetfolder)
         shutil.rmtree(tmpdir)
         if verbose:
             print("Successfully extracted file %s to path %s" % (local_zip_fname, target_dir))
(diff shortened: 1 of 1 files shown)


tslearn/datasets.py
```diff
@@ -22,6 +22,32 @@
...
+    """
+    Checks if tilename corresponds to the dataset name format and deletes .DS_Store files if exists
+
+        <UCR_cache>/
+            <datasetname>/
+                <datasetname>_TRAIN.txt
+                <datasetname>_TEST.txt
+                ...
+
+    :param folder: unzipped dataset folder
```
rtavenar: Here, could you use `numpy` style docstring, as done in the rest of the tslearn package?

rtavenar: Same comment for all your docstrings btw :)

tslearn/datasets.py
```diff
@@ -296,3 +331,13 @@ def load_dataset(self, dataset_name):
         y_train = npzfile["y_train"]
         y_test = npzfile["y_test"]
         return X_train, y_train, X_test, y_test
+
+if __name__ == "__main__":
```
rtavenar: Are there still failures on your side (eg because of files not available)?

tslearn/datasets.py
```diff
@@ -214,20 +250,19 @@ def load_dataset(self, dataset_name):
...
             for fname in [fname_train, fname_test]:
                 if os.path.exists(os.path.join(full_path, fname)):
                     os.remove(os.path.join(full_path, fname))
             extract_from_zip_url(url, target_dir=self._data_dir, verbose=False)
-        try:
-            data_train = numpy.loadtxt(os.path.join(full_path, fname_train), delimiter=",")
-            data_test = numpy.loadtxt(os.path.join(full_path, fname_test), delimiter=",")
-        except:
-            return None, None, None, None
+
```
rtavenar: You changed the behaviour here (returning `None` if the data is not available): have you checked whether doctests should be updated (it seems so, if you have a look at Travis CI failures)?

tslearn/datasets.py
MarcCoru: All datasets are loading correctly. I tested it on a new machine without cached datasets.

In future, if remote files change (i.e., UCR zip archives, or http://www.timeseriesclassification.com/singleTrainTest.csv) the loading may fail again, though.

tslearn/datasets.py
MarcCoru: I fixed the doc strings in recent commits.
Have a look if that works out for you :-)
Unfortunately, I can't get the Cognitive Complexity below 5.
I assume that it is connected with the level of indentations. I added comments and more useful variable names, though. In my opinion `harmonize_dataset_filenames` is now quite readable.
However, I am open to more suggestions :-)

tslearn/datasets.py
MarcCoru: In recent commits, I have reverted the changes back to the try-catch blog not to change the behavior

tslearn/datasets.py
```diff
@@ -295,4 +343,4 @@ def load_dataset(self, dataset_name):
         X_test = npzfile["X_test"]
         y_train = npzfile["y_train"]
         y_test = npzfile["y_test"]
-        return X_train, y_train, X_test, y_test
+        return X_train, y_train, X_test, y_test
```
rtavenar: Could you please add a newline at the end of line (PEP8 rule)?

PR #77, opened 2018-10-21
Increase training and inference performance for GAK kernel

lejafar:
Hi! I've been using the `TimeSeriesSVC` with the `gak` kernel for a while now and I've made some changes that increased the performance quite a bit, the most time consuming part in my case was computing the kernel and its there where I've made some changes

### Changes that apply to training:

To compute the self-similarity, scikit-learn evaluates the [callable kernel using a reference of X](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/svm/base.py#L251) as the second argument, in the [current implementation will check if the second argument is `None`](https://github.com/rtavenar/tslearn/blob/master/tslearn/metrics.py#L353) which it isn't, hence it computes everything twice

### Changes that apply to inference:

When doing prediction, [scikit-learn evaluates the callable kernel for every point (`X_fit`) seen up until then](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/svm/base.py#L331), but the only thing  libsvm really needs is the distance to its support vectors, hence we can omit computing the others.

### DIFF
--- a/tslearn/svm.py
+++ b/tslearn/svm.py
@@ -40,7 +40,7 @@ def _sparse_kernel_func_gak(sz, d, gamma, slice_support_vectors=None):
         
         if X is X_fit:
             X = X.reshape((-1, sz, d))
-            return cdist_gak(X, X, sigma=numpy.sqrt(gamma / 2.))
+            return cdist_gak(X, None, sigma=numpy.sqrt(gamma / 2.))
         
         if slice_support_vectors is not None:
             # slice out support vectors and only compute cross sim with them
@@ -173,6 +173,7 @@ class TimeSeriesSVC(BaseSVC):
         self.d = d
         if kernel == "gak":
             kernel = _sparse_kernel_func_gak(sz=sz, d=d, gamma=gamma)
+            # kernel = _kernel_func_gak(sz=sz, d=d, gamma=gamma)
         super(TimeSeriesSVC, self).__init__(C=C, kernel=kernel, degree=degree, gamma=gamma, coef0=coef0,
                                             shrinking=shrinking, probability=probability, tol=tol,
                                             cache_size=cache_size, class_weight=class_weight, verbose=verbose,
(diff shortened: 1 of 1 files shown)


tslearn/svm.py
```diff
@@ -169,8 +192,10 @@ def fit(self, X, y, sample_weight=None):
         sklearn_X = _prepare_ts_datasets_sklearn(X)
         if self.kernel == "gak" and self.gamma == "auto":
             self.gamma = gamma_soft_dtw(to_time_series_dataset(X))
-            self.kernel = _kernel_func_gak(sz=self.sz, d=self.d, gamma=self.gamma)
-        return super(TimeSeriesSVC, self).fit(sklearn_X, y, sample_weight=sample_weight)
+            self.kernel = _sparse_kernel_func_gak(sz=self.sz, d=self.d, gamma=self.gamma)
+        _self = super(TimeSeriesSVC, self).fit(sklearn_X, y, sample_weight=sample_weight)
```
rtavenar: Why create a new variable here? Why not just do something like:

```python 
super(TimeSeriesSVC, self).fit(sklearn_X, y, sample_weight=sample_weight)
self.kernel = _sparse_kernel_func_gak(sz=self.sz, d=self.d, gamma=self.gamma, slice_support_vectors=self.support_)
return self
```

tslearn/svm.py
```diff
@@ -32,6 +32,29 @@ def _kernel_func_gak(sz, d, gamma):
...
+            return cdist_gak(X.reshape((-1, sz, d)), None, sigma=numpy.sqrt(gamma / 2.))
+        
+        if slice_support_vectors is not None:
+            # slice out support vectors
+            sliced_X_fit = X_fit[slice_support_vectors]
+            gak_sim_dense = cdist_gak(X.reshape((-1, sz, d)), sliced_X_fit.reshape((-1, sz, d)), sigma=numpy.sqrt(gamma / 2.))
+
+            # act like nothing has happend ...
+            gak_sim = numpy.empty((len(X), len(X_fit)))
+            gak_sim[:, slice_support_vectors] = gak_sim_dense
```
rtavenar: Should we document somewhere that, at predict time, only those entries corresponding to SVs will be meaningful?

rtavenar:
Woops sorry for double posting my review...

tslearn/svm.py
lejafar: It can be mentioned but I see it more as standard behaviour for an SVM, I don't completely see why scikit-learn would not do this slicing by default for callable kernels, maybe there is a use case where this would be a problem ...

tslearn/svm.py
lejafar: You are right, no need for that extra variable :)

rtavenar: OK, could you commit accordingly?

tslearn/svm.py
rtavenar: You are right, and if I check the API for this class on sklearn, I do not see a single place where a end-user would end up playing with that matrix, so I guess keeping that warning in the code could be sufficient.

tslearn/svm.py
lejafar: Done!

lejafar:
@rtavenar anything else?

rtavenar:
Hi @lejafar 

Sorry it took me so long to get back to you. Your suggestion is still welcome but the codebase has changed quite a lot since you posted it. Specifically, we are working on a future release in the  dev branch.

So, my question is : would you be willing to close this PR and open a new one based on the current dev branch ?

The place where your hack could be inserted is probably around:
https://github.com/rtavenar/tslearn/blob/dev/tslearn/svm.py#L58-L60

PR #81, opened 2018-11-20
fix loading of datasets in datasets.py fixes issue-80

JermellBeane:
fixes: https://github.com/rtavenar/tslearn/issues/80


rtavenar:
Hi @MarcCoru ,

Could you check that? It seems that you changed that part because of a new format in UEA zip files.

PR #90, opened 2019-03-22
Add pre-requisites section

thusithathilina:
C++ build tools should be available on the machine in order to install through the pip


rtavenar:
Now that the docs are back, I will merge this PR if it passes all checks.

Issue #51, opened 2018-06-23
softdtw_barycenter is broken


fmailhot:
Trying to get barycenter averaging working on some audio data, but can't even get the examples in the docs to work.

```python
 $ ipython --no-banner

In [1]: from tslearn.utils import to_time_series_dataset
   ...: X = to_time_series_dataset([[1, 2, 3, 4], [1, 2, 3], [2, 5, 6, 7, 8, 9]])
   ...:

In [2]: from tslearn.barycenters import softdtw_barycenter
   ...: from tslearn.utils import ts_zeros
   ...: initial_barycenter = ts_zeros(sz=5)
   ...: bar = softdtw_barycenter(X, init=initial_barycenter)
   ...:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-7221f8ce968c> in <module>()
      2 from tslearn.utils import ts_zeros
      3 initial_barycenter = ts_zeros(sz=5)
----> 4 bar = softdtw_barycenter(X, init=initial_barycenter)

/Users/fredmailhot/anaconda/envs/turkish_nlp/lib/python2.7/site-packages/tslearn/barycenters.pyc in softdtw_barycenter(X, gamma, weights, method, tol, max_iter, init)
    484         # The function works with vectors so we need to vectorize barycenter.
    485         res = minimize(f, barycenter.ravel(), method=method, jac=True, tol=tol,
--> 486                        options=dict(maxiter=max_iter, disp=False))
    487         return res.x.reshape(barycenter.shape)
    488     else:

/Users/fredmailhot/anaconda/envs/turkish_nlp/lib/python2.7/site-packages/scipy/optimize/_minimize.pyc in minimize(fun, x0, args, method, jac, hess, hessp, bounds, constraints, tol, callback, options)
    448     elif meth == 'l-bfgs-b':
    449         return _minimize_lbfgsb(fun, x0, args, jac, bounds,
--> 450                                 callback=callback, **options)
    451     elif meth == 'tnc':
    452         return _minimize_tnc(fun, x0, args, jac, bounds, callback=callback,

/Users/fredmailhot/anaconda/envs/turkish_nlp/lib/python2.7/site-packages/scipy/optimize/lbfgsb.pyc in _minimize_lbfgsb(fun, x0, args, jac, bounds, disp, maxcor, ftol, gtol, eps, maxfun, maxiter, iprint, callback, maxls, **unknown_options)
[...]

Issue #94, opened 2019-04-01
Classifiers not supporting variable length 


rshtirmer:
```
X = utils.to_time_series_dataset([[1, 2, 3, 4], [1, 2, 3], [2, 5, 6, 7, 8, 9]])
y = [1, 1, 0]
clf = TimeSeriesSVC(kernel="gak", gamma=.1, sz=X.shape[1], d=X.shape[2])
clf.fit(X, y)
OR
clf = neighbors.KNeighborsTimeSeriesClassifier(metric="dtw")
clf.fit(X, y)
```
Both methods result in the following error:
```
 File "main.py", line 77, in <module>
    clf.fit(X, y)
  File "/usr/local/lib/python3.7/site-packages/tslearn/neighbors.py", line 224, in fit
    return super(KNeighborsTimeSeriesClassifier, self).fit(X_, y)
  File "/usr/local/lib/python3.7/site-packages/sklearn/neighbors/base.py", line 891, in fit
    X, y = check_X_y(X, y, "csr", multi_output=True)
  File "/usr/local/lib/python3.7/site-packages/sklearn/utils/validation.py", line 756, in check_X_y
    estimator=estimator)
  File "/usr/local/lib/python3.7/site-packages/sklearn/utils/validation.py", line 573, in check_array
    allow_nan=force_all_finite == 'allow-nan')
  File "/usr/local/lib/python3.7/site-packages/sklearn/utils/validation.py", line 56, in _assert_all_finite
    raise ValueError(msg_err.format(type_err, X.dtype))
ValueError: Input contains NaN, infinity or a value too large for dtype('float64').
```

mjip:
The SVC and KNN classifiers require the input data to be of equal length in order to fit properly. As a temporary fix, you could do some preprocessing on your own to replace the missing values: https://scikit-learn.org/stable/modules/impute.html#impute

rtavenar:
This is now fixed in dev branch and forthcoming version 0.2

PR #95, opened 2019-04-01
Tentative fix for #51

mjip:
Tentative fix for the issue seen in #51. SoftDTW is broken with time series of unequal lengths. This is a (tentative) fix I'm using, not sure how correct it is but the output looks consistent for my data. If every `barycenters` method enforced time series data/datasets instead of doing the conversion on the first line if the method, I think it would solve any other potential issues with re-inserting NaNs as well.


rtavenar:
Hi @mjip 

Thanks a lot for your help. I'm waiting for TravisCI to complete its inspection before merging, but if that solved bug #51 that would be great!

mjip:
I think the issue with #94 is that you'd have to define a policy to replace the missing values in the time series, since the classifiers need data of equal length to fit on. e.g. something like this: https://scikit-learn.org/stable/modules/impute.html#impute

PR #104, opened 2019-05-15
Update DTW and constraint regions computations.

johannfaouzi:


rtavenar:
Hi @johannfaouzi and thanks a lot for this PR

Just to give some context:
- this PR is related to the reddit thread there: https://www.reddit.com/r/MachineLearning/comments/bb59ji/p_pyts_a_python_package_for_time_series/ekgtkhw?utm_source=share&utm_medium=web2x
- your `numba` implementation makes DTW computations much faster, which is great
- you also fixed some bugs in global constraints (Itakura parallelogram and Sakoe-Chiba band)

On the last point, it seems your implementation do not fit the doctest:
https://travis-ci.org/rtavenar/tslearn/jobs/532821769#L1027

I will check asap if it is the doctest that is incorrect or if the issue is in your implementation.

johannfaouzi:
The `sakoe_chiba_band` test is failing:

```python
Failed example:
    sakoe_chiba_mask(7, 3, 1)  # doctest: +NORMALIZE_WHITESPACE

Expected:
    array([[  0., 0., inf],
           [  0., 0., inf],
           [  0., 0., inf],
           [  0., 0.,  0.],
           [ inf, 0.,  0.],
           [ inf, 0.,  0.],
           [ inf, 0.,  0.]])
Got:
    array([[  0.,   0.,  inf],
           [  0.,   0.,   0.],
           [  0.,   0.,   0.],
           [  0.,   0.,   0.],
           [  0.,   0.,   0.],
           [  0.,   0.,   0.],
           [ inf,   0.,   0.]])
```

In my opinion, the actual result is the good result. The absolute difference between both sizes is 4. When one adds a radius of 1, one should a difference of 2 (* 1) (one for each side), hence a band of size 4 + (2 * 1), and not 4.

rtavenar:
I agree with your definitions for the Sakoe-Chiba mask and Itakura parallelogram.

I still have a few more questions:
1. Shouldn't we add a `itakura_max_slope` additional parameter to the `dtw` function then?
2. Wouldn't there be more functions in the `metrics` module that could benefit from `numba` for fast computations? I was thinking of: `dtw_path`, `dtw_subsequence_path`, `sakoe_chiba_mask`, `itakura_mask`, and `cdist_dtw`. If so, let me know which ones you can handle, I should be able to code the other ones.

Romain

johannfaouzi:
I will update the doctests accordingly.

You're right, `itakura_max_slope` should be added in all the functions that also have `sakoe_chiba_radius` as a parameter, I will update them.

Regarding the following functions:

* `sakoe_chiba_mask` is already sped up thanks to `numba`,

* `itakura_mask` makes heavy use of numpy so I don't think that we should get much improvement for this one,

* `cdist_dtw`, `dtw_path` and `dtw_subsequence_path` could probably be sped up with `numba`, I will work on them.

Just a question about `cdist_dtw`: it computes the DTW for each pair from `dataset1` and `dataset2`, right? If `dataset2` is None, it is like `dataset2=dataset1`, like [sklearn.metrics.pairwise.pairwise_distances](https://scikit-learn.org/0.19/modules/generated/sklearn.metrics.pairwise.pairwise_distances.html)?

rtavenar:
> Just a question about `cdist_dtw`: it computes the DTW for each pair from `dataset1` and `dataset2`, right? If `dataset2` is None, it is like `dataset2=dataset1`, like `sklearn.metrics.pairwise.pairwise_distances`?

Yes, this is it. Hence in the None case, we should save computations (symmetric matrix).

rtavenar:
`numba` is not happy with variable return types...
Maybe the easiest way would be to have a `numba`-optimized function that returns both the distance and the path and the other one that calls the first one and can return either both or only the distance depending on a argument that is passed, no?

johannfaouzi:
I am working on this :) I am not used to doctest as a test suite and did not check my code enough, but I will fix this.

johannfaouzi:
I am not sure that I understand very well what `dtw_subsequence_path` does, so I did not change it. If you can give me more details, I can try to improve it with `numba`.

johannfaouzi:
When I run `` python -m doctest -v metrics.py`` locally, it works, but the CI does not seem to succeed. I'm not sure why it is happening, maybe the doctests in ``metrics.py`` don't cover the usages in the other files.

johannfaouzi:
I finally fixed the issues. Let me know if you think that more stuff could be changed or added.

rtavenar:
I just pushed a simple bugfix, will work on `dtw_subsequence_path` next week and then merge to dev branch.

PR #110, opened 2019-06-07
Use scipy to load arff data

yichangwang:
Hi Romain,

I made an implementation to load the arff file from UEA / UCR dataset with SciPy.

SciPy is not capable to load the arff file for the SpokenArabicDigits, because there's an extra `\n` for each train/test example in the arff file. So for compatibility, I still kept the `remake_files` function to transfer `arff` to `csv`.

Cheers,
Yichang

### DIFF
--- a/tslearn/datasets.py
+++ b/tslearn/datasets.py
@@ -19,6 +19,7 @@ except ImportError:
     from zipfile import BadZipfile as BadZipFile
 
 from tslearn.utils import to_time_series_dataset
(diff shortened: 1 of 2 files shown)


yichangwang:
Hi Romain,

Scipy1.3.0 (which supports relational attributes in arff file) doesn't work with Python<=3.4.

So users using an older scipy or older python may get compatibility problem with this feature.

PR #115, opened 2019-06-28
[MRG] Use both doctests and standard unit tests

rtavenar:
This PR aims at using doctests only when the test is helpful for the end user and pushing tests that do not help the end-user in a dedicated folder tests that would contain those ones.


rtavenar:
I just changed my mind, I think we'd rather merge that to dev once ready. Hence I started resolving conflicts.

Now it seems some tests fail, will have to look at it.

Also, I will not fix the decrease in code coverage since it mainly concerns the deprecated estimators from `tslearn.barycenters` that will be removed in 0.2.

rtavenar:
It was a good idea to merge to `dev`, since it revealed an issue with string comparison in `numba`, `nopython` mode. Now, things seem to be fixed.

PR #119, opened 2019-07-08
[MRG] Remove Cython compiled files

johannfaouzi:
#### What does this PR fix?

This PR removes the Cython compiled files (`*.c`) and adds this extension in the `.gitignore` file.


johannfaouzi:
I'm not sure to understand why there are conflicts on my branch...

rtavenar:
Hi @johannfaouzi and thanks for this fix

I agree that this should be merged to dev, but that means that conflicts should be resolved (probably nothing). I also noticed that the travis file you use is not up to date with dev branch, if I'm correct, could you check that?

Romain

rtavenar:
> I'm not sure to understand why there are conflicts on my branch...

Did you create your branch from dev or from master? Maybe that is the source (and maybe that also answers my question about the travis file)?

johannfaouzi:
From master, you're right! I'm going to fix that.

rtavenar:
Thanks @johannfaouzi !

Issue #120, opened 2019-07-09
ShapeletModel extensions


GillesVandewiele:
The following stuff would be very interesting to implement/integrate in the current version of ShapeletModel.

1) Remove all logics for the binary case, and always use a "softmax" activation in the output layer (even for binary case).
--> `predict_proba()` would just be `model.predict()`
--> `predict` would be `predict_proba().argmax(axis=1)`

2) Making ShapeletModel serializable/picklable and removing SerializableShapeletModel afterwards.
--> This will require patching of some keras functions. We need to make sure we do not need to use a very new version in order to have backward compatibility.

3) During the refactoring to make all estimators sklearn-compliant, we noticed weird issues on a simple dataset with a low learning rate. The dataset should be very easy to classify, but the model achieved 0% accuracy, which is much worse than random. It seems like it is learning the opposite of what it should learn.
```
dataset:
----------
random_walk_blobs(n_ts_per_blob=100, random_state=42,
                                 n_blobs=3, noise_level=0.1, sz=100)

learning rate:
-----------------
K.set_value(self.model.optimizer.lr, 0.001)
```

4) Implement #118 and create an example notebook with some plot.

GillesVandewiele:
We have also been noticing some weird convergence issues in the patched `check_classifiers_train` test for both shapelet models that should be looked at. (See PR #116 )

GillesVandewiele:
Related to this are the following issues: #96 , #85 , and #53 

@rtavenar this is something I could do when my next PR is finished.

rtavenar:
OK, great!

Once again, I disagree with removing all logics that are specific to the binary case, since they correspond to what is written in the original paper, and I think we should stick to implementing what is in the paper.

But I fully agree that it is a mess at the moment and should be made clearer. Thanks for taking care of that @GillesVandewiele

rtavenar:
@GillesVandewiele could you check which of these items are still on our todo list ? It seems some of them are outdated,at least.

GillesVandewiele:
It indeed looks like most of those points are handled, except for the SerializableShapeletModel (here I am again :) ). You did indeed mention in a previous PR that there are some issues, but I remember that we were able to fix these some time ago, but that it required a too recent version of tensorflow.

I still have to take a look whether it can actually be fixed. It would allow us to remove a significant chunk of code that does not have that much added value.

rtavenar:
[...]

PR #126, opened 2019-07-25
[MRG] Shapelet Support Tensorflow 2

page1:
Made a few changes to support Tensorflow 2, and remove Keras as a separate dependancy. I'm just testing out tslearn and am not sure if these changes are wanted. No offense will be taken if these don't get included. :)

Have a great day, I'm excited to see what tslearn has to offer.


page1:
Build failures are to be expected since this is working on TF2

rtavenar:
Dear @page1 

Thank you for your interest in tslearn, and for this nice port to TF2.

However, I would prefer tslearn not to depend too much on recent versions of the required packages, since it could prevent some people in a restricted environment from using tslearn. So I will not consider this PR for merging at the moment since TF2 is rather new.

Thank you anyway,
Best regards,
Romain

page1:
Makes sense!

rtavenar:
OK, now time has gone and moving to TF2 would make more and more sense (and it would allow to fully support python3.8 also). So I'm re-opening this one and will check sooner or later if I agree with the upgrade :)

Thanks again @page1 :)

rtavenar:
@GillesVandewiele you might like this one ;) If you can have a look, that would be great.

rtavenar:
@page1 this was great job, I only had to merge new changes from master, update tests and dependencies. Thanks a lot for this PR!

rtavenar:
I think this PR fixes #193 #158 #120

rtavenar:
I think this PR fixes #193 #158 #120

GillesVandewiele:
@rtavenar I will definitely take a look :), but it will be somewhere next week if that's alright?

rtavenar:
> @rtavenar I will definitely take a look :), but it will be somewhere next week if that's alright?

That's perfect. As far as I'm concerned, the PR is ready. And you can check it when you can.

rtavenar:
Note: the failing test is just an issue with codecov report uploader: our code works OK :)

rtavenar:
And we have a bonus: now the `ShapeletModel` is able to deal with variable-length inputs!
The only pity is that is raises an error in `sklearn` checks since `ShapeletModel` is a `Transformer` and there is a check that ensures transformers raise an exception if the data passed at transform time does not have the same number of features as the data passed at fit time. Will probably require one more patch...

rtavenar:
> EDIT: Oh, and by the way, it seems TF is not happy with input tensors not having the same shape (but NaNs are OK), so if predict-time inputs are smaller than fit-time ones, we can still pad them with nans, but if they are longer, I don't know what to do... :/ --> for this one I could maybe try to have `None` values in the input shapes, I'll see...

I added an extra parameter. Its default keeps the original behaviour (ie. works OK if fit/predict are performed on series of the same length) and setting another value allows to secure some room for later calls with longer series (up to the provided size).

However sklearn checks should still fail (still need to be patched).

CHANGELOG.md
```diff
@@ -26,12 +30,22 @@ of which the DTW algorithm can be run
...
 * Multivariate datasets from the UCR/UEA archive are now available through
 `UCR_UEA_datasets().load_dataset(...)`
+* `ShapeletModel` now accepts variable-length time series dataset; a `max_size`
+parameter has been introduced to save room at fit time for possibly longer
+series to be fed to the model afterwards
+* `ShapeletModel` now accepts a `scale` parameter that drives time series
+pre-processing for better convergence
+* `ShapeletModel` now has a public `history_` attribute that stores
+loss and accuracy along fit epochs
+* SAX and variants now accept a `scale` parameter that drives time series
```
GillesVandewiele, line 40: Perhaps we could make a small note here that these will be applied on the entire dataset as opposed to the regular per-timeseries normalization?

tslearn/bases.py
```diff
@@ -233,7 +233,12 @@ def from_json(cls, path):
         for k in model['model_params'].keys():
```
GillesVandewiele, line 233: Where in the code is from_json exactly used? Is it called by some dependency such as keras? tf.keras has a `model_from_json` as well, perhaps that could be used to solve this "hack"?

tslearn/docs/gen_modules/tslearn.shapelets.rst
```diff
@@ -25,7 +25,6 @@ tslearn.shapelets
       :template: class.rst
    
       ShapeletModel
-      SerializableShapeletModel
```
GillesVandewiele, line 28: :tada: :partying_face:

tslearn/shapelets.py
```diff
@@ -317,14 +345,21 @@ def __init__(self, n_shapelets_per_size=None,
         self.weight_regularizer = weight_regularizer
         self.shapelet_length = shapelet_length
         self.total_lengths = total_lengths
+        self.max_size = max_size
+        self.scale = scale
         self.random_state = random_state
 
         if max_iter == 10000:
-            warnings.warn("The default value of max_iter has changed "
+            warnings.warn("The default value for max_iter has changed "
```
GillesVandewiele, line 353: Perhaps we could remove this warning now since we are at a new major version?

tslearn/shapelets.py
```diff
@@ -172,7 +180,7 @@ def grabocka_params_to_shapelet_size_dict(n_ts, ts_sz, n_classes, l, r):
     Returns
     -------
     dict
-        Dictionnary giving, for each shapelet length, the number of such
+        Dictionary giving, for each shapelet length, the number of such
```
GillesVandewiele, line 183: Perhaps we should do a check here that the maximum shapelet length in the dictionary is smaller than the shortest timeseries in the dataset? Or will the distance function work there as well (but probably with a smaller value)?

We could include this in a test somewhere.

tslearn/shapelets.py
```diff
@@ -410,12 +444,13 @@ def fit(self, X, y):
...
         self._set_model_layers(X=X, ts_sz=sz, d=d, n_classes=n_labels)
         self._set_weights_false_conv(d=d)
-        self.model_.fit(
+        h = self.model_.fit(
             [X[:, :, di].reshape((n_ts, sz, 1)) for di in range(d)], y_,
             batch_size=self.batch_size, epochs=self.max_iter,
             verbose=self.verbose
         )
-        self.n_iter_ = len(self.model_.history.history)
+        self.history_ = h.history
```
GillesVandewiele, line 452: Are there any other interesting attributes in `h`? Perhaps we should store this instead of only the history?

tslearn/shapelets.py
```diff
@@ -893,12 +958,17 @@ def __init__(self, n_shapelets_per_size=None,
                              total_lengths=total_lengths,
                              random_state=random_state)
         self.learning_rate = learning_rate
+        warnings.warn(
```
GillesVandewiele, line 961: The old SerializableShapeletModel allowed to specify a learning rate as well. As this is redundant now, as that can be specified within the Optimizer-object, we could perhaps add it to the ShapeletModel until version 0.6 to allow backward compatibility even more?

CHANGELOG.md
rtavenar: At the moment, it is done per time series. We should decide which one to pick and document it clearly before release of v0.4, I agree!

GillesVandewiele:
This looks great Romain! I left a few comments will probably test the code a bit more thoroughly later today.

Another thing: since you now implemented a matrix profile as well, we could perhaps use that to initialize shapelets as well? 

We recently discussed how one could do this in an issue of tsfresh: https://github.com/blue-yonder/tsfresh/issues/658

CHANGELOG.md
GillesVandewiele: If it is done per timeseries, this will only capture "shape information" because every timeseries will range from 0 to 1. Originally, it was applied on the entire dataset. Perhaps we could make this configurable?

GillesVandewiele: The dataset normalization is a little bit harder. It will require you to store a mu and sigma per dimension during fit time and re-use these during predict time to normalize.

tslearn/bases.py
rtavenar: It's used when one wants to load a model from disk.
`model_from_json` loads the model structure but not the weights. I think we need this hack.

tslearn/shapelets.py
rtavenar: Yep, good idea!

tslearn/shapelets.py
rtavenar: OK, yes.

tslearn/shapelets.py
rtavenar: I've never used anything else personally. And `keras` docs (description of the `fit` method) seem to only stress out that dict:

```
Returns

A History object. Its History.history attribute is a record of training loss values and metrics values at successive epochs, as well as validation loss values and validation metrics values (if applicable).
```

Finally, a dict is much easier to serialize than a custom class, which is why I made this choice.

tslearn/shapelets.py
rtavenar: Well, I think I'd prefer that in future versions, there is only one way to set the learning rate (through the optimizer). We had this parameter for `SerializableShapeletModel` because we were unable to clone when a custom optimizer was provided. But we no longer have this problem.

rtavenar:
> This looks great Romain! I left a few comments will probably test the code a bit more thoroughly later today.

OK, thanks a lot for your review! I'll work on these asap.

> Another thing: since you now implemented a matrix profile as well, we could perhaps use that to initialize shapelets as well?
> 
> We recently discussed how one could do this in an issue of tsfresh: [blue-yonder/tsfresh#658](https://github.com/blue-yonder/tsfresh/issues/658)

I am far from being an expert of Matrix Profile, so I'm probably not the best person to ask if we should add that. Anyway, I think we can think about it and try to include it in a future release. Let me know what you think.

GillesVandewiele:
Ok, I could take a look at that in the future Romain! Perhaps we could combine your research concerning the location-information in there as well? Of course, all these additions should be configurable (such that people can still use the vanilla implementation as originally proposed by Grabocka)

CHANGELOG.md
rtavenar: I agree the dataset-level normalization is the right one for SAX (and it's not very difficult to implement I think). I'll fix this is another PR.

rtavenar:
> Ok, I could take a look at that in the future Romain! Perhaps we could combine your research concerning the location-information in there as well? Of course, all these additions should be configurable (such that people can still use the vanilla implementation as originally proposed by Grabocka)

Yep, we'll see. If we have the localized random shapelets, it should probably be another class.

GillesVandewiele:
If they are random, indeed, but do they have to be random? Can't the location information be included in the loss function and optimized?

rtavenar:
Well, localization is not smooth by nature, if you change the shapelet content a bit, localization can vary a lot, so gradient descent based optimization on a model that includes localization seems à risky option.

Sent from a mobile device. Please excuse my brevity.

________________________________

rtavenar:
OK, I think I implemented the improvements. Let me know if you have further comments.

GillesVandewiele:
> Well, localization is not smooth by nature, if you change the shapelet content a bit, localization can vary a lot, so gradient descent based optimization on a model that includes localization seems à risky option. Sent from a mobile device. Please excuse my brevity.

Aren't there any smooth variants of the argmin operator that can be implemented? But we can have a discussion about that sometime & somewhere else :). I'll quickly go another time through the changes, perhaps I can think of things which I could think of the first time :).

rtavenar:
> > Well, localization is not smooth by nature, if you change the shapelet content a bit, localization can vary a lot, so gradient descent based optimization on a model that includes localization seems à risky option. Sent from a mobile device. Please excuse my brevity.
> 
> Aren't there any smooth variants of the argmin operator that can be implemented? But we can have a discussion about that sometime & somewhere else :). I'll quickly go another time through the changes, perhaps I can think of things which I could think of the first time :).

You're right. Let's say I've never tested it and I cannot guarantee it would work.

GillesVandewiele:
When no shapelet size dictionary is provided, you currently call `grabocka_params_to_shapelet_size_dict` with the maximum timeseries. Is this correct? Shouldn't this be the minimum size as well in order to avoid that the longest shapelet length exceeds the minimum timeseries length?

```python3
        n_ts, sz, d = X.shape

        if self.n_shapelets_per_size is None:
            sizes = grabocka_params_to_shapelet_size_dict(n_ts, sz, n_labels,
                                                          self.shapelet_length,
                                                          self.total_lengths)
```

rtavenar:
> When no shapelet size dictionary is provided, you currently call `grabocka_params_to_shapelet_size_dict` with the maximum timeseries. Is this correct? Shouldn't this be the minimum size as well in order to avoid that the longest shapelet length exceeds the minimum timeseries length?
> 
> ```python
>         n_ts, sz, d = X.shape
> 
>         if self.n_shapelets_per_size is None:
>             sizes = grabocka_params_to_shapelet_size_dict(n_ts, sz, n_labels,
>                                                           self.shapelet_length,
>                                                           self.total_lengths)
> ```

Good catch! I'll do that and document it.

tslearn/shapelets.py
```diff
@@ -530,16 +568,72 @@ def locate(self, X):
...
+                                 "Longest time series is of "
+                                 "length {} and max_size is "
+                                 "{}".format(max_sz_X, self.max_size))
+
+    def _preprocess_series(self, X):
+        if self.scale:
+            X = TimeSeriesScalerMinMax().fit_transform(X)
+        else:
+            X = to_time_series_dataset(X)
+        if self.max_size is not None and self.max_size != X.shape[1]:
```
GillesVandewiele, line 624: Shouldn't this be `<=` instead of `!=` ?

rtavenar: Well:

* if `self.max_size == X.shape[1]` then we have no problem
* if `X.shape[1] > self.max_size` then we raise an error

so the only real reshaping is done when `self.max_size < X.shape[1]`, which seems ok, no?

GillesVandewiele: Ok yes I looked over that if on the next line :)

seanlaw:
> I am far from being an expert of Matrix Profile, so I'm probably not the best person to ask if we should add that. Anyway, I think we can think about it and try to include it in a future release. Let me know what you think.

@GillesVandewiele @rtavenar Regarding matrix profiles, please let me know if there's anything I or [STUMPY](https://github.com/TDAmeritrade/stumpy) can do to help!

rtavenar:
Hi @seanlaw 

Now that you ask... :-p

As for now, we have a minimal naive implementation of matrix profile. In #256 we are working on a user guide page for matrix profiles (which references stumpy by the way :). Your feedback would probably be highly valuable.

At the moment, we believe `tslearn` is probably not the best place for implementation of high-performance tweaks related to matrix profiles, but maybe:
* there are variants of matrix profile that we don't know of and that we could implement even with our pure-numpy based approach (ie. to target more diverse tasks)
* we could at some point work harder on the casting utilities to allow better integration of high-performance stumpy implementations in `tslearn` typical pipelines

WDYT?

rtavenar:
By the way, thanks for your intro tutorial on MPs: it's so easy to follow! My implementation of MPs is based on my understanding of your tutorial.

GillesVandewiele:
Just to complement Romain here, you can find the user guide [here](https://tslearn--256.org.readthedocs.build/en/256/user_guide/userguide.html), as I can imagine it is not that trivial to get there for someone doesn't know the CI/CD of tslearn.

Moreover, I think it would be interesting to know whether it would be possible to implement the matrix profiling on a GPU using Keras (as that is one of our dependencies currently).

As Romain said, your tutorials are amazing and very intuitive!

seanlaw:
> Just to complement Romain here, you can find the user guide here, as I can imagine it is not that trivial to get there for someone doesn't know the CI/CD of tslearn.

Thank you! This link really helped as I was staring at SVG code and trying to figure out how to render it. What's the best way to offer suggestions? Shall I leave comments in the WIP link?

> Moreover, I think it would be interesting to know whether it would be possible to implement the matrix profiling on a GPU using Keras (as that is one of our dependencies currently).

So, I can't comment on whether or not it can be implemented in Keras (I've only used it briefly for RNNs) but we did consider implementing our GPU matrix profile code in CuPy or TF and, ultimately, we didn't want to add yet another dependency (STUMPY technically only has three dependencies). While I'm sure that others may be able to accomplish it, I found the TF/Keras API to be too high level and I wanted something that was closer to writing a raw CUDA kernel but still in "Python-land".

From your `requirements.txt` it looks like you already have `Numba` as a dependency. Currently, STUMPY's `gpu_stump` implementation is CUDA JIT-compiled by `Numba`. While respecting and adhering to the BSD-3 license, you are free to borrow whatever code you need from the STUMPY code base and, since we maintain 100% code coverage, our unit tests should be sufficient (as a starting point) to let give you an idea if you've broken anything. Of course, I can assist with any questions or concerns that you may have.

Does that answer your question or did I misunderstand your point?

> As Romain said, your tutorials are amazing and very intuitive!

I'm glad that you've found value in them! Sometimes, writing documentation can feel like talking to the wind and I never get to "hear" what people think. I appreciate you taking the time to let me know!

rtavenar:
Yeah, I think you can either leave comments in the PR thread or directly comment on the code related to the user guide:
https://github.com/tslearn-team/tslearn/pull/256/files#diff-fb818d30c932a83c1b6cbe43d4ca87d1

Regarding better efficiency, if anyone wants to work on a gpu+numba improvement based on stumpy code that would be an idea (I won't have time for it, I'm afraid).

seanlaw:
> As for now, we have a minimal naive implementation of matrix profile. In #256 we are working on a user guide page for matrix profiles (which references stumpy by the way :). Your feedback would probably be highly valuable.

Okay, I will take a look and leave some comments/suggestions where it may be appropriate. If you can't tell, I tend to be quite verbose so please feel free to accept/ignore any suggestions that I make and know that I will not be offended. I'd rather be thorough than miss anything.

> At the moment, we believe tslearn is probably not the best place for implementation of high-performance tweaks related to matrix profiles, but maybe:
>
> there are variants of matrix profile that we don't know of and that we could implement even with our pure-numpy based approach (ie. to target more diverse tasks)
we could at some point work harder on the casting utilities to allow better integration of high-performance stumpy implementations in tslearn typical pipelines

As I mentioned above, if you adhere to and respect the BSD3 license then I don't see any problem with you borrowing code from STUMPY (attribution is nice but it's open source 😄). What's really interesting is that a "pure-NumPy" approach technically already exists in the [`stumpy.stomp` implementation](https://github.com/TDAmeritrade/stumpy/blob/master/stumpy/stomp.py) (it was purposely left in the code base as a reference). While it calls out to other `Numba` functions, you can easily just remove/comment out the `Numba` decorators and it will revert back to pure-NumPy/Python.

seanlaw:
> Regarding better efficiency, if anyone wants to work on a gpu+numba improvement based on stumpy code that would be an idea (I won't have time for it, I'm afraid).

Firstly, I don't mean to sound intrusive as I am in no position to dictate how you should develop your software (and I have also struggled with whether or not to add Dask as a dependency in STUMPY so I completely understand) but is there a reluctance/hesitancy to add STUMPY as a (soft) dependency? STUMPY only depends on `scipy`, `numpy`, and `numba` (these are the only hard dependencies) and they are already in your `requirements.txt`. We take testing very seriously and we purposely maintain 100% code coverage in order commit to certain a level of confidence for our users. This way, you don't have to support matrix profile computations (this can become a bigger task). Again, I'm just talking out loud so feel free to ignore me 😄.

rtavenar:
You're right, that would probably be a better option. My main concern is that I'm not familiar with stumpy, so I'm not sure how much work that would be to wrap STUMPY functions into sklearn-compatible transformers as we have.

Maybe that would be the best way to start: we could advertise that in a new issue asking for help with improving the matrixprofile module by making it an interface for stumpy utilities.
If so, would you be OK to review the PR if someone offers help?

seanlaw:
> You're right, that would probably be a better option. My main concern is that I'm not familiar with stumpy, so I'm not sure how much work that would be to wrap STUMPY functions into sklearn-compatible transformers as we have.

STUMPY was purposely developed to be as "flat" as possible in order to remove any complexity. So, we ultimately chose not to employ OOP/classes (though we had considered it) and, instead, STUMPY is implemented as a set of stateless Python functions that are placed inside of organized modules. This means that there is no sublcassing/inheritance to wrap your head around and it should, in theory, be relatively straightforward to wrap. Having said that, I've not created an sklearn transformer before (I've only consumed them) but if you want to direct me to a typical/common example then I can take a look?

> Maybe that would be the best way to start: we could advertise that in a new issue asking for help with improving the matrixprofile module by making it an interface for stumpy utilities.
If so, would you be OK to review the PR if someone offers help?

Absolutely! I am excited and here to help.

Issue #55, opened 2018-07-30
Is there a way to parallelise the algorithms?


tolgamalachi:
Hello guys,

first of all thanks for this great project!
Right now i am heavily using the k-Shape-Algorithm for clustering (imho) huge amounts of data.
n = 5.000.000, k = 6;8;10;12, m = 56
O(max{n·k ·m· log(m), n·m2, k ·m3}) (complexity of k-Shape from the original paper)

And this takes about 150hours on an Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz. Sadly the algorithm only uses one core. How can we fix this? How can we make this faster? I am quite new to Python but maybe you have an idea?

Thanks in advance =)

GillesVandewiele:
Hello, I do not really know the k-shape algorithm, but I would try to parallelize the following methods:

`_update_centroids`: https://github.com/rtavenar/tslearn/blob/master/tslearn/clustering.py#L694

AND

`_assign`: https://github.com/rtavenar/tslearn/blob/master/tslearn/clustering.py#L704

But this would require you to also move away from the current numpy code, so I cannot ensure that it would actually work much faster...

jlc-christie:
@tolgamalachi might be worth installing numba and seeing if you can wrap the methods @GillesVandewiele in the `@jit` decorator, e.g. `@jit(nopython=True, parallel=True)`.

Doesn't always work well but if the function is simple enough it often produces good optimizations with almost no effort. However, the parallel=True argument is an experimental feature according to numba so you might have to look in to the implications of that.

andmib:
Has anybody tried wrapping `TimeSeriesKMeans` with `@jit`? Was it successfully? I'd also like to find a way to speed up extremely large `n` datasets; particularly when doing dynamic time warping, it takes a lot of computation time but is only utilizing a single core.

choltz95:
Also requesting parallel DTW & GAK

GillesVandewiele:
As far as I know, DTW and GAK are stateless, you provide it with two timeseries and you get a similarity score returned. Most often, you want to construct a pairwise distance matrix, for which the calculation can easily be parallellized using `multiprocessing` from Python. You can divide your matrix in blocks and let each process calculate its own block.

choltz95:
Yes - it would be great to integrate multiprocessing into the cdist_dtw/cdist_gak computation.

rtavenar:
[...]

PR #128, opened 2019-08-17
[MRG] Enable parallel computations in cdist_*

rtavenar:
At the moment, I have worked on a first attempt for cdist_dtw.

This is related to the discussion in #55


tslearn/metrics.py
```diff
@@ -9,6 +9,7 @@
     _jacobian_product_sq_euc
 from sklearn.metrics.pairwise import euclidean_distances
 from numba import njit, prange
+from sklearn.externals.joblib import Parallel, delayed
```
johannfaouzi: joblib will be vendored from scikit-learn in a future release (0.23 or 0.24 I think), and is already a dependency of scikit-learn. It would be better to add it as a dependency and to import from the original package instead of the version saved in scikit-learn.

tslearn/metrics.py
```diff
@@ -79,7 +79,7 @@ def njit_dtw(s1, s2, mask):
         Second time series
 
     mask : array, shape = (sz1, sz2)
-        Mask. Unconsidered cells must have infinite values. d
+        Mask. Unconsidered cells must have infinite values.
```
GillesVandewiele: Just a general comment: but wouldn't it be somewhat more user-friendly/intuitive to accept a binary matrix here and just multiply it (or 1 - matrix) with np.inf ourselves?

tslearn/metrics.py
```diff
@@ -653,21 +654,28 @@ def cdist_dtw(dataset1, dataset2=None, global_constraint=None,
...
-        global_constraint_str = ""
-
     if dataset2 is None:
+        if global_constraint is not None:
+            global_constraint_str = global_constraint
+        else:
+            global_constraint_str = ""
+
+        # TODO
         return njit_cdist_dtw_self(
```
GillesVandewiele: Can't we do something similar here (I'm assuming we want pairwise distances between all timeseries within `dataset1`). Perhaps here we could just say `dataset2 = dataset1` and then always call the code of L670-677

GillesVandewiele:
I added 2 small comments as well (not too familiar with the DTW calculation code in tslearn), they do not appear here though as they are single comments (you'll need to check the "files changed" tab)

tslearn/metrics.py
johannfaouzi: * From a Python point of view, I would agree: When I think of masks, I usually think of a numpy.array with boolean dtype (with True and False values only).

* From a math point of view, I would disagree: It's quite standard to use infinite values for dynamic programming.

However, it's a Python package so I would agree ;)

tslearn/metrics.py
johannfaouzi: When computing pairwise distances between all timeseries within `dataset1`, the matrix is symmetric and all the elements on the diagonale are zeros. There are only `n * (n-1) / 2` values to compute instead of  `n * n`.

tslearn/metrics.py
GillesVandewiele: The boolean numpy array as you suggested is a better alternative than a binary matrix as well I'd say!

johannfaouzi:
I will try to give it a look next week (or probably the week after), but I think it could be interesting to run some benchmarks, something like:

```python
import numpy as np
from tslearn.metrics import cdist_dtw

n_samples, n_features, n_timestamps = 100, 3, 500
rng = np.random.RandomState(42)
X = rng.randn(n_samples, n_features, n_timestamps)
Y = rng.randn(n_samples, n_features, n_timestamps)

# With the version on this branch
%timeit cdist_dtw(X, Y, n_jobs=4)  # a lot of laptops have 4 cores

# With the dev branch
%timeit cdist_dtw(X, Y)
```

and see if it's 4 times (probably a bit less, there is some overhead when using several cores) faster or not.

tslearn/metrics.py
rtavenar: The idea here is to save half the computations using the fact that dtw(x,y)=dtw(y,x) and dtw(x,x)=0

rtavenar:
@johannfaouzi That would be great indeed, I can improve the code on my side. Also, if this works, we should think about extending it to all cdist functions and all estimators that rely on cdist functions.

tslearn/metrics.py
GillesVandewiele: Hmm ok, it would still be possible to call the Parallel code on this right?

```
matrix = np.zeros((len(dataset1), len(dataset2)))
matrix[np.triu_indices(len(dataset(1), m=len(dataset2))] = Parallel(n_jobs=n_jobs)(
            delayed(dtw)(
                dataset1[i], dataset2[j],
                global_constraint=global_constraint,
                sakoe_chiba_radius=sakoe_chiba_radius,
                itakura_max_slope=itakura_max_slope)
            for i in range(len(dataset1)) for j in range(i + 1, len(dataset2))
        )
matrix = matrix + matrix.T
```

rtavenar: This is exactly the kind of code I was looking for, thanks ! I will give it a try.

tslearn/metrics.py
rtavenar: OK, but I will not change it in this PR (for readability), let's do that afterwards.

rtavenar:
Ok, so I implemented the suggestion by @GillesVandewiele and it seems to work pretty well, thanks!

Regarding benchmarking, this is quite unsatisfactory at the moment (on my 2 core machine):

```python
import numpy as np
from tslearn.metrics import cdist_dtw, cdist_dtw_dev
import time

n_samples, n_features, n_timestamps = 500, 3, 500
rng = np.random.RandomState(42)
X = rng.randn(n_samples, n_features, n_timestamps)
Y = rng.randn(n_samples, n_features, n_timestamps)

# With the version on this branch
tic = time.time()
mat_new = cdist_dtw(X, Y, n_jobs=2)  # a lot of laptops have 4 cores
toc = time.time()
print("Elapsed time (2 jobs): {}".format(toc - tic))

# With the version on this branch and a single core
tic = time.time()
mat_new_bis = cdist_dtw(X, Y, n_jobs=1)
toc = time.time()
print("Elapsed time (1 job): {}".format(toc - tic))

# With the dev branch
tic = time.time()
mat_ref = cdist_dtw_dev(X, Y)
toc = time.time()
print("Elapsed time (dev branch): {}".format(toc - tic))

np.testing.assert_allclose(mat_new, mat_ref)
np.testing.assert_allclose(mat_new_bis, mat_ref)
```

gives:

```
Elapsed time (2 jobs): 40.6740939617157
Elapsed time (1 job): 21.428598880767822
Elapsed time (dev branch): 2.417752981185913
```

I do not understand why the overhead is so large. Maybe I should start playing with threads, but a simple `backend="threading"` + `nogil=True` option passed to `numba` decorator for the `njit_dtw` function does not seem to help.

GillesVandewiele:
That's strange indeed. Just to make sure, maybe you should increase the number of timeseries and their length/dimensionality to check if there is some point where the multiprocessing starts performing better. If not, there must be something really weird going on..

johannfaouzi:
Maybe try with a larger number of timestamps (it should increase the computation time for each DTW score while keeping the same overhead).

johannfaouzi:
Keep the number of timeseries constant to keep the overhead constant.

rtavenar:
Thanks for the tips.

Now with this code (and latest version with nogil and so on):

```python
import numpy as np
from tslearn.metrics import cdist_dtw, cdist_dtw_dev
import time

n_samples, n_features, n_timestamps = 50, 3, 1000000
rng = np.random.RandomState(42)
X = rng.randn(n_samples, n_features, n_timestamps)
Y = rng.randn(n_samples, n_features, n_timestamps)

# With the version on this branch, no n_jobs specified
tic = time.time()
mat_new = cdist_dtw(X, Y)
toc = time.time()
print("Elapsed time: {}".format(toc - tic))

# With the version on this branch
tic = time.time()
mat_new = cdist_dtw(X, Y, n_jobs=2)  # a lot of laptops have 4 cores
toc = time.time()
print("Elapsed time (2 jobs): {}".format(toc - tic))

# With the version on this branch and a single core
tic = time.time()
mat_new_bis = cdist_dtw(X, Y, n_jobs=1)
toc = time.time()
print("Elapsed time (1 job): {}".format(toc - tic))

# With the dev branch
tic = time.time()
mat_ref = cdist_dtw_dev(X, Y)
toc = time.time()
print("Elapsed time (dev branch): {}".format(toc - tic))

np.testing.assert_allclose(mat_new, mat_ref)
np.testing.assert_allclose(mat_new_bis, mat_ref)
```

I get:

```
Elapsed time: 71.70546102523804
Elapsed time (2 jobs): 45.451483964920044
Elapsed time (1 job): 70.92678189277649
Elapsed time (dev branch): 37.87125778198242
```

So I guess it could be worth trying the same code on a machine with more cores. And if, at some point, we can see some improvement, I would still suggest to fall back to old code for `n_jobs=1` case.

johannfaouzi:
I am not very surprised: there are 2500 DTW scores computed and the total runtime is about 1 minute, so each DTW score takes very little time to be computed. Using parallel computing is more efficient when you have few independent tasks that each takes quite a bit of time. Maybe we should divide the matrix into n rectangles (where n is the number of jobs), and use the cores this way (there would be little overhead in that case, but it’s a bit harder to implement)

johannfaouzi:
Something like this:

```python
import numpy as np


def split_matrix(n_rows, n_jobs):
    """Split a matrix into small matrices.
    
    Parameters
    -------
    n_rows : int
        The number of rows in the matrix.
    
    n_jobs : int
        The number of jobs.
    
    Returns
    -------
    row_indices : array
        Array with the bounds of the indices for the rows.
    
    """
    row_indices = np.linspace(0, n_rows, n_jobs + 1).astype('int64')
    return row_indices
```
Then you could use several jobs with `n_jobs` overheads instead of `n1 * n2`:
```python
matrix = Parallel(n_jobs=n_jobs, backend="threading")(
    delayed(njit_cdist_dtw)(
        dataset1[i:i + 1], dataset2,
        global_constraint=global_constraint,
        sakoe_chiba_radius=sakoe_chiba_radius,
        itakura_max_slope=itakura_max_slope)
    for i in split_matrix(len(dataset1), n_jobs)
)
matrix = np.array(matrix).reshape((len(dataset1), -1))
```

rtavenar:
OK, in fact there was a bug in the code, we were inverting timestamps and modalities, so we had very short time series (3 timestamps). Now that it is fixed, the overhead seems minimal:

```python
import numpy as np
from tslearn.metrics import cdist_dtw, cdist_dtw_no_parallel
import time

n_samples, n_features, n_timestamps = 20, 3, 1000
rng = np.random.RandomState(42)
X = rng.randn(n_samples, n_timestamps, n_features)
Y = rng.randn(n_samples, n_timestamps, n_features)

# With the version on this branch, no n_jobs specified
tic = time.time()
mat_new = cdist_dtw(X, Y)
toc = time.time()
print("Elapsed time: {}".format(toc - tic))

# With the version on this branch
tic = time.time()
mat_new = cdist_dtw(X, Y, n_jobs=2)  # a lot of laptops have 4 cores
toc = time.time()
print("Elapsed time (2 jobs): {}".format(toc - tic))

# With the version on this branch and a single core
tic = time.time()
mat_new_bis = cdist_dtw(X, Y, n_jobs=1)
toc = time.time()
print("Elapsed time (1 job): {}".format(toc - tic))

# With the dev branch
tic = time.time()
mat_ref = cdist_dtw_no_parallel(X, Y)
toc = time.time()
print("Elapsed time (dev branch): {}".format(toc - tic))

np.testing.assert_allclose(mat_new, mat_ref)
np.testing.assert_allclose(mat_new_bis, mat_ref)
```

gives:

```
Elapsed time: 13.200323104858398
Elapsed time (2 jobs): 6.840126991271973
Elapsed time (1 job): 11.459343194961548
Elapsed time (dev branch): 11.609601974487305
```

So I guess we are on a good track :)

Next on my TODO-list:
- `cdist_gak`, `cdist_soft_dtw` in `metrics.py`
- `silhouette_score`, `GlobalAlignmentKernelKMeans`, `TimeSeriesKMeans` in `clustering.py`
- All kNN estimators
- `_kernel_func_gak` in `svm.py`

GillesVandewiele:
That's great news! I think @johannfaouzi his strategy of dividing the matrix in blocks could be a further optimization as well, as it will have less overhead.

rtavenar:
> That's great news! I think @johannfaouzi his strategy of dividing the matrix in blocks could be a further optimization as well, as it will have less overhead.

For sure, but it did not work as such and I wanted to move forward with the simple basic implementation. Typically, if we cannot have @johannfaouzi 's method working in this PR, we could still merge the PR to dev and work on that for a future patch release, since it would not change the API.

In fact, I think if we can have these parallel computations in 0.2 it could be nice since 0.2 already has timings improvements for DTW thanks to @johannfaouzi 's work.

GillesVandewiele:
Alright, maybe I can take a quick look at it :) let me pull the latest code to my computer.

GillesVandewiele:
Do you happen to have the code for `cdist_dtw_no_parallel` ?

rtavenar:
Sorry, Just pushed the code, now you can play with it.

GillesVandewiele:
I have smth up and running, but it's not really performing significantly better than the current implementation.

With `n_samples, n_features, n_timestamps = 20, 3, 500`:
```
Elapsed time (4 jobs): 6.074059009552002
Elapsed time (block & 4 jobs): 5.629055738449097
```

With `n_samples, n_features, n_timestamps = 100, 3, 500`:
```
Elapsed time (4 jobs): 61.467506408691406
Elapsed time (block & 4 jobs): 63.15717124938965
```

With `n_samples, n_features, n_timestamps = 200, 3, 50`:
```
Elapsed time (4 jobs): 15.544788360595703
Elapsed time (block & 4 jobs): 4.916748285293579
```

So it seems to work really well on a larger number of small timeseries (where the computational cost per DTW operation is rather low, bcs there the overhead will be a significant fraction)


CODE:
-----------

```python3
def split_matrix(n_rows, n_jobs):
    """Split a matrix into small matrices.
    Credits to @johannfaouzi.

    Parameters
    -------
    n_rows : int
        The number of rows in the matrix.
    
    n_jobs : int
        The number of jobs.
    
    Returns
    -------
    row_indices : array
        Array with the bounds of the indices for the rows.
    
    """
    row_indices = numpy.linspace(0, n_rows, n_jobs + 1).astype('int64')
    return row_indices


def cdist_dtw_block(dataset1, dataset2=None, global_constraint=None,
                    sakoe_chiba_radius=1, itakura_max_slope=2., n_jobs=None):
    
    dataset1 = to_time_series_dataset(dataset1)

    if dataset2 is None:
        # TODO: Place code below here in this if-clause
        pass
    else:
        dataset2 = to_time_series_dataset(dataset2)
        indices = split_matrix(len(dataset1), n_jobs)
        blocks = Parallel(n_jobs=n_jobs, backend="multiprocessing")(
            delayed(njit_cdist_dtw)(
                dataset1[indices[i]:indices[i + 1], :, :], dataset2,
                global_constraint=global_constraint,
                sakoe_chiba_radius=sakoe_chiba_radius,
                itakura_max_slope=itakura_max_slope)
            for i in range(len(indices) - 1)
        )

        return numpy.vstack(blocks).reshape((len(dataset1), -1))
```

GillesVandewiele:
Here's some more measurements. Seems to work quite nicely when n_timeseries >> n_timestamps!

| n_samples | n_features | n_timestamps | no block implementation | block implementation |
|-----------|------------|--------------|-------------------------|----------------------|
| 200       | 3          | 50           | 15.544                  | 4.916                |
| 200       | 3          | 100          | 17.761                  | 11.129               |
| 50        | 3          | 200          | 6.295                   | 5.226                |
| 50        | 3          | 350          | 10.508                  | 9.625                |
| 50        | 25         | 100          | 4.910                   | 4.031                |
| 50        | 10         | 750          | 38.755                  | 40.51                |

rtavenar:
Thanks for that. Not sure it would be a great idea to maintain both versions. I think we should choose at some point. Also, not sure how to extend the block implem to self-similarity case (blocks based on row indices would have quite heterogeneous computation times).

GillesVandewiele:
I think we just need to replace `dataset2` by `dataset1` for the self-similarity case. Each of those Parallel calls will calculate distances between `1/N_JOBS * len(dataset1)` timeseries of dataset1 and all timeseries of dataset2.

rtavenar:
If we do that, then we no longer use the symmetry in the returned matrix, and hence we compute twice as many distances as we need. This is not a good idea I think.

GillesVandewiele:
You are right... It isn't that trivial as I suggested :(

johannfaouzi:
For the self-similarity case, we could divide the vector of size `n * (n - 1) / 2` into `n_jobs` vectors of (almost) identical sizes (`n * (n - 1) / (2 * n_jobs)`) and use [scipy.spatial.distance.squareform](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.squareform.html) to create the cross-similarity matrix.

rtavenar:
OK, I will leave the discussion about parallelization granularity aside for now.

I tried to do something similar to what we have done for `cdist_dtw` with the GAK kernel.
I tried either with a `numba` re-implementation (code available in the last commit) of `gak` or with the Cython one, but I cannot get reasonable timings. I get things like:

```
Elapsed time: 41.13146114349365
Elapsed time (2 jobs): 19.91181206703186
Elapsed time (1 job): 38.51606512069702
Elapsed time (dev branch): 14.60821795463562
```

where `dev` branch is the current implementation (Cython, no parallelization).

So here, the problem is not related to parallelization overheads but rather to the fact that I can't get a `numba` version that is as fast as the Cython one for some reason. If anyone has any idea why my `numba` implem is inefficient, I would be happy to read it.

rtavenar:
OK, so for DTW (+related algos), I think I have done what I could do (waiting for Travis though).

There a few questions left:
- can we make `gak`'s and `soft_dtw`'s cdist faster too?
- if so, I should make the related estimators `n_jobs`-ready (GAK-kernel-kmeans, kNN+softdtw, kmeans+softdtw)
- should we work on the per-block computations in that PR or implement that later?

rtavenar:
If the answer to 1. is no, then I assume 3. is the only thing to be implemented for this release.

rtavenar:
Oh, and maybe it could be a good idea to an example to the gallery with a kNN search with one or several jobs. But I am not sure we can have several cores on readthedocs :/

rtavenar:
> OK, I will leave the discussion about parallelization granularity aside for now.
> 
> I tried to do something similar to what we have done for `cdist_dtw` with the GAK kernel.
> I tried either with a `numba` re-implementation (code available in the last commit) of `gak` or with the Cython one, but I cannot get reasonable timings. I get things like:
> 
> ```
> Elapsed time: 41.13146114349365
> Elapsed time (2 jobs): 19.91181206703186
> Elapsed time (1 job): 38.51606512069702
> Elapsed time (dev branch): 14.60821795463562
> ```
> 
> where `dev` branch is the current implementation (Cython, no parallelization).
> 
> So here, the problem is not related to parallelization overheads but rather to the fact that I can't get a `numba` version that is as fast as the Cython one for some reason. If anyone has any idea why my `numba` implem is inefficient, I would be happy to read it.

OK, so if I'm a bit less stupid and perform normalisation in the cdist_ function rather than in the `fast_gak` one, things should be faster (by a factor close to 3 I guess, but I still have to check tomorrow on my machine).

If that succeeds, we have a fix for GAK (I still have to finalise the cdist function for self-similarity case: one test should fail for this one). And I could test the `n_jobs` argument passed as a `metric_params` key in GAK kmeans and GAK SVM.

rtavenar:
Ok, ignore my last 2 commits, they make no sense, I will code that properly tomorrow.

rtavenar:
OK, so now I get something reasonable for my `numba` implementation of GAK in terms of timings:

```
Elapsed time: 12.770931959152222
Elapsed time (2 jobs): 6.97777795791626
Elapsed time (1 job): 12.86403512954712
Elapsed time (dev branch): 13.620815992355347
```

If tests pass, I would be happy to get your feedback @johannfaouzi and @GillesVandewiele since I don't plan to work neither on better parallelization granularity nor on numba soft-dtw in coming days, and I would prefer to focus on releasing v0.2

GillesVandewiele:
Hi Romain,

I will try to review either today (this afternoon or tonight) or tomorrow. I'm leaving on holidays though on Wednesday and have some exams to correct, so I'll do my best!

tslearn/clustering.py
```diff
@@ -227,6 +246,9 @@ class GlobalAlignmentKernelKMeans(BaseEstimator, ClusterMixin):
         Bandwidth parameter for the Global Alignment kernel. If set to 'auto',
         it is computed based on a sampling of the training set
         (cf :ref:`tslearn.metrics.sigma_gak <fun-sigmagak>`)
+    n_jobs : int or None (default: None)
```
GillesVandewiele: Perhaps a small note could be made that -1 will use all available cores?

tslearn/clustering.py
```diff
@@ -473,9 +496,16 @@ class TimeSeriesKMeans(BaseEstimator, ClusterMixin,
         Number of iterations for the barycenter computation process. Only used
         if `metric="dtw"` or `metric="softdtw"`.
     metric_params : dict or None
```
GillesVandewiele: Perhaps, for consistency reasons, n_jobs should be taken away from metric_params as well and it should be possible to pass this directly?

tslearn/clustering.py
```diff
@@ -163,16 +174,24 @@ def silhouette_score(X, labels, metric=None, sample_size=None,
     sklearn_metric = None
     if metric_params is None:
         metric_params = {}
+    else:
+        metric_params = metric_params.copy()
+    if "gamma_sdtw" in metric_params.keys():
+        warnings.warn(
```
GillesVandewiele: This is much cleaner than the other warnings.warn I inserted previously (which didn't use the type of Warning and the stacklevel; I was not that familiar with the warning.warn syntax). Perhaps the other warnings could be refactored as well? A simple grep & sed can probably do the trick

tslearn/metrics.py
```diff
@@ -654,6 +655,85 @@ def cdist_dtw(dataset1, dataset2=None, global_constraint=None,
     """
     dataset1 = to_time_series_dataset(dataset1)
 
+    if dataset2 is None:
+        # Inspired from code by @GillesVandewiele:
+        # https://github.com/rtavenar/tslearn/pull/128#discussion_r314978479
+        matrix = numpy.zeros((len(dataset1), len(dataset1)))
```
GillesVandewiele: One nice unit test for this could be to assert that `cdist_dtw(dataset1, dataset1)` is equal to `cdist_dtw(dataset1)`

tslearn/metrics.py
```diff
@@ -740,6 +896,79 @@ def cdist_gak(dataset1, dataset2=None, sigma=1.):
...
+    --------
+    gak : Compute Global Alignment kernel
+
+    References
+    ----------
+    .. [1] M. Cuturi, "Fast global alignment kernels," ICML 2011.
+    """
+    dataset1 = to_time_series_dataset(dataset1)
+
+    if dataset2 is None:
```
GillesVandewiele: Same comment for unit test here

tslearn/neighbors.py
```diff
@@ -115,6 +115,8 @@ class KNeighborsTimeSeries(KNeighborsTimeSeriesMixin, NearestNeighbors):
         <https://docs.scipy.org/doc/scipy/reference/spatial.distance.html>`_.
     metric_params : dict or None (default: None)
         Dictionnary of metric parameters.
+        For metrics that accept parallelization of the cross-distance matrix
+        computations, `n_jobs` is a valid `metric_params` field.
```
GillesVandewiele: Also here, it could be nice to have n_jobs as a direct parameter, and not encapsulate it within metric_params

tslearn/neighbors.py
```diff
@@ -286,6 +296,9 @@ def predict(self, X):
                 X_ = cdist_dtw(X, self._ts_fit, **metric_params)
             elif self._ts_metric == "softdtw":
                 X_ = cdist_soft_dtw(X, self._ts_fit, **metric_params)
+            else:
```
GillesVandewiele: Will this code ever be called? As it is enclosed within `if self.metric in self.variable_length_metrics:`. It is a good check though for in case a metric ever gets added to that list but the corresponding code isn't written here

tslearn/svm.py
```diff
@@ -54,10 +54,15 @@ def _preprocess_sklearn(self, X, y=None, fit_time=False):
             assert self.kernel == "gak"
             self.estimator_kernel_ = "precomputed"
             if fit_time:
-                sklearn_X = cdist_gak(X, X, sigma=numpy.sqrt(self.gamma_ / 2.))
+                sklearn_X = cdist_gak(X,
```
GillesVandewiele: Isn't it better to call it with dataset2=None (double as fast)

tslearn/clustering.py
johannfaouzi: Agree, it is mentioned in scikit-learn (see [sklearn.ensemble.RandomForestClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html) for instance)

tslearn/clustering.py
johannfaouzi: Agree, the docstring for `n_jobs` could mention that it is ignored if the cross-distance matrix cannot be computed using parallelization.

tslearn/neighbors.py
```diff
@@ -320,6 +333,9 @@ def predict_proba(self, X):
...
```
johannfaouzi: Likewise

tslearn/neighbors.py
johannfaouzi: The code was there before the PR, but since it is related I will mention this here.

I think it would be cleaner to have `variable_length_metrics` defined in `metrics.py`, and to make it a constant, like `sklearn.metrics.SCORERS`. Constants are written with all caps, so it should be `VARIABLE_LENGTH_METRICS`.

tslearn/svm.py
```diff
@@ -120,6 +125,10 @@ class TimeSeriesSVC(TimeSeriesSVMMixin, BaseEstimator, ClassifierMixin):
         weights inversely proportional to class frequencies in the input data
         as ``n_samples / (n_classes * np.bincount(y))``
 
+    n_jobs : int or None (default: None)
```
johannfaouzi: Likewise

tslearn/svm.py
```diff
@@ -325,6 +335,10 @@ class TimeSeriesSVR(TimeSeriesSVMMixin, BaseEstimator, RegressorMixin):
     cache_size :  float, optional (default=200.0)
         Specify the size of the kernel cache (in MB).
 
+    n_jobs : int or None (default: None)
```
johannfaouzi: Likewise

tslearn/clustering.py
```diff
@@ -163,16 +174,24 @@ def silhouette_score(X, labels, metric=None, sample_size=None,
     sklearn_metric = None
     if metric_params is None:
         metric_params = {}
+    else:
+        metric_params = metric_params.copy()
```
johannfaouzi: This code is a bit weird (doing a copy of itself): either you create a new object (with a different name), or the copy is unnecessary

tslearn/clustering.py
```diff
@@ -117,16 +118,26 @@ def silhouette_score(X, labels, metric=None, sample_size=None,
         on a random subset of the data.
         If ``sample_size is None``, no sampling is used.
     metric_params : dict or None
-        Parameter values for the chosen metric. Value associated to the
-        `"gamma_sdtw"` key corresponds to the gamma parameter in Soft-DTW.
+        Parameter values for the chosen metric.
```
johannfaouzi: The default values are missing for all the parameters in the docstring

tslearn/clustering.py
```diff
@@ -473,9 +496,16 @@ class TimeSeriesKMeans(BaseEstimator, ClusterMixin,
         Number of iterations for the barycenter computation process. Only used
         if `metric="dtw"` or `metric="softdtw"`.
     metric_params : dict or None
-        Parameter values for the chosen metric. Value associated to the
-        `"gamma_sdtw"` key corresponds to the gamma
+        Parameter values for the chosen metric.
```
johannfaouzi: Some default values are missing in the docstring

tslearn/clustering.py
```diff
@@ -591,15 +629,21 @@ def _fit_one_init(self, X, x_squared_norms, rs):
         return self
 
     def _assign(self, X, update_class_attributes=True):
+        if self.metric_params is None:
+            metric_params = {}
+        else:
+            metric_params = self.metric_params.copy()
```
johannfaouzi: Here it is relevant to make a copy, since both objects are different

tslearn/neighbors.py
```diff
@@ -210,13 +212,21 @@ class (default: 'dtw')
         Metric to be used at the core of the nearest neighbor procedure
     metric_params : dict or None (default: None)
         Dictionnary of metric parameters.
+        For metrics that accept parallelization of the cross-distance matrix
```
johannfaouzi: Likewise

rtavenar:
Thanks a lot for your feedback. 
Your comments really helped improving the code.
I took all your comments into account, I think.
I will now wait to see if tests pass.

tslearn/metrics.py
rtavenar: Please open a new issue on this specific point describing the context and the error you get, so that we can see what could be done.

PR #129, opened 2019-08-19
[WIP] Accept variable-length time series for some pairs metrics/estimators

rtavenar:
This is an attempt to make it possible to use estimators with metrics like DTW on variable-length time series.

The first attempt here is to make DTW/soft-DTW usable for kNN estimators on variable-length time series.

The test I ran is:

```python
from tslearn.neighbors import KNeighborsTimeSeriesClassifier
from tslearn.utils import to_time_series_dataset


X = to_time_series_dataset([[1, 2, 3, 4], [1, 2, 3], [2, 5, 6, 7, 8, 9]])
y = [0, 0, 1]

clf = KNeighborsTimeSeriesClassifier(metric="dtw",
                                     n_neighbors=1,
                                     metric_params={"global_constraint": "sakoe_chiba"})
clf.fit(X, y)
print("---", clf._ts_fit)
print(clf.predict(X))
```

First, we have to think about whether the hack I introduced is a good way to reach our goal and second, once we have chosen a way to proceed, we will have to:
- do the same for other estimators (all those that accept dtw, soft-dtw, gak as metrics, ideally)
- find a way to hack sklearn k-fold variants, since there are some checks for all-finite entries in the datasets there which fail for variable-length time series, if I remember correctly

@GillesVandewiele since you recently worked on making the estimators sklearn-compatible, could you review this PR?

### DIFF
--- a/tslearn/docs/examples/plot_barycenter_interpolate.py
+++ b/tslearn/docs/examples/plot_barycenter_interpolate.py
@@ -18,7 +18,7 @@ import matplotlib.pyplot as plt
 import matplotlib.colors
 
 from tslearn.preprocessing import TimeSeriesScalerMinMax
-from tslearn.barycenters import SoftDTWBarycenter
+from tslearn.barycenters import softdtw_barycenter
 from tslearn.datasets import CachedDatasets
 
 
@@ -69,7 +69,7 @@ for pos in range(2, 25):
     w[1] = (4 - idxr) * idxc / 16
     w[2] = idxr * (4 - idxc) / 16
     w[3] = idxr * idxc / 16
-    plt.plot(SoftDTWBarycenter(weights=w).fit(X_out).ravel(),
+    plt.plot(softdtw_barycenter(X=X_out, weights=w).ravel(),
              color=matplotlib.colors.rgb2hex(get_color(w)),
              linewidth=2)
     plt.xticks([])
--- a/tslearn/neighbors.py
+++ b/tslearn/neighbors.py
@@ -316,3 +316,6 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
             X_ = to_sklearn_dataset(X)
             X_ = check_dims(X_, self._X_fit, extend=False)
         return super(KNeighborsTimeSeriesClassifier, self).predict_proba(X_)
+
+    def _get_tags(self):
+        return {'allow_nan': True}
--- a/tslearn/neighbors.py
+++ b/tslearn/neighbors.py
@@ -335,4 +335,4 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
                          self).predict_proba(X_)
 
     def _get_tags(self):
-        return {'allow_nan': True, 'no_validation': True}
+        return {'allow_nan': True, 'allow_variable_length': True}
--- a/tslearn/neighbors.py
+++ b/tslearn/neighbors.py
@@ -321,7 +321,8 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
                 X_ = cdist_dtw(X, self._ts_fit, **metric_params)
             elif self._ts_metric == "softdtw":
                 X_ = cdist_soft_dtw(X, self._ts_fit, **metric_params)
-            pred = super(KNeighborsTimeSeriesClassifier, self).predict(X_)
+            pred = super(KNeighborsTimeSeriesClassifier,
+                         self).predict_proba(X_)
             self.metric = self._ts_metric
             return pred
         else:
@@ -330,7 +331,8 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
             X = to_time_series_dataset(X)
             X_ = to_sklearn_dataset(X)
             X_ = check_dims(X_, self._X_fit, extend=False)
-            return super(KNeighborsTimeSeriesClassifier, self).predict_proba(X_)
+            return super(KNeighborsTimeSeriesClassifier,
+                         self).predict_proba(X_)
 
     def _get_tags(self):
         return {'allow_nan': True, 'no_validation': True}
--- a/tslearn/neighbors.py
+++ b/tslearn/neighbors.py
@@ -306,7 +306,10 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
         X : array-like, shape (n_ts, sz, d)
             Test samples.
         """
-        if self.metric == "precomputed" and hasattr(self, '_ts_metric'):
+        if self.metric in self.variable_length_metrics:
+            self._ts_metric = self.metric
+            self.metric = "precomputed"
+
             if self.metric_params is None:
                 metric_params = {}
             else:
--- a/tslearn/neighbors.py
+++ b/tslearn/neighbors.py
@@ -259,7 +259,10 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
                                        self._ts_fit.shape[0]))
         else:
             self._X_fit, self._d = to_sklearn_dataset(X, return_dim=True)
-        return super(KNeighborsTimeSeriesClassifier, self).fit(self._X_fit, y)
+        super(KNeighborsTimeSeriesClassifier, self).fit(self._X_fit, y)
+        if hasattr(self, '_ts_metric'):
+            self.metric = self._ts_metric
+        return self
 
     def predict(self, X):
         """Predict the class labels for the provided data
@@ -269,7 +272,10 @@ class KNeighborsTimeSeriesClassifier(KNeighborsTimeSeriesMixin,
         X : array-like, shape (n_ts, sz, d)
             Test samples.
         """
-        if self.metric == "precomputed" and hasattr(self, '_ts_metric'):
+        if self.metric in self.variable_length_metrics:
+            self._ts_metric = self.metric
+            self.metric = "precomputed"
+
             if self.metric_params is None:
                 metric_params = {}
             else:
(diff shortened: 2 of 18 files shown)


rtavenar:
@GillesVandewiele : seems the tests fail because of NaN input values even if I explicitly add the tag `{'allow_nan': True}`: any idea why?

GillesVandewiele:
Hi Romain, I will try to review as soon as I have some time. About those NaN values, the tags were not supported on older sklearn versions yet... To fix those issues, I had to patch the failing subtest of check_estimator (in `sklearn_patches`), which was basically a copy-paste but I would add an if in the beginning of the code:

```
    if _safe_tags(estimator_orig, 'non_deterministic'):
        msg = name + ' is non deterministic'
        raise SkipTest(msg)
```

GillesVandewiele:
I just remembered that for these specific NaN cases, I could get around without having to patch all the functions... On older sklearn versions, they had a `ALLOW_NAN` array in `sklearn.check_estimator.py`.

So in `tslearn.test_estimators.py` we can do:
```
if hasattr(checks, 'ALLOW_NAN'):
    checks.ALLOW_NAN.append('TimeSeriesKMeans')
```
You'll just have to append the KNN as well

GillesVandewiele:
I think we will have to remove the check_dims all, or pass X_fit=None to it to support variable length ts

rtavenar:
> I think we will have to remove the check_dims all, or pass X_fit=None to it to support variable length ts

Probably.

What about adding our own tag (eg. `'allow_variable_length'`) and skip the `check_dims` in the following loop if that tag is True?

```python
    for check in _yield_checks(name, estimator):
        yield check
```

GillesVandewiele:
that would be nice to have indeed! get_tags is reimplemented for all python versions in `sklearn_patches`, so the tests that assert for a ValueError when variable sized data is passed can be skipped there. It is rather weird that it fails on `check_estimators_dtypes` though...

rtavenar:
> that would be nice to have indeed! get_tags is reimplemented for all python versions in `sklearn_patches`, so the tests that assert for a ValueError when variable sized data is passed can be skipped there. It is rather weird that it fails on `check_estimators_dtypes` though...

I removed my comment since it seems it fails because of *my* call to `check_dims` in `predict_proba`.

rtavenar:
It was my mistake: I changed `predict` but not `predict_proba`.

GillesVandewiele:
So those are the tests that assert that ValueError I was speaking about... Seems like it happens in a function that is already patched, and that a tag `no_validation` exists to skip them already :)

GillesVandewiele:
I added the following as an edit, but is perhaps better as a standalone comment:

> `no_validation` will remove more than just checking (such as checking that you don't pass continuous y-data to a classifier, or complex data, ...) if the dimensions passed to fit and predict match though, so a custom tag would be a bit nicer, but more work...

rtavenar:
> I added the following as an edit, but is perhaps better as a standalone comment:
> 
> > `no_validation` will remove more than just checking (such as checking that you don't pass continuous y-data to a classifier, or complex data, ...) if the dimensions passed to fit and predict match though, so a custom tag would be a bit nicer, but more work...

Thanks for the notice.
I think I will definitely have to create a new `"allow_variable_length"` tag and dig in the tests for which ones should be skipped if this tag is on.

GillesVandewiele:
OK, seems to be building now :). 

About the custom tag, you would just have to add it to `_DEFAULT_TAGS` in the `sklearn_patches` file, and add `and not tags['allow_variable_length']` to the if-statements in `check_classifiers_train` (`if not tags["no_validation"] and not tags['allow_variable_length']:`). 

Not sure if it will raise errors on other tests as well, if that would be the case, these will have to be patched...

rtavenar:
It seems better, yes.

For the tags, I would like to skip only those specific tests that are related to variable-length issues, so it will take some time, but seems doable.

GillesVandewiele:
Based on the error msg they throw, I would start with these 2 lines:

https://github.com/rtavenar/tslearn/blob/dev-variablelength/tslearn/tests/sklearn_patches.py#L301
https://github.com/rtavenar/tslearn/blob/dev-variablelength/tslearn/tests/sklearn_patches.py#L348

rtavenar:
Thanks, tests with the new tag now pass on my machine, let's see on Travis.

GillesVandewiele:
OK great, I'll check the code some time later, but nice that everything is passing already :)

rtavenar:
Update: I was too pessimistic! Cross validation now works without any problem on variable-length time series (for kNN classifier). So I won't have to hack that part. Will now focus on checking each estimator listed in the variable-length doc page and upgrading it if necessary.

rtavenar:
OK, I've done something that seems to work on my side, let's see what it gives on Travis. I had to hack once again the GAK kernel, so we do not need the related class anymore if that succeeds.

tslearn/neighbors.py
```diff
@@ -243,12 +271,31 @@ def predict(self, X):
...
+            return pred
+        else:
+            check_is_fitted(self, '_X_fit')
+            X = check_array(X, allow_nd=True)
+            X = to_time_series_dataset(X)
+            X_ = to_sklearn_dataset(X)
+            X_ = check_dims(X_, self._X_fit, extend=False)
+            return super(KNeighborsTimeSeriesClassifier, self).predict(X_)
 
     def predict_proba(self, X):
```
GillesVandewiele: It looks like the predict and predict_proba functions share a lot of common code. Perhaps it would be nice to create a private function and let predict & predict_proba re-use its code?

tslearn/svm.py
```diff
@@ -227,48 +232,82 @@ def fit(self, X, y, sample_weight=None):
         return self
 
     def predict(self, X):
```
GillesVandewiele: similar comment about common code between predict, predict_proba, ...

tslearn/svm.py
```diff
@@ -197,25 +199,28 @@ def support_vectors_time_series_(self, X):
...
-
-        gamma = self.gamma
-        kernel = self.kernel
-        if gamma == "auto":
-            gamma = gamma_soft_dtw(to_time_series_dataset(X))
-        if kernel == "gak":
-            kernel = GAKKernel(sz, d, gamma)
+        if self.kernel in self.variable_length_kernels:
+            assert self.kernel == "gak"
+            estimator_kernel = "precomputed"
```
GillesVandewiele: Can't we do something similar as here in neighbors.py? Instead of overwriting self.metric and then resetting it before the return statement, a new variable `kernel_metric` could be made and that could be passed to cdist

tslearn/svm.py
GillesVandewiele: e.g. isn't predict_log_proba = np.log(predict_proba) ?

tslearn/docs/examples/plot_barycenter_interpolate.py
```diff
@@ -69,7 +69,7 @@ def get_color(weights):
     w[1] = (4 - idxr) * idxc / 16
     w[2] = idxr * (4 - idxc) / 16
     w[3] = idxr * idxc / 16
-    plt.plot(SoftDTWBarycenter(weights=w).fit(X_out).ravel(),
+    plt.plot(softdtw_barycenter(X=X_out, weights=w).ravel(),
```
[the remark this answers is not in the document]
rtavenar: In fact the deprecation for that is old. And I don't think having barycenters in a pipeline would make sense, which is why I chose to deprecate.

tslearn/neighbors.py
rtavenar: Good idea! Will do.

tslearn/svm.py
rtavenar: We should let sklearn/libsvm deal with that I would say.

tslearn/svm.py
rtavenar: Well, the issue is that for kNN we do not have an external estimator that we fit, but instead we fit `self`.

rtavenar:
Not sure why tests fail, maybe will have to dig in clustering checks, but I dont know why kmeans passes...

GillesVandewiele:
It apparently fails on a test that checks if a ValueError is raised when input data with NaNs or infinites is provided... Does the class have an `allow-nan` tag?

rtavenar:
OK, the reason was that I did not register that class in `test_all_estimators.py` as allowing NaN values. Since I would expect that such issues might happen again, I made this registration automatic.

Let's see what Travis says, at least the tests pass for kernel KMeans and SVM classes on my machine.

rtavenar:
OK, tests seem to pass.
So now, I would like to add a few more tests to `test_variablelength.py` and then merge. Let me know what you think @GillesVandewiele .

GillesVandewiele:
Great change to skip NaN tests with a tag! The test_variablelength could indeed be extended with more classifiers, and perhaps also to check if they work with e.g. sklearn.Pipeline or GridSearch. Also some other tests could be useful:
* the number of dimensions per timeseries (the final element in X.shape) probably still have to match, so we could add a test that checks if some error is raised when this is not the case
* test with more metrics (e.g. softdtw as well)
* ...

Perhaps some example in the docs could be adjusted to show this capability as well?

rtavenar:
OK, I will do that, but probably not today nor tomorrow.

Thanks for your help, and this PR was the occasion for me to see how much your work helped for the development of tslearn estimators. I am grateful for that.

GillesVandewiele:
Thanks! It can be a hassle at times to make the tests pass, but they are helpful in detecting bugs I'd say :).

It should also be noted that it was a little bit of an overkill (not all tests are necessary in order to make it work with GridSearch or Pipeline, but these tests could be useful for other sklearn utilities that I do not know of and can be seen as "good API usability tests").

rtavenar:
I found that default `sigma` value for kernel-kmeans was not consistent with default `gamma` for SVMs, so I changed it. Other changes are improvements in the tests. We could do better of course, but I guess that's a start, at least, and I would like to merge soon if possible.

Do you think that version is reasonable @GillesVandewiele ?

GillesVandewiele:
I reviewed the code around noon. Not sure if there are any significant changes you want me to take another look at? If not, LGTM! :)

PR #133, opened 2019-08-27
Fix documentation for dtw_path

felixdivo:
The values allowed for the parameter `global_constraint` were probably changed somewhere down the road, and the parameter value in `itakura_max_slope` was not updated accordingly.


rtavenar:
Hi @felixdivo 

Thanks for this patch, indeed the doc was wrong.

felixdivo:
You're welcome. :smile:

PR #135, opened 2019-08-28
Replace implicit imports with explicit imports

### LINKED ISSUE #134: Very low code coverage for sklearn_patches.py
Codeclimate shows that [sklearn_patches.py](https://codeclimate.com/github/rtavenar/tslearn/tslearn/tests/sklearn_patches.py/source) has only 13% code coverage. Since it is in the `tests` folder, it should have 100% code coverage. 

A lot of code is currently never used (all the functions, the current code coverage comes mostly from the imports). You can check the `coverage`box to see which lines are never run (red lines).

johannfaouzi:
Fixes #134 

As title says, the implicit imports are replaced with explicit imports in `test_estimators.py`.
It was a bit hard to find some of them from scikit-learn. Let's see if it improves code coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,12 @@ script:
 - python -c "import keras.backend";
   sed -i -e 's/"backend":[[:space:]]*"[^"]*/"backend":\ "'$KERAS_BACKEND'/g' ~/.keras/keras.json;
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
-- if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then
+- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then
   python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
-  if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then codeclimate-test-reporter; fi
-  else travis_wait 40 python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE; fi
+  else python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
+
+after_success:
+- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then codeclimate-test-reporter
+
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,8 @@ script:
   sed -i -e 's/"backend":[[:space:]]*"[^"]*/"backend":\ "'$KERAS_BACKEND'/g' ~/.keras/keras.json;
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
 - if [ "$NUMBA_DISABLE_JIT" == 1 ];
-  then python -m pytest -v --pyargs tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
-  else python -m pytest -v --pyargs tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
+  then python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
+  else python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
   fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,9 +56,8 @@ script:
   sed -i -e 's/"backend":[[:space:]]*"[^"]*/"backend":\ "'$KERAS_BACKEND'/g' ~/.keras/keras.json;
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
 - if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then
-  python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py --ignore tslearn/tests/test_estimators.py $KERAS_IGNORE --cov=tslearn;
+  python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
   if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then codeclimate-test-reporter; fi
   else travis_wait 40 python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE; fi
 notifications:
   email: false
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,12 +55,13 @@ script:
 - python -c "import keras.backend";
   sed -i -e 's/"backend":[[:space:]]*"[^"]*/"backend":\ "'$KERAS_BACKEND'/g' ~/.keras/keras.json;
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
-- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then
-  python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
+- if [ "$NUMBA_DISABLE_JIT" == 1 ];
+  then python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
   else python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
+  fi 
 
 after_success:
-- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then codeclimate-test-reporter
+- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then codeclimate-test-reporter fi
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,9 +56,9 @@ script:
   sed -i -e 's/"backend":[[:space:]]*"[^"]*/"backend":\ "'$KERAS_BACKEND'/g' ~/.keras/keras.json;
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
 - if [ "$NUMBA_DISABLE_JIT" == 1 ];
-  then python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
-  else python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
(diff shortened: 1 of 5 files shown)


johannfaouzi:
Well it does not. I will try to replace them in `sklearn_patches.py` too.

rtavenar:
For some reason, test_estimators.py is not listed on the coverage report, which probably explains why sklearn_patches.py is almost not covered. It is unclear to me why test_estimators is ignored

rtavenar:
Well, it is explicitly ignored in travis. Im not sure why, though. Could you check in this direction ?

johannfaouzi:
That may be a better explanation. I will try to remove it and see how things go.

johannfaouzi:
With the current travis configuration file, the doctests are not run, because it is specified to run only the tests in the `tests` folder: `pytest -v tslearn tslearn/tests/`

It seems like an unintended behavior imho.

johannfaouzi:
The clustering algorithms do not pass the tests. I am a bit tired, so I may say something stupid, but `predict` returns the argmin of the cross-similarity matrix:
https://github.com/rtavenar/tslearn/blob/eb37c4fbebc2e575c16a955c4bb895cf016faf29/tslearn/clustering.py#L1102

So an array of non-negative integers. So the smallest label can never be -1, and may be 1 when there is noise?
https://github.com/rtavenar/tslearn/blob/eb37c4fbebc2e575c16a955c4bb895cf016faf29/tslearn/tests/sklearn_patches.py#L117-L118

johannfaouzi:
Never mind, it seems like the tests pass for some configurations, and lead to a more reasonable code coverage. A bit weird...

johannfaouzi:
`test_all_estimators` should be refactored with `pytest.mark.parametrize`: it would make the output cleaner and it would avoid the runtime issues.

rtavenar:
OK, so the failing tests are related to clustering algorithms that do not assign data to every possible cluster. Maybe adding a bit more samples to the dataset could make this more stable?

And I agree with your remark on `pytest.mark.parametrize`.

I can try to implement fixes, but I am not sure I can push to your master. Let me know what is best for you.

johannfaouzi:
I will try to fix this and get back to you later on.

rtavenar:
Locally, if I just set the number of time series per class (for the clustering tests only) to 15, all tests pass.

johannfaouzi:
> Locally, if I just set the number of time series per class (for the clustering tests only) to 15, all tests pass.

Did you change the number of time series in `_create_small_ts_dataset()`?

johannfaouzi:
My code with `pytest.mark.parametrize` doesn't seem to work on Python 2.7...

One issue I faced during this PR: a lot of the functions used in `sklearn_patches.py` and `test_estimators.py` are not part of the public API of scikit-learn. I have the master branch installed and some functions were removed. I copy pasted them, but overall I don't think that it will be easy to maintain them, because these functions can be removed without any deprecation warning.

rtavenar:
> > Locally, if I just set the number of time series per class (for the clustering tests only) to 15, all tests pass.
> 
> Did you change the number of time series in `_create_small_ts_dataset()`?

I did the following :

```python
def _create_small_ts_dataset(n_ts_per_class=5):
    return random_walk_blobs(n_ts_per_blob=n_ts_per_class, n_blobs=3,
                             random_state=1, sz=10, noise_level=0.025)
```

johannfaouzi:
So I removed the `@ignore_warnings()` decorator, which enables the code to run on Python 2. KShape test fails on Python 2 only, with the predicted clusters being [0, 2] instead of [0, 1, 2]...

https://travis-ci.org/rtavenar/tslearn/jobs/578251283#L822-L836

rtavenar:
> One issue I faced during this PR: a lot of the functions used in `sklearn_patches.py` and `test_estimators.py` are not part of the public API of scikit-learn. I have the master branch installed and some functions were removed. I copy pasted them, but overall I don't think that it will be easy to maintain them, because these functions can be removed without any deprecation warning.

Argh, this is an important point indeed. Not sure how to deal with it in the future...

rtavenar:
Seems like we have an issue here: There are failing tests but the Travis indicator for those jobs is green :/

johannfaouzi:
My code for `pyts` is slightly different, I don't know if that would solve this:
https://github.com/johannfaouzi/pyts/blob/5b4310f581af774eb9d4575c0e40251ef707b5c8/.travis.yml#L32-L39

johannfaouzi:
I know nothing about bash, but the `if else` loop seems weird, doesn't it ?
https://travis-ci.org/rtavenar/tslearn/jobs/578251283#L1029

Unix always exits with the status of the last command in the pipeline, so if there are several commands and the last one succeeds, it doesn't matter that previous commands failed.

johannfaouzi:
Do you agree with the way the noise is added @rtavenar ?
https://github.com/rtavenar/tslearn/blob/eb37c4fbebc2e575c16a955c4bb895cf016faf29/tslearn/tests/sklearn_patches.py#L78

Right now it adds new points that are noise, instead of adding noise to existing points.

rtavenar:
> Do you agree with the way the noise is added @rtavenar ?
> 
> https://github.com/rtavenar/tslearn/blob/eb37c4fbebc2e575c16a955c4bb895cf016faf29/tslearn/tests/sklearn_patches.py#L78
> 
> Right now it adds new points that are noise, instead of adding noise to existing points.

Hum, this is not noise imo, this is a new dataset that includes the previous one. Not sure why we should do that instead of adding white noise to the original dataset.

rtavenar:
> My code for `pyts` is slightly different, I don't know if that would solve this:
> https://github.com/johannfaouzi/pyts/blob/5b4310f581af774eb9d4575c0e40251ef707b5c8/.travis.yml#L32-L39

I agree that doing it this way is probably better (and is much more readable), and I guess the exit code we get is the one from the command that sends the coverage report rather than the one that corresponds to the tests themselves.

.travis.yml
```diff
@@ -55,9 +55,12 @@ script:
...
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
-- if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then
+- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then
   python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
-  if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then codeclimate-test-reporter; fi
-  else travis_wait 40 python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE; fi
+  else python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
+
+after_success:
+- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then codeclimate-test-reporter
```
rtavenar: Isn't there a `fi;` missing here?

.travis.yml
```diff
@@ -55,10 +55,12 @@ script:
 - python -c "import keras.backend";
   sed -i -e 's/"backend":[[:space:]]*"[^"]*/"backend":\ "'$KERAS_BACKEND'/g' ~/.keras/keras.json;
   echo -e "Running tests with the following config:\n$(cat ~/.keras/keras.json)";
-- if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then
-  python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py --ignore tslearn/tests/test_estimators.py $KERAS_IGNORE --cov=tslearn;
-  if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then codeclimate-test-reporter; fi
-  else travis_wait 40 python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE; fi
+- if [ "$NUMBA_DISABLE_JIT" == 1 ]; then
+  python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE --cov=tslearn;
+  else python -m pytest -v tslearn tslearn/tests/ --doctest-modules --ignore tslearn/docs --ignore tslearn/deprecated.py $KERAS_IGNORE;
```
rtavenar: Another `fi;` missing here, no?

johannfaouzi: Yes, I'm a noob at bash :|

tslearn/tests/sklearn_patches.py
```diff
@@ -75,7 +130,8 @@ def check_clustering(name, clusterer_orig, readonly_memmap=False):
     X, y = _create_small_ts_dataset()
     X, y = shuffle(X, y, random_state=7)
     X = TimeSeriesScalerMeanVariance().fit_transform(X)
-    X_noise = np.concatenate([X, random_walks(n_ts=2, sz=10)])
+    rng = np.random.RandomState(42)
+    X_noise = X + (rng.randn(*X.shape()) / 5)
```
rtavenar: `*X.shape` :)

johannfaouzi: Skill display 101 today :(

rtavenar:
By the way, @johannfaouzi : all your commits are assigned to `johann.faouzi` (unknown github user) and hence you do not appear as a tslearn contributor, which is a pity

johannfaouzi:
> By the way, @johannfaouzi : all your commits are assigned to `johann.faouzi` (unknown github user) and hence you do not appear as a tslearn contributor, which is a pity

Thanks! I was wondering why, and the issue is that I was using another ID (because I once worked on GitLab and had to use my email from the institute I work in) and never changed it back to my GitHub ID...

johannfaouzi:
I tried to ignore warnings by modifying the `setup.cfg` file but it didn't work. All the tests are passing (I think), but the output is ugly. I had to remove the `@ignore_warnings` decorator because it wasn't working well with `@pytest.mark.parametrize` for Python 2.7.

Let me know what you think about the current state of the PR.

rtavenar:
Have you tried options described [there](https://docs.pytest.org/en/latest/warnings.html) ?

Like, for example, would it be a problem to disable warnings while running tests using `--disable-warnings`? Or use the dedicated decorator [`@pytest.mark.filterwarnings`](https://docs.pytest.org/en/latest/reference.html#pytest-mark-filterwarnings)?

johannfaouzi:
Thanks for the links @rtavenar! I think that it works quite well right now. I disabled:
* `DeprecationWarning`,
* `PendingDeprecationWarning`, 
* `sklearn.exceptions.SkipTestWarning`, and
* `UserWarning`

There are a few `RunTimeWarning`s for Python 3.5 that could be worth looking at (but out of scope for this PR), and I'm not sure why it's happening for this version only.

tslearn/tests/sklearn_patches.py
```diff
@@ -114,8 +170,8 @@ def check_clustering(name, clusterer_orig, readonly_memmap=False):
     assert_array_equal(labels_sorted, np.arange(labels_sorted[0],
                                                 labels_sorted[-1] + 1))
 
-    # Labels are expected to start at 0 (no noise) or -1 (if noise)
-    assert labels_sorted[0] in [0, -1]
+    # Labels are expected to start at 0 (no noise) or 1 (if noise)
```
rtavenar: I quite don't understand why you changed this statement, and I have to admit I also don't understand why assigned clusters could be equal to -1. Could you give a bit of insight on this?

johannfaouzi: I agree that a predicted label can never be -1 (since the predicted label is the argmin). I just changed what I thought to be a typo, but even with noise I think that we should expect to have at least one sample for each original cluster.

I would remove this line and replace
https://github.com/rtavenar/tslearn/blob/eb37c4fbebc2e575c16a955c4bb895cf016faf29/tslearn/tests/sklearn_patches.py#L114-L115

with
```python
assert_array_equal(labels_sorted, np.arange(0, 3))
```

rtavenar: Makes sense to me (though maybe for some clustering methods, labels could be computed without resorting to an argmin, but I still wouldn't understand why it could be -1).

GillesVandewiele: Not all clustering algorithms assign each points to a cluster (and are more robust to noise). E.g. DBSCAN. The non-assigned samples get -1

johannfaouzi: I didn't know that! However I don't think that any of the clustering algorithm currently available in this package do that, so the changes are acceptable imho. A good thing to keep in mind for the future though :)

Issue #137, opened 2019-08-29
Feature Request: Adjustable start point for sDTW path calculation


AKuederle:
With the new implementation of  sDTW, the cost matrix is now calculated first (`njit_accumulated_matrix_subsequence`) and in a second step the path is calculated (`_return_path_subsequence`).

During the path calculation the minimum of the last row (i.e. the best match) is used as starting point. However, there are applications, where not only the best possible path, but the position of potential matches are of interest (e.g. template matching applications). For these cases it would be awesome to be able to specify the starting point of the path calculation:

```
@njit()
def _return_path_subsequence(acc_cost_mat, start_index):
    ...
```

This would move calculation of the argmin to `dtw_subsequence_path`, which would then pass the result as argument to `_return_path_subsequence`.

I would be open to create a small pull request for this feature. In this context I would also propose to make `_return_path_subsequence` public API (i.e. removing the underscore and documenting it).

rtavenar:
Dear @AKuederle 

First, thanks for your interest and the kind words you wrote about `tslearn` :)

Then, I understand that such a feature can be useful, but I guess it has applications in very limited cases. Could you elaborate on typical settings in which you would want to do that instead of, for example `dtw(s1, s2[:start_index])` ?

AKuederle:
Thank you for your fast response!

DTW or rather sDTW is widely used as a template matching algorithm in some fields. For example to find the timepoints of all steps in a movement signal or of all words in a speech signal. In these applications, you would produce a template for the part of the signal you are searching for (e.g. a single step), then calculate the accumulated cost matrix between you long signal and your template. Because you expect multiple occurrences of your template in the signal, the last row of you cost matrix will have local minima (at the specific positions, where your template would match well). You would then extract these matches using some sort of local minima algorithms and then calculate the warping path starting from each of the matches you found.
To give you an example here is a paper using this method to extract strides from IMU signals: [Barth2013](https://ieeexplore.ieee.org/document/6611104).

Because the goal is to find multiple matches, `dtw(s1, s2[:start_index])` would recalculate the cost matrix multiple times, which is not necessary. What I have in mind would something like this:

```python
from tslearn import metrics
cost_mat = metrics.njit_accumulated_matrix_subsequence(subseq, longseq)
matches = find_local_minima(cost_mat[-1, :])  # find_local_minima will depend on the application
for m in matches:
     path = _return_path_subsequence(acc_cost_mat, m)
     ...  # Do something with the path
```

[...]

PR #138, opened 2019-08-29
[MRG] Configurable subsequent DTW path computation

AKuederle:
See #137

### DIFF
--- a/tslearn/docs/examples/plot_sdtw.py
+++ b/tslearn/docs/examples/plot_sdtw.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""
+r"""
 sDTW multi path matching
 ========================
 
@@ -8,7 +8,7 @@ matches of a sequence in a longer sequence.
 
 A potential usecase is to identify the occurrence of certain events in
 continuous sensor signals. As one example Barth et al. [1] used this method
-to find stride in sensor recordings of gait
+to find stride in sensor recordings of gait.
 
 The example demonstrates the use of the functions
 `subsequence_cost_matrix` and `subsequence_path`
@@ -16,8 +16,9 @@ to manually calculate warping paths from multiple potential alignments. If
 you are only interested in finding the optimal alignment, you can directly use
 `dtw_subsequence_path`.
 
-[1] Barth, et al. (2013). Subsequence dynamic time warping as a method for
-    robust step segmentation using gyroscope signals of daily life activities,
+[1] Barth, et al. (2013):
+    Subsequence dynamic time warping as a method for robust step segmentation
+    using gyroscope signals of daily life activities,
     EMBS, https://doi.org/10.1109/EMBC.2013.6611104
 
 """
--- a/tslearn/docs/examples/plot_sdtw.py
+++ b/tslearn/docs/examples/plot_sdtw.py
@@ -11,7 +11,7 @@ continuous sensor signals. As one example Barth et al. [1] used this method
 to find stride in sensor recordings of gait
 
 The example demonstrates the use of the functions
-`accumulated_subsequence_cost_matrix` and `calculate_subsequence_path`
+`subsequence_cost_matrix` and `subsequence_path`
 to manually calculate warping paths from multiple potential alignments. If
 you are only interested in finding the optimal alignment, you can directly use
 `dtw_subsequence_path`.
@@ -52,8 +52,8 @@ print('Shape long sequence: {}'.format(long_sequence.shape))
 print('Shape short sequence: {}'.format(short_sequence.shape))
 
 # Calculate the accumulated cost matrix
-mat = metrics.accumulated_subsequence_cost_matrix(short_sequence,
-                                                  long_sequence)
+mat = metrics.subsequence_cost_matrix(short_sequence,
+                                      long_sequence)
 
 # Calculate cost function
 cost_func = mat[-1, :]
@@ -63,7 +63,7 @@ cost_func = mat[-1, :]
 potential_matches = find_peaks(-cost_func, distance=sz * 0.75, height=-50)[0]
 
 # Calculate the optimal warping path starting from each of the identified minima
-paths = [metrics.calculate_subsequence_path(mat, match) for match in
+paths = [metrics.subsequence_path(mat, match) for match in
          potential_matches]
 
 plt.figure(1, figsize=(6 * n_repeat, 6))
--- a/tslearn/docs/examples/plot_sdtw.py
+++ b/tslearn/docs/examples/plot_sdtw.py
@@ -1,0 +1,104 @@
+# -*- coding: utf-8 -*-
+"""
+sDTW multi path matching
+========================
+
+This example illustrates how subsequent DTW can be used to find multiple
+matches of a sequence in a longer sequence.
+
+A potential usecase is to identify the occurrence of certain events in
+continuous sensor signals. As one example Barth et al. [1] used this method
+to find stride in sensor recordings of gait
+
+The example demonstrates the use of the functions
+`accumulated_subsequence_cost_matrix` and `calculate_subsequence_path`
+to manually calculate warping paths from multiple potential alignments. If
+you are only interested in finding the optimal alginement, you can directly use
+`dtw_subsequence_path`.
+
+[1] Barth, et al. (2013). Subsequence dynamic time warping as a method for
+    robust step segmentation using gyroscope signals of daily life activities,
+    EMBS, https://doi.org/10.1109/EMBC.2013.6611104
+
+"""
+
+# Author: Arne Kuederle
+# License: BSD 3 clause
+
+import matplotlib.pyplot as plt
+import numpy
+from scipy.signal import find_peaks
+
+from tslearn import metrics
+from tslearn.generators import random_walks
+from tslearn.preprocessing import TimeSeriesScalerMeanVariance
+
+numpy.random.seed(0)
+n_ts, sz, d = 2, 100, 1
+n_repeat = 5
+dataset = random_walks(n_ts=n_ts, sz=sz, d=d)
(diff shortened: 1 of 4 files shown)


tslearn/tests/test_metrics.py
```diff
@@ -42,6 +43,24 @@ def test_dtw_subseq():
...
+    subseq = to_time_series(subseq)
+    longseq = to_time_series(longseq)
+    cost_matrix = tslearn.metrics.njit_accumulated_matrix_subsequence(subseq, longseq)
+
+    path = tslearn.metrics.return_path_subsequence(cost_matrix, 3)
+    np.testing.assert_equal(path, [(0, 2), (1, 3)])
+
+    path = tslearn.metrics.return_path_subsequence(cost_matrix, 1)
+    np.testing.assert_equal(path, [(0, 0), (1, 1)])
+
```
rtavenar: Maybe adding a test that asserts that on some randomly-generated data, `dtw_subsequence_path` gives the same path / cost as `return_path_subsequence` when starting from the argmin?

rtavenar:
As stated in #137, this use case should be properly documented as I am unsure how straight forward it can be for an end user to understand it properly. I plan to improve on the current example for DTW path soon, maybe both notebooks could share a more or less common appearance, something like [this example](https://tslearn.readthedocs.io/en/latest/_images/dtw.png)?

Also, I am unsure of why a Travis job was failing: I restarted the corresponding job to make sure this is reproducible (it was an apparently unrelated test that was failing, so...)

Finally, this PR will change the public API for `tslearn` (without regression, though), so it should be shipped with `tslearn` 0.3. I will prepare a `dev` branch asap for all the improvements to be included in `tslearn` 0.3, and this PR should be merged to that `dev` branch instead of the `master` branch then.

AKuederle:
Thank your for your comment! I will look into it and update the pull request!

rtavenar:
OK, I prepared a new DTW visualization example for the docs in `dev` branch, cf [here](https://github.com/rtavenar/tslearn/blob/dev/tslearn/docs/examples/plot_dtw.py).

Maybe you could use more or less the same kind of visualization but:
* Use `tslearn` function to compute cross-similarity matrix (and update my example so that it does the same, since the function becomes public with your PR) instead of `cdist` from `scipy`;
* have several paths exhibited for several end points.

What do you think?

AKuederle:
Sounds good! Btw. should we drop the njit prefix for the functions, which will become public API. I guess the naming was introduced, because your transitioned to numba from Cython, but for public API, this might be confusing.

rtavenar:
Yep! Good idea!

tslearn/metrics.py
```diff
@@ -400,7 +433,7 @@ def dtw_subsequence_path(subseq, longseq):
 
     Compared to traditional DTW, here, border constraints on admissible paths
     :math:`P` are relaxed such that :math:`P_0 = (0, ?)` and
-    :math:`P_L = (N-1, ?)` where :math:`L` is the length of the considered path
+    :math:`P_0 = (0, ?)` where :math:`L` is the length of the considered path
```
rtavenar: Hum, I don't agree with your edit. I think the original formulation was the correct one.

tslearn/metrics.py
```diff
@@ -362,10 +361,44 @@ def njit_accumulated_matrix_subsequence(subseq, longseq):
 
 
 @njit()
-def _return_path_subsequence(acc_cost_mat):
+def calculate_subsequence_path(acc_cost_mat, idx_path_end):
```
rtavenar: `calculate_subsequence_path` is a bit of a complicated name for public API no? What about `subsequence_path`?

rtavenar:
> I added the example and a little bit of explanation. Do you think this is sufficient to understand the usecase?

Yep, I think it helps a lot!

tslearn/docs/examples/plot_sdtw.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+sDTW multi path matching
+========================
+
+This example illustrates how subsequent DTW can be used to find multiple
+matches of a sequence in a longer sequence.
+
+A potential usecase is to identify the occurrence of certain events in
+continuous sensor signals. As one example Barth et al. [1] used this method
+to find stride in sensor recordings of gait
```
rtavenar: Add "." at the end of the sentence.

tslearn/docs/examples/plot_sdtw.py
```diff
@@ -0,0 +1,105 @@
...
+
+The example demonstrates the use of the functions
+`accumulated_subsequence_cost_matrix` and `calculate_subsequence_path`
+to manually calculate warping paths from multiple potential alignments. If
+you are only interested in finding the optimal alignment, you can directly use
+`dtw_subsequence_path`.
+
+[1] Barth, et al. (2013). Subsequence dynamic time warping as a method for
```
rtavenar: Looks like the reference here does not render nicely. From `sklearn` docs (eg [here](https://scikit-learn.org/stable/auto_examples/applications/plot_species_distribution_modeling.html#sphx-glr-auto-examples-applications-plot-species-distribution-modeling-py), it seems that using bullet lists for references is more convenient.

tslearn/metrics.py
```diff
@@ -362,10 +361,44 @@ def njit_accumulated_matrix_subsequence(subseq, longseq):
...
+    idx_path_end: int
+        The end position of the matched subsequence in the longer sequence.
+
+    Returns
+    -------
+    path: list of tuples of integer pairs
+        Matching path represented as a list of index pairs. In each pair, the
+        first index corresponds to `subseq` and the second one corresponds to
+        `longseq`. The startpoint of the Path is :math:`P_0 = (0, ?)` and it
+        ends at :math:`P_0 = (0, idx_path_end)`
```
rtavenar: Again, I disagree (cf comment above): the last element in the path should be of the form `(sz1-1, ?)`

rtavenar:
Also you should add `accumulated_subsequence_cost_matrix` and `calculate_subsequence_path` to the list of documented items in metrics module, which is defined in `tslearn/docs/gen_modules/tslearn.metrics.rst`

tslearn/docs/examples/plot_sdtw.py
AKuederle: This does not seem to be consistent. At other instances they used the same citation style I used (e.g. [here](https://github.com/scikit-learn/scikit-learn/blob/master/examples/covariance/plot_mahalanobis_distances.py)

tslearn/docs/examples/plot_sdtw.py
AKuederle: Done

tslearn/metrics.py
AKuederle: I renamed them to `subsequence_path` and `subsequence_cost_matrix`

AKuederle:
> Also you should add `accumulated_subsequence_cost_matrix` and `calculate_subsequence_path` to the list of documented items in metrics module, which is defined in `tslearn/docs/gen_modules/tslearn.metrics.rst`

Done

tslearn/docs/examples/plot_sdtw.py
rtavenar: You are perfectly right. My main concern in fact was about the obtained formatting: in the case you mention, the first example is not properly rendered (the name of the journal starts in bold on the first line and then continues without bold on the second line) while formatting of the second item is OK (title + authors in bold, the rest in non-bold). If you can find a way to have some reasonable rendering in your case that would be perfect.

AKuederle: I will have a look and make sure it looks good in the hml output :)

rtavenar: OK, great! If you build the doc locally, you may have to uncomment the following line (at least I have to on my side :)

https://github.com/rtavenar/tslearn/blob/master/tslearn/docs/conf.py#L23

tslearn/metrics.py
AKuederle: I don't know why I changed that. You are absolutely right

rtavenar:
Does the doc compilation work on your side? On my side, I get the following error:

```
WARNING: autodoc: failed to import function 'subsequence_cost_matrix' from module 'metrics'; the following exception was raised:
No module named 'metrics'
WARNING: autodoc: failed to import function 'subsequence_path' from module 'metrics'; the following exception was raised:
No module named 'metrics'
```

Not sure why I get those, maybe it is just a local config issue, though.

AKuederle:
> Does the doc compilation work on your side? On my side, I get the following error:
> 
> ```
> WARNING: autodoc: failed to import function 'subsequence_cost_matrix' from module 'metrics'; the following exception was raised:
> No module named 'metrics'
> WARNING: autodoc: failed to import function 'subsequence_path' from module 'metrics'; the following exception was raised:
> No module named 'metrics'
> ```
> 
> Not sure why I get those, maybe it is just a local config issue, though.

Mmh I got the same thing. I thought it was an issue with my local config. I couldn't figure out what the issue was

AKuederle:
> > Does the doc compilation work on your side? On my side, I get the following error:
> > ```
> > WARNING: autodoc: failed to import function 'subsequence_cost_matrix' from module 'metrics'; the following exception was raised:
> > No module named 'metrics'
> > WARNING: autodoc: failed to import function 'subsequence_path' from module 'metrics'; the following exception was raised:
> > No module named 'metrics'
> > ```
> > 
> > 
> > Not sure why I get those, maybe it is just a local config issue, though.
> 
> Mmh I got the same thing. I thought it was an issue with my local config. I couldn't figure out what the issue was

I think I figured it out. The issue appears to be the `njit` decorator. With the decorator apidoc does not seem to be able to find the function.

AKuederle:
A possible workaround would be to still have the njit funcs as private api and have a wrapper for as puplic api with documentation

rtavenar:
Thanks a lot @AKuederle ! This is very good job!

Issue #140, opened 2019-08-30
Itakura parallelogram definition with time series of different lengths


rtavenar:
I am experimenting with `tslearn` v0.2 and I have the following issue:

```python

In [1]: import tslearn

In [2]: print(tslearn.__version__)                                                                                                                                                                                                  
0.2.0

In [3]: from tslearn.metrics import dtw                                                                                                                                                                                             

In [4]: import numpy as np                                                                                                                                                                                                          

In [5]: ts1 = np.zeros((100, 1))                                                                                                                                                                                                    

In [6]: ts2 = np.zeros((10, 1))                                                                                                                                                                                                     

In [7]: dtw(ts1, ts2, global_constraint="sakoe_chiba", sakoe_chiba_radius=1)                                                                                                                                                        
Out[7]: inf
```

This is due to the definition of the Sakoe-Chiba band as it is at the moment: if time series have different length, it might be that the term at index `(n, m)` (where n and m are the series lengths) is outside the band, which doesn't make sense.

I think we should fix it asap.

My suggestion for the fix is to consider, for a given row `i`, the element at index `i * sz2 / sz1` (or somethig like that) as the median element of the band.

rtavenar:
@johannfaouzi I can take care of that PR, but I would be interested in your feedback about how to properly deal with it.

rtavenar:
A bit of insight. In fact, I was wrong, the problem is not at the end index:

```python
In [13]: from tslearn.metrics import itakura_mask                                                                                                                                                                                   

In [14]: itakura_mask(10, 3)                                                                                                                                                                                                        
Out[14]: 
array([[ 0., inf, inf],
       [inf, inf, inf],
       [inf, inf, inf],
       [inf,  0., inf],
       [inf,  0., inf],
       [inf,  0., inf],
       [inf,  0., inf],
       [inf, inf, inf],
       [inf, inf, inf],
       [inf, inf,  0.]])
```

I still believe that the notion of median index per row would make sense to avoid such cases.

rtavenar:
OK I was confusing Itakura and SakoeChiba in my previous post.

And the surprising bit is that:

```python
In [1]: from tslearn.metrics import sakoe_chiba_mask                                                                                                                                                                                

In [2]: sakoe_chiba_mask(10, 3, 1)                                                                                                                                                                                                  
Out[2]: 
array([[ 0.,  0., inf],
       [ 0.,  0.,  0.],
       [ 0.,  0.,  0.],
[...]

PR #141, opened 2019-08-30
[MRG] Sakoe-Chiba band and Itakura parallelogram broken in DTW computations

rtavenar:
tslearn version: v0.2.0
Related to Issue: #140 

To be fixed:
1. itakura & sakoe-chiba are swapped in compute_mask (already fixed in initial commit)
2. for itakura: there should always be at least one path, whatever time series shapes, but this is not the case, eg:

```python
In [13]: from tslearn.metrics import itakura_mask                                                                                                                                                                                   

In [14]: itakura_mask(10, 3)                                                                                                                                                                                                        
Out[14]: 
array([[ 0., inf, inf],
       [inf, inf, inf],
       [inf, inf, inf],
       [inf,  0., inf],
       [inf,  0., inf],
       [inf,  0., inf],
       [inf,  0., inf],
       [inf, inf, inf],
       [inf, inf, inf],
       [inf, inf,  0.]])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,19 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-The format is based on 
+The format is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to 
+and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 Changelogs for this project are recorded in this file since v0.2.0.
 
 ## [v0.2.1]
+
+### Added
+
+* A `RuntimeWarning` is raised when an `'itakura'` constraint is set
+that is unfeasible given the provided shapes.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


rtavenar:
In fact, I am not even sure of point 2. : since for Itakura, the parameter is the maximum slope allowed, it could be somehow left to the user to decide if this slope makes sense or not.

But in this case we should at least:
* state it clearly in the docs (docs of dtw, compute_mask and itakura_mask)
* maybe raise a warning when this is not the case, 2 options here:
  1. maybe check at the DTW level if the result is `inf` and the mask is Itakura, then raise a warning?)
  2. or check at the mask level if there is at least one row/column with all infinite values, then raise a warning)

rtavenar:
3rd option: pre-check (if the dimensions make the slope constraint un-doable, then raise a warning)

rtavenar:
When I think about it, I think that option 2 is probably the easiest one to set up: we could for example in the mask function check that lower and upper indices, at each row, are in the correct order, and if not, plan to raise a warning. I will do that when I have time.

rtavenar:
I will merge that PR to master but will not be able to deploy version 0.2.1 (with that bug fix) before next monday.

Issue #146, opened 2019-09-10
[HELP!!!] Any way to set 'sakoe_chiba_radius' in metric_param for tslearn.clustering.TimeSeriesKMeans


chinghoo:
any way to set 'sakoe_chiba_radius' in metric_param for tslearn.clustering.TimeSeriesKMeans when metric is 'dtw'?
the following does not work!!! this makes me crazy for all night today!!!Help!!!

TimeSeriesKMeans(n_clusters=3, metric="dtw", max_iter=5,
...                            max_iter_barycenter=5,
...                            metric_params={"sakoe_chiba_radius": .5},
...                            random_state=0)

GillesVandewiele:
After closely inspecting the code, I may have found what is going on.

cdist_dtw is called within TimeSeriesKMeans. The definition is:
```
def cdist_dtw(dataset1, dataset2=None, global_constraint=None,
              sakoe_chiba_radius=1, itakura_max_slope=2., n_jobs=None):
```

Now when looking at the comments at `sakoe_chiba_radius`, I see: 
```
Radius to be used for Sakoe-Chiba band global constraint. Used only if
        ``global_constraint="sakoe_chiba"``.
```

So `global_constraint: "sakoe_chiba"` must also be passed with metric_params!

@rtavenar this is pretty unclear from the docs I'd say and I had to look into the code to figure this out. It would be more user-friendly that this global_constraint is inferred whenever needed (e.g. when either `sakoe_chiba_radius` or `itakura_max_slope` is passed in the `metric_params` dictionary.

rtavenar:
Hi @chinghoo 

Thanks @GillesVandewiele for your prompt reply. Indeed, it is expected that global_constraint is set to "sakoe_chiba" for sakoe_chiba_radius to be used.

The thing is, as it is implemented now, distances to centroids are computed using this constrained DTW, but DBA algorithm does not have a sakoe_chiba_radius or itakura_max_slope parameter, so this is not satisfactory.

And this is also true for other estimators that use DTW, but for kNN for example, the only problem is the one raised by @chinghoo since only cdist_dtw is used there.

Finally, I am not sure how to fix things, since sakoe_chiba_radius has a default value that cannot be easily differentiated from an user-input value (no None by default). 
Maybe the fix would be to set the default for these parameters as None, but document the fact that this default corresponds to specific values (the current default ones) such that we do not introduce any backward incompatibility.

If you can work on that PR, that would be cool (branch from dev if you do so).

[...]

PR #148, opened 2019-09-18
[WIP] Constrained DTW without setting global_constraint explicitly

rtavenar:
This PR implements a suggestion from #146 that is to allow using constrained DTW by setting only `itakura_max_slope` or `sakoe_chiba_radius` (and not necessarily `global_constraint`).

The choice I made here is as follows:
* if `global_constraint` is set, then it is used even if the corresponding constrain parameter (`itakura_max_slope` or `sakoe_chiba_radius`) is left to default
* if `global_constraint` is not set, then:
  - if none of `itakura_max_slope` or `sakoe_chiba_radius` is set, then no global constraint is applied
  - if one of `itakura_max_slope` or `sakoe_chiba_radius` is set, then the global constraint is considered to be the corresponding one
  - if both `itakura_max_slope` and `sakoe_chiba_radius` are set, a `ValueError` is raised since the constraint cannot be inferred


rtavenar:
Just added DTW constraints to DBA barycenter estimation algorithm.

tslearn/metrics.py
```diff
@@ -671,17 +628,30 @@ def compute_mask(s1, s2, global_constraint=0,
     """
     sz1 = s1.shape[0]
     sz2 = s2.shape[0]
-    if global_constraint == 2:
+    if (global_constraint == 0 and sakoe_chiba_radius is not None
+            and itakura_max_slope is not None):
+        raise ValueError("global_constraint is not set for DTW, but both"
```
GillesVandewiele: This is an option. An alternative would be to raise a warning and make an arbitrary choice (which wouldn't crash the code).

rtavenar: OK, and then I guess the "ideal" default would be "no constraint", no?

tslearn/metrics.py
```diff
@@ -643,26 +594,32 @@ def itakura_mask(sz1, sz2, max_slope=2.):
...
 
     Parameters
     ----------
     s1 : array
         A time series.
+
     s2: array
         Another time series.
+
     global_constraint : {0, 1, 2} (default: 0)
```
GillesVandewiele: What is the reason for having this one in there as well? Is it for backward compatibility? Can't we just infer the type of constraint based on whether the values are set or not and remove this parameter?

tslearn/metrics.py
GillesVandewiele: That is indeed the best option in my opinion as well.

tslearn/metrics.py
```diff
@@ -198,23 +136,30 @@ def dtw_path(s1, s2, global_constraint=None,
...
     global_constraint : {"itakura", "sakoe_chiba"} or None (default: None)
         Global constraint to restrict admissible paths for DTW.
-    sakoe_chiba_radius : int (default: 1)
-        Radius to be used for Sakoe-Chiba band global constraint. Used only if
-        ``global_constraint="sakoe_chiba"``.
-    itakura_max_slope : float (default: 2.)
-        Maximum slope for the Itakura parallelogram constraint. Used only if
-        ``global_constraint="itakura"``.
+
+    sakoe_chiba_radius : int or None (default: None)
```
GillesVandewiele: I would update the docs and explicitly say that itakura_max_slope cannot be set if this one is set, and vice versa.

tslearn/metrics.py
```diff
@@ -117,69 +116,8 @@ def _return_path(acc_cost_mat):
...
-            s2 = s2[numpy.isfinite(s2[:, 0])]
-            mask = compute_mask(s1, s2, global_constraint,
-                                sakoe_chiba_radius, itakura_max_slope)
-            cdist[i, j] = cdist[j, i] = njit_dtw(s1, s2, mask)
-    return cdist
-
-
-def dtw_path(s1, s2, global_constraint=None,
-             sakoe_chiba_radius=1, itakura_max_slope=2.):
+def dtw_path(s1, s2, global_constraint=None, sakoe_chiba_radius=None,
```
GillesVandewiele: Didn't find the line in the diff to comment on, but you can add None as a key to the dictionary as well. They would make the following block redundant in both `dtw_path` and `dtw`:

```
    if global_constraint is not None:
        global_constraint_str = global_constraint
    else:
        global_constraint_str = ""
```

tslearn/tests/test_metrics.py
```diff
@@ -116,6 +116,32 @@ def test_masks():
                                [np.inf, np.inf, np.inf, np.inf, np.inf,  0.]])
     np.testing.assert_allclose(i_mask, reference_mask)
 
+    sz = 10
```
GillesVandewiele: Would there be a way to test that the output of a call with `metric_params` in there actually used those parameters (KMeans or KNN)? E.g. by checking which values are actually calculated in the matrix and comparing that to the calculated mask? The most simple way would be to pass metric_params with both `itakura_max_slope` and `sakoe_chiba_radius` and assert for Warning/Error

tslearn/metrics.py
rtavenar: Backward compatibility implies that we have it for `dtw`, `dtw_path` & co. Then, we could remove this one but we would need to perform the checks in all these individual places, so I suggest we keep it here, unless you have a better idea of how to deal with it.

tslearn/metrics.py
rtavenar: Good idea! Done (locally, will be pushed soon).

tslearn/metrics.py
GillesVandewiele: Hmm don't we only need to do those checks in `compute_mask` and only there?

The checks would then be:

```python3
if sakoe_chiba_radius is not None and itakura_max_slope is not None:
  raise warning
elif sakoe_chiba_radius is not None:
  mask = sakoe_chiba_mask(sz1, sz2, radius=sakoe_chiba_radius)
elif itakura_max_slope is not None:
  mask = itakura_mask(sz1, sz2, max_slope=itakura_max_slope)
else:
  mask = numpy.zeros((sz1, sz2))
```

tslearn/metrics.py
rtavenar: Done (locally, will push soon).

tslearn/metrics.py
rtavenar: Then, the case where `global_constraint` is set to either `sakoe_chiba` or `itakura` and both parameters are filled would end up raising a warning, which is a pity I would say.

GillesVandewiele: except if we remove it entirely or never use it. For backward compatibility, we could raise a warning that `global_constraint` is ignored/deprecated

tslearn/metrics.py
rtavenar: Done (in every place in `metrics.py` that have these parameters).

tslearn/metrics.py
rtavenar: OK, but if we deprecate it, we should warn it in v0.3 and remove it actually in v0.5. I can add the deprecation warning if that seems ok to you.

GillesVandewiele: It is definitely not a necessity, it just seems a redundant parameter as we can infer it based on whether or not one of the variables is set to something else to None.

tslearn/tests/test_metrics.py
rtavenar: Not sure how we could test that. Maybe ensure that we get a similar result as with Euclidean distance (when radius is minimal / slope constraint is the toughest possible)

GillesVandewiele: Yeah that is a nice idea, you would want to have a mask where only the diagonal is filled with 0's for that, right? Not sure how that can be created through `itakura_max_slope` or `sakoe_chiba_radius` (I'm not entirely familiar with those)

rtavenar: Done.

GillesVandewiele:
Great! I think all additions + the new functionality is now well tested. LGTM :)

rtavenar:
Thanks @GillesVandewiele !

PR #149, opened 2019-09-24
Add utilities to cast from/to other python time series packages

rtavenar:
The idea is to have a way to easily cast time series datasets from/to tslearn format to/from formats used by other python time series libraries.

I started with `pyts`. @johannfaouzi could you tell if what I assume to be the type of `pyts` datasets is OK or not?

The full list I would like to have would be:
* [x] pyts
* [x] sktime
* [x] cesium-ml
* [x] seglearn
* [x] tsfresh
* [x] stumpy
* [x] pyflux

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ install:
 - conda update -q conda
 - conda info -a
 - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION Cython numpy>=1.14 scipy tensorflow keras
-  scikit-learn numba joblib>=0.12 pandas>=0.24
+  scikit-learn numba joblib>=0.12 pandas>=0.24 cesium
 - source activate test-environment
 - pip install pytest
 - if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then pip install coverage==4.3 "pytest-cov<2.6"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,9 @@ install:
 - conda update -q conda
 - conda info -a
 - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION Cython numpy>=1.14 scipy tensorflow keras
-  scikit-learn numba joblib>=0.12 pandas>=0.24 cesium
+  scikit-learn numba joblib>=0.12 pandas>=0.24
 - source activate test-environment
-- pip install pytest
+- pip install cesium pytest
 - if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then pip install coverage==4.3 "pytest-cov<2.6"; fi
 - pip install argparse
 - pip install codeclimate-test-reporter codecov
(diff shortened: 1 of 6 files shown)


johannfaouzi:
pyts can deal with multivariate time series, although most of the provided features are for univariate time series for the moment.

pyts convention:
* univariate: (n_ts, sz)
* multivariate: (n_ts, d, sz)

The first axis is always the sample axis and the last axis is always the time axis.

rtavenar:
OK, thanks, I will change that.
Romain Tavenard
Maître de conférences / Assistant professor
Univ. Rennes - LETG
Tél. / Phone : +33 2 99 14 18 04
http://rtavenar.github.io/research/ <http://rtavenar.github.io/research/>

johannfaouzi:
A PR from Hicham Janati was merged into master last week that extends the use of DTW and DTW-related stuff (constraint regions, DTW variants) to variable length time series. So most of the pyts.metrics module can deal with variable length time series, but not the other modules.

I will try to check it out tomorrow or the day after!

rtavenar:
OK, I do not really need to know which algos support them. But are they supposed to be encoded just as in tslearn (NaN padded at the end of shorter time series, or is it done another way?

In other words, if I have a dataset that would contain 2 time series: `[1, 2, 3]` and `[1, 2]`: would it be formatted like:
```
[[1, 2, 3    ],
 [1, 2, nan]]
```

or another way (list of arrays, etc)?

rtavenar:
For `sktime`, info is available there: https://alan-turing-institute.github.io/sktime/examples/loading_data.html

rtavenar:
Seems that for `sktime` and `tsfresh`, `pandas` dataframes are used. At the moment, pandas is not a dependency for `tslearn`, and I would prefer not to add hard dependencies if possible, but having a soft-dependency that is required only if casting to those formats would probably make sense.

johannfaouzi:
> OK, I do not really need to know which algos support them. But are they supposed to be encoded just as in tslearn (NaN padded at the end of shorter time series, or is it done another way?
> 
> In other words, if I have a dataset that would contain 2 time series: `[1, 2, 3]` and `[1, 2]`: would it be formatted like:
> 
> ```
> [[1, 2, 3  ],
>  [1, 2, nan]]
> ```
> 
> or another way (list of arrays, etc)?

For the moment it is not supported at all. I would like to have two different values for missing and padding values (probably `np.nan ` for missing values and maybe `np.inf` for padding values). Except for DTW-related stuff, nothing in `pyts` can currently deal with variable length time series (and for practical reasons I think that it will stay like this for most algorithms that are not related to DTW).

Conclusion: not yet supported, should be supported soon for KNN with DTW, will keep you updated

rtavenar:
@johannfaouzi : thanks for the detailed answer. 

So, is that OK for you if, for the moment, we have a check in `tslearn` that the dataset is made of equal-length time series and later (when it is supported in `pyts`), we change the conversion functions to mirror the change?

PR #152, opened 2019-09-29
[MRG] Adding SAX+MINDIST to KNN

GillesVandewiele:
This PR contains the following changes:

* 'sax' is now a valid metric for KNN:
```python3
knn = KNeighborsTimeSeriesClassifier(n_neighbors=1, metric='sax')
```

* Added BaseEstimator to classes in `preprocessing` module so that they can be used within a Pipeline (errors were raised when using `TimeSeriesScalerMeanVariance`)

* Fixed a bug in `kneighbors` method which would always return [0] as nearest neighbor for every sample.
```python3
knn = KNeighborsTimeSeriesClassifier(n_neighbors=1, metric='dtw')
knn.fit(X_train, y_train)
_, ind = knn.kneighbors(X_test)
# ind would be filled with 0's
```

* Slightly changed to code of `kneighbors` so that its result is consistent with sklearn. There was a small difference in breaking ties (tslearn would pick largest index while sklearn would pick the smallest index). Now the following code is equivalent:
```
knn = KNeighborsTimeSeriesClassifier(n_neighbors=1, metric='dtw')
knn.fit(X_train, y_train)
_, ind = knn.kneighbors(X_test)

knn = KNeighborsTimeSeriesClassifier(n_neighbors=1, metric='precomputed')
all_X = numpy.vstack((X_train, X_test))
distances = pairwise_distances(all_X, metric=dtw)
X_train = distances[:len(X_train), :len(X_train)]
X_test = distances[len(X_train):, :len(X_train)]
knn.fit(X_train, y_train)
_, ind = knn.kneighbors(X_test)

# both ind vectors are now equal (while that was not the case before this PR)
```

Some remarks:

* I am unexperienced with numba; adding an `njit` decorator to cdist_sax did not work immediately, I could perhaps use some help with that.


GillesVandewiele:
I merged with the latest dev branch before working futher, hence why @AKuederle his commits are in here as well. #GitNoob

GillesVandewiele:
Could also maybe important to note that the technique doesn't really work that well, the results are quite consistent with the paper though, but I thought SAX would work insanely fast, which turns out not to be the case currently:

```
|      dataset       | sax error  |  sax time  | eucl error | eucl time  |
--------------------------------------------------------------------------
|    SyntheticControl|     0.02667|        0.12|        0.12|     0.06491|
|            GunPoint|     0.20667|     0.08667|     0.08667|     0.00989|
|             OSULeaf|     0.47521|     0.47934|     0.47934|     0.03286|
|               Trace|        0.52|        0.24|        0.24|     0.01209|
|            FaceFour|     0.14773|     0.21591|     0.21591|     0.01297|
|          Lightning2|     0.21311|      0.2459|      0.2459|     0.00842|
|          Lightning7|     0.46575|     0.42466|     0.42466|     0.00799|
|              ECG200|        0.12|        0.12|        0.12|     0.00945|
|                Fish|     0.50286|     0.21714|     0.21714|     0.02922|
|               Plane|     0.04762|      0.0381|      0.0381|     0.01164|
|                 Car|        0.35|     0.26667|     0.26667|     0.00749|
|                Beef|     0.53333|     0.33333|     0.33333|     0.00439|
|              Coffee|     0.46429|         0.0|         0.0|     0.00389|
|            OliveOil|     0.83333|     0.13333|     0.13333|      0.0043|
--------------------------------------------------------------------------
```

rtavenar:
Great @GillesVandewiele 

I will try to review this PR asap, probably early next week. I am still unsure if we should target to get a fast SAX+MINDIST in this PR or delay that to a later PR.

GillesVandewiele:
Hi @rtavenar there's no rush in reviewing I'd say. I still need to fix Travis etc, so maybe after that passes reviewing is more appropriate.

rtavenar:
That looks great @GillesVandewiele !

Maybe a minimal test would be to check that the distances returned by kNN's kneighbors are indeed lower bounding L2 (I think they should be, if I remember the original paper correctly, but maybe it would be worth checking :)

Hope this helps

GillesVandewiele:
Hi Romain, I added the test. I'm not sure how this codecov report is actually generated, it's weird that it is reporting about barycenters.py in this PR... 

There's also one other point that's maybe worth discussing. Currently, I always perform a SAX transformation on the provided input data, but this then actually assumes that the user has not already done that. Moreover, the current SAX transformation will fail pretty badly if the data is not standard-normalized. So maybe it could be better to actually change the `metric` to `mindist` and require the user to standard-normalize + sax-transform the data?

rtavenar:
@GillesVandewiele 

This is a very good question. I do not think we can use Pipelines here, because we need parameters of the previous step for the distance computation, so I guess the way you did it is reasonable, or at least I do not see a more straight-forward approach. 

Maybe having a Pipeline (standardize + SAX kNN) in your gallery example (together with some comment saying that this is good practice for SAX to standardize your data beforehand) would be a good idea. And a note in the kNN doc (under the metric param) also.

Let me know when I should start my review, I could probably do it next week if you feel your code is in sufficiently advanced state.

Have a nice week-end,
Romain

GillesVandewiele:
Two other options would be the following:

1) always standard-normalize the input that we get. As far as I know, standard-normalizing is quite idempotent (so normalizing data that is already normalized will not really have great effects). Moreover, we do the normalization per timeseries individually which makes it a lot easier to do this as well.

2) Perform some hypothesis testing on (a sample of) the input data and raise a warning or perform normalization when results of hypothesis test are that the data isn't normally distributed.


What do you want me to do about the codecov btw?

rtavenar:
Hi,

I would be in favor of option 1. 
However, 
1. this should be documented
2. normalization should be performed dataset-wise, not per time series, in this case, I think

Concerning codecov, I can have a look when I'll do a review.

GillesVandewiele:
Forgot to mention it yesterday, but I think the code is now ready to be reviewed, whenever you have time @rtavenar. I refactored quite a lot of code in `neighbors.py` so we definitely need to be careful there.

rtavenar:
I did not find time to review it this week, sorry. I hope I can do it next week.

tslearn/docs/examples/plot_sax_mindist_knn.py
```diff
@@ -0,0 +1,103 @@
...
+        s += '{:>12}|'.format(time_sax)
+        s += '{:>12}|'.format(eucl_error)
+        s += '{:>12}|'.format(time_euclidean)
+        print(s.strip())
+
+    print('-'*(len(columns) * 13 + 22))
+
+
+numpy.random.seed(0)
+# Generate a random walk time series
```
rtavenar: This comment is not appropriate

tslearn/docs/examples/plot_sax_mindist_knn.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
+"""
+1-NN with SAX + MINDIST
+=======================
+
+This example presents a comparison performs kNN with k=1, using MINDIST from
```
rtavenar: Suggestion for improvement:

This example presents a comparison between k-Nearest Neighbor runs with k=1.
It compares the use of:
- MINDIST (see [1]) on SAX-transformed data;
- Euclidean distance on raw time series.

The comparison is based on test accuracy computed for several benchmark datasets.

[1] Cite the paper here

tslearn/docs/examples/plot_sax_mindist_knn.py
```diff
@@ -0,0 +1,103 @@
...
+    ('Coffee', 128),
+    ('OliveOil', 256)
+]
+
+accuracies = {}
+times = {}
+for dataset, w in datasets:
+    X_train, y_train, X_test, y_test = data_loader.load_dataset(dataset)
+    metric_params = {'n_segments': w, 'alphabet_size_avg': 10}
+    knn = KNeighborsTimeSeriesClassifier(n_neighbors=1, metric='sax',
```
rtavenar: You could probably define the classifiers outside the loop.

tslearn/docs/examples/plot_sax_mindist_knn.py
```diff
@@ -0,0 +1,103 @@
...
+                                         metric_params=metric_params)
+    start = time.time()
+    knn.fit(X_train, y_train)
+    acc_sax = accuracy_score(y_test, knn.predict(X_test))
+    time_sax = time.time() - start
+
+    knn = KNeighborsTimeSeriesClassifier(n_neighbors=1, metric='euclidean')
```
rtavenar: Same comment here.

tslearn/docs/examples/plot_sdtw.py
```diff
@@ -62,7 +62,7 @@
 # fit this example)
 potential_matches = find_peaks(-cost_func, distance=sz * 0.75, height=-50)[0]
 
-# Calculate the optimal warping path starting from each of the identified minima
+# Calculate the optimal warping path starting from each of the minima
```
rtavenar: I prefer the original formulation since `find_peaks` could have many different forms, depending on application.

tslearn/metrics.py
```diff
@@ -12,14 +12,77 @@
...
+from tslearn.cysax import cydist_sax
 
 from tslearn.utils import to_time_series, to_time_series_dataset, ts_size, \
     check_equal_size
 
 __author__ = 'Romain Tavenard romain.tavenard[at]univ-rennes2.fr'
 
 GLOBAL_CONSTRAINT_CODE = {None: 0, "": 0, "itakura": 1, "sakoe_chiba": 2}
-VARIABLE_LENGTH_METRICS = ["dtw", "gak", "softdtw"]
+VARIABLE_LENGTH_METRICS = ["dtw", "gak", "softdtw", "sax"]
```
rtavenar: Is SAX really supposed to be a variable-length metric? Isn't is supposed to run on equal size time series?

tslearn/metrics.py
```diff
@@ -12,14 +12,77 @@
...
+
+
+def cdist_sax(dataset1, breakpoints_avg, size_fitted, dataset2=None,
```
rtavenar: The reason why you cannot @njit this function is probably because it relies on Cython code. This code could probably be ported to numba at some point, but this is not the point of this PR, I would say.

tslearn/neighbors.py
```diff
@@ -12,16 +12,75 @@
...
+                n_segments=n_segments,
+                alphabet_size_avg=alphabet_size_avg
+            )
+
+        X = to_time_series_dataset(X)
+        X = self._sax.fit_transform(X)
+
+        return X
+
+    def _calculate_X_(self, X, other_X=None):
```
rtavenar: could maybe use a more explicit name, such as `_precompute_cross_distances`?

tslearn/neighbors.py
```diff
@@ -64,37 +123,33 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
+            elif hasattr(self, '_ts_fit') and self._ts_fit is not None:
+                fit_X = self._ts_fit
             else:
                 fit_X = self._X_fit
-            if parallelize:
-                full_dist_matrix = cdist_fun(X, fit_X, n_jobs=self.n_jobs,
-                                             verbose=self.verbose, 
-                                             **metric_params)
+
+            if (self.metric in VARIABLE_LENGTH_METRICS or
```
rtavenar: The test should not be this one, it should be something like "is it a tslearn metric?" maybe we could define such a list in the header of `neighbors.py`, something like:

```python
TSLEARN_VALID_METRICS = ['dtw', 'softdtw', 'sax']
neighbors.VALID_METRICS['brute'].extend(TSLEARN_VALID_METRICS)
```

tslearn/neighbors.py
```diff
@@ -104,6 +159,8 @@ def cdist_fun(X, Xp):
 
         n_ts = X.shape[0]
         sample_range = numpy.arange(n_ts)[:, None]
+        ind = ind[
```
rtavenar: What does this do?

tslearn/neighbors.py
```diff
@@ -119,18 +176,19 @@ class KNeighborsTimeSeries(KNeighborsTimeSeriesMixin, NearestNeighbors):
...
     n_neighbors : int (default: 5)
         Number of nearest neighbors to be considered for the decision.
-    metric : {'dtw', 'softdtw', 'euclidean', 'sqeuclidean', 'cityblock'}
-    (default: 'dtw')
+    metric : {'dtw', 'softdtw', 'euclidean', 'sqeuclidean', 'cityblock',
+              'sax'} (default: 'dtw')
         Metric to be used at the core of the nearest neighbor procedure.
-        DTW is described in more details in :mod:`tslearn.metrics`.
+        DTW and SAX are described in more detail in :mod:`tslearn.metrics`.
+        When SAX is provided as a metric, each timeseries standard-normalized
```
rtavenar: each time series is...

tslearn/preprocessing.py
```diff
@@ -112,8 +114,8 @@ class TimeSeriesScalerMinMax(TransformerMixin):
     """
     def __init__(self, value_range=(0., 1.), min=None, max=None):
         self.value_range = value_range
-        self.min_ = min
```
rtavenar: Why changing that?

Seems like a usual way to name attributes in sklearn that we try to mimic in tslearn, eg:
https://github.com/scikit-learn/scikit-learn/blob/1495f6924/sklearn/preprocessing/data.py#L233

tslearn/preprocessing.py
```diff
@@ -145,34 +149,40 @@ def transform(self, X, y=None, **kwargs):
...
-        if self.value_range[0] >= self.value_range[1]:
+        if value_range[0] >= value_range[1]:
             raise ValueError("Minimum of desired range must be smaller"
-                             " than maximum. Got %s." % str(self.value_range))
+                             " than maximum. Got %s." % str(value_range))
 
+        X = check_array(X, allow_nd=True, force_all_finite=False)
+        X = check_dims(X, self._X_fit, extend=False)
         X_ = to_time_series_dataset(X)
         min_t = numpy.min(X_, axis=1)[:, numpy.newaxis, :]
```
rtavenar: If we allow nans (cf tags below), shouldn't we use https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanmin.html ?

tslearn/preprocessing.py
```diff
@@ -145,34 +149,40 @@ def transform(self, X, y=None, **kwargs):
...
         max_t = numpy.max(X_, axis=1)[:, numpy.newaxis, :]
```
rtavenar: Same thing here with nanmax

tslearn/preprocessing.py
```diff
@@ -197,10 +207,8 @@ class TimeSeriesScalerMeanVariance(TransformerMixin):
             [ 1.22474487]]])
     """
     def __init__(self, mu=0., std=1.):
-        self.mu_ = mu
-        self.std_ = std
-        self.global_mean = None
-        self.global_std = None
+        self.mu = mu
```
rtavenar: As above: why use mu instead of mu_? (same for std, global_mean and global_std)

rtavenar:
Hi @GillesVandewiele 

Sorry for the delay. I have made my review, not much to say (except individual comments), this seems to be in pretty advanced state, thanks for this!

GillesVandewiele:
Thanks for this Romain! 

Small heads up: I'll be abroad for three weeks, including two weeks without laptop starting from today though. I'll do my best to integrate as much comments as possible!

rtavenar:
No problem, of course!

tslearn/preprocessing.py
GillesVandewiele: I don't know by heart anymore, but I think it was some kind of fix to make sure the `check_estimator` tests passed (same for the mean-variance scaler). 

I'll re-check on this and see if the tests pass without this change as well!

tslearn/neighbors.py
GillesVandewiele: That is indeed a much better and clear name!

tslearn/metrics.py
GillesVandewiele: Hmm that is a good call of which I am not entirely sure... But since SAX converts a longer timeseries to a fixed number of segments, it could maybe work on variable length timeseries as well? I would have to do a bit more research on this to see whether that makes sense...

tslearn/docs/examples/plot_sdtw.py
GillesVandewiele: This is actually a change that I just did to make sure the PEP8 checks passed. I will revert it and split the comments over multiple lines or smth.

tslearn/neighbors.py
GillesVandewiele: It will do the same as the previous code, but this is similar to how sklearn does it. This will result in differences when breaking ties (two or more samples having minimal distances), and therefore ensures that the output using an sklearn pipeline (without any tslearn classifiers) is the same as when we are using tslearn classifiers.

Tbh, I think this code is much less clear then the previous one, so it could definitely be reverted back.

tslearn/preprocessing.py
GillesVandewiele: Good call (and indeed same for nanmax, nanmean, nanstd, ...).

We should probably have tests for this in place as well.

tslearn/neighbors.py
rtavenar: OK, then could you add that info in a comment?

GillesVandewiele:
Oh, and before I forget.. Do you happen to know why the codecov is complaining (about code unrelated to this PR)?

rtavenar:
No, but I do not really care (I could take care of that just before merging to master, if necessary)

tslearn/neighbors.py
GillesVandewiele: Agreed, I will document this better :) (bcs even I have troubles understanding it clearly now already)

tslearn/docs/examples/plot_sax_mindist_knn.py
GillesVandewiele: I have made this change, but I had to do the following in order to get the same results as the previous version:
```python3
knn_sax = clone(knn_sax).set_params(metric_params=metric_params)
```

So not sure whether this is actually better than just re-instantiating the model in every loop.

GillesVandewiele:
Did the tensorflow version for tslearn recently got upgraded to >= 2? Seems like Travis is failing due to the fact that `set_random_seed` is no longer a valid import...

rtavenar:
> Did the tensorflow version for tslearn recently got upgraded to >= 2? Seems like Travis is failing due to the fact that `set_random_seed` is no longer a valid import...

I changed requirements.txt on master so that tensorflow version is < 2, but did not change it on dev: could you do it in your PR?

rtavenar:
I just refactored code for cdist calls, FYI, I will have to resolve conflicts before merge.

rtavenar:
Hi @GillesVandewiele ,

When you have time to take my last comments into account, we could proceed to merge. 
No hurry (I am myself very busy these days so I can understand if you are too :)

GillesVandewiele:
Hey @rtavenar,

Just arrived back in Belgium after three weeks of travelling around in New-Zealand in which I tried to refrain from working as much as possible :laughing:. You can expect an update on this PR in the nearby future!

tslearn/metrics.py
GillesVandewiele: So I think SAX can handle variable lengths:
* The timeseries are converted to a fixed number of segments
* In the original paper (https://www.cs.ucr.edu/~eamonn/SAX.pdf) they state: "SAX allows a time series of arbitrary length n to be reduced to a string  of  arbitrary  length w"
* Dr. Lin uses SAX as pre-processing step in another work (https://arxiv.org/pdf/1802.04883.pdf) titled "Efficient Discovery of Variable-length Time SeriesMotifs with Large Length Range in Million ScaleTime Series"


Perhaps we could put a test in place for these variable-length metrics?

tslearn/neighbors.py
GillesVandewiele: VARIABLE_LENGTH_METRICS is already defined in the header of tslearn/metrics.py, so I think this does check whether it is a valid tslearn metric?

tslearn/neighbors.py
GillesVandewiele: Ok so I've been more closely inspecting the code, and it does not only mimic the output of sklearn this way (TimeSeriesKNeighbor~(metric='euclidean') has the same results as feeding a distance matrix to sklearn.KNeighbors~(metric='euclidean')), it is also an optimization as `argpartition` happens in O(n) and then the sort happens on the k nearest neighbors in O(k log(k)), and k << n.

tslearn/metrics.py
GillesVandewiele: Had to make some changes in PAA to make this work. It is no longer vectorized anymore though:
https://github.com/rtavenar/tslearn/pull/152/commits/b96ebe95e2901cdb27b8c7ea5bead2c0b9265e59#diff-a071054ac597f00ec51b4bf94e99d23fR143

Perhaps this can be written more efficiently to avoid all these for-loops?

tslearn/preprocessing.py
GillesVandewiele: I think the reason for this is the following: all the objects/parameters passed by the user to the constructor have to be stored under the same name in the self object. As the name of the parameters here are min and max, these have to be the same in the self-object to be sklearn-compatible. An alternative is to refactor the parameter names to min_ and max_ but that is kind of weird, and they are going to be deprecated soon anyway (and replaced for value_range)

I will undo the change one more time just to double check this, but I think the check_estimator will fail.

GillesVandewiele: I double checked it, and it will raise this error:

```
AssertionError: Estimator TimeSeriesScalerMinMax should not set any attribute apart from parameters during init. Found attributes ['max_', 'min_'].
```

GillesVandewiele:
I think I've handled most comments now (left responses on some), I also refactored the PAA a bit so that it can now handle variable-length time series, which was required so that SAX can handle it as well. This was discovered by extending the variable_length test. The code for that could perhaps be more efficient/vectorized though:
```python3
    def _transform(self, X, y=None):
        n_ts, sz, d = X.shape
        X_transformed = numpy.empty((n_ts, self.n_segments, d))
        for i_ts in range(n_ts):
            for i_dim in range(d):
                sz = numpy.sum(~numpy.isnan(X[i_ts, :, i_dim]))
                sz_segment = sz // self.n_segments
                for i_seg in range(self.n_segments):
                    start = i_seg * sz_segment
                    end = start + sz_segment
                    segment = X[i_ts, start:end, i_dim]
                    X_transformed[i_ts, i_seg, i_dim] = numpy.nanmean(segment)
```

Seems like Travis succeeds except that it takes quite some time for python2 (had to increase travis_wait), that all pep8 issues are solved as well, and that the coverage has increased :)

rtavenar:
Thanks! I should be available for a review by the end of the week.

tslearn/neighbors.py
rtavenar: Yes, but I meant: at this point, we want to add `tslearn` metrics to kNN class, even if these functions do not operate on variable-length data. So, in terms of semantics, what we want to do is add tslearn-specific metrics, not necessarily variable-length metrics. 

Not sure if I'm clear here.

GillesVandewiele: Ok makes sense. So do you want to keep track of both where the variable-length one is a subset of all tslearn metrics?

rtavenar:
> I think I've handled most comments now (left responses on some), I also refactored the PAA a bit so that it can now handle variable-length time series, which was required so that SAX can handle it as well. This was discovered by extending the variable_length test. The code for that could perhaps be more efficient/vectorized though:
> 
> ```python
>     def _transform(self, X, y=None):
>         n_ts, sz, d = X.shape
>         X_transformed = numpy.empty((n_ts, self.n_segments, d))
>         for i_ts in range(n_ts):
>             for i_dim in range(d):
>                 sz = numpy.sum(~numpy.isnan(X[i_ts, :, i_dim]))
>                 sz_segment = sz // self.n_segments
>                 for i_seg in range(self.n_segments):
>                     start = i_seg * sz_segment
>                     end = start + sz_segment
>                     segment = X[i_ts, start:end, i_dim]
>                     X_transformed[i_ts, i_seg, i_dim] = numpy.nanmean(segment)
> ```
> 
> Seems like Travis succeeds except that it takes quite some time for python2 (had to increase travis_wait), that all pep8 issues are solved as well, and that the coverage has increased :)

Hum, I agree that we can no longer vectorize on the first dimension (since different time series can have different lengths), but the assumption is that, for a time series, all modalities have the same length, so we could probably vectorize on that and get a mix between original version and your version, something like:

```python
    def _transform(self, X, y=None):
        n_ts, sz, d = X.shape
        X_transformed = numpy.empty((n_ts, self.n_segments, d))
        for i_ts in range(n_ts):
            sz_segment = ts_size(X[i_ts]) // self.n_segments
            for i_seg in range(self.n_segments):
                start = i_seg * sz_segment
                end = start + sz_segment
                X_transformed[i_ts, i_seg, :] = X[i_ts, start:end, :].mean(axis=1)
```

tslearn/neighbors.py
rtavenar: Yes, something like that would be great.

rtavenar:
Hi @GillesVandewiele 

Do you think you could find time to finalize this PR soon?

Romain

GillesVandewiele:
Hi @rtavenar 

I am so sorry that this have been dragging on for so long... I will try to finalize it this weekend! It should not be too much of work anymore.

Kind regards,
Gilles

rtavenar:
> Hi @rtavenar
> 
> I am so sorry that this have been dragging on for so long... I will try to finalize it this weekend! It should not be too much of work anymore.
> 
> Kind regards,
> Gilles

No pb! That's perfect if you can finalize it soon.

tslearn/neighbors.py
```diff
@@ -12,16 +12,75 @@
...
 
-neighbors.VALID_METRICS['brute'].extend(['dtw', 'softdtw'])
+neighbors.VALID_METRICS['brute'].extend(['dtw', 'softdtw', 'sax'])
 
 
 class KNeighborsTimeSeriesMixin(KNeighborsMixin):
     """Mixin for k-neighbors searches on Time Series."""
 
+    def _sax_preprocess(self, X, n_segments=10, alphabet_size_avg=4):
+        # First, we standard-normalize the data using mu and sigma
```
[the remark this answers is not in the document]
GillesVandewiele: I don't know. I think this again depends on the type of application... 
Screenshot from 2020-03-08 10-18-11

Like here, would you prefer that ts2 is more similar to ts3 than to ts1, or do you prefer that all pairs of timeseries have the same similarity?

tslearn/metrics.py
```diff
@@ -669,8 +671,89 @@ def compute_mask(s1, s2, global_constraint=0,
...
+
+    dataset2 : array-like (default: None)
+        Another dataset of time series. If `None`, self-similarity of
+        `dataset1` is returned.
+
+    n_jobs : int or None, optional (default=None)
+        The number of jobs to run in parallel.
+        ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
+        ``-1`` means using all processors. See scikit-learns'
+        `Glossary <https://scikit-learn.org/stable/glossary.html#term-n-jobs>`_
```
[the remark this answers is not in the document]
GillesVandewiele: Did I just have to append two underscores to every link? Or did you want me to use a reference and then only have the URI in the docs once?

tslearn/neighbors.py
GillesVandewiele: Did you mean something like this? (see latest commit)

GillesVandewiele:
I handled most of the feedback and left some comments where needed.

tslearn/neighbors.py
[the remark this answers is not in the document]
johannfaouzi: I would expect all the pairs to have the same similarity, but it's just my point of view. I'm not an expert on this but I don't think that I read a single paper in which they compute the mean and the standard deviation on the whole dataset.

In the SAX paper, they say that they normalize the time series so that each time series has zero mean and unit variance.
Capture d’écran 2020-03-08 à 13 58 18

tslearn/metrics.py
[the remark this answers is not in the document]
johannfaouzi: Two underscores for external links (links to `scikit-learn` or `joblib` documentation for instance).

tslearn/neighbors.py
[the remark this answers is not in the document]
GillesVandewiele: Ok, I missed that part! I'll change it to have time-series-wise normalization later!

johannfaouzi:
One test if failing for KNN with SAX. My guess would be that
```python
X = to_time_series_dataset([[1, 2, 3, 4],
                            [1, 2, 3],
                            [2, 5, 6, 7, 8, 9],
                            [3, 5, 6, 7, 8]])
```
is transformed into SAX with two segements
```python
clf = KNeighborsTimeSeriesClassifier(metric="sax", n_neighbors=1,
                                     metric_params={'n_segments': 2})
```
and all the new time series are equal to `[0, 1]`, and when you perform classification with 1NN you may get the class of the first sample (`np.argmin` returns the first index).

Could you try to print the SAX transformation for this dataset? You could also try to make the time series for class 1 decreasing instead of increasing, it should fix this.

GillesVandewiele:
> One test if failing for KNN with SAX. My guess would be that
> 
> ```python
> X = to_time_series_dataset([[1, 2, 3, 4],
>                             [1, 2, 3],
>                             [2, 5, 6, 7, 8, 9],
>                             [3, 5, 6, 7, 8]])
> ```
> 
> is transformed into SAX with two segements
> 
> ```python
> clf = KNeighborsTimeSeriesClassifier(metric="sax", n_neighbors=1,
>                                      metric_params={'n_segments': 2})
> ```
> 
> and all the new time series are equal to `[0, 1]`, and when you perform classification with 1NN you may get the class of the first sample (`np.argmin` returns the first index).
> 
> Could you try to print the SAX transformation for this dataset? You could also try to make the time series for class 1 decreasing instead of increasing, it should fix this.

Thanks Johann! Let me see if I can fix it :)

GillesVandewiele:
> LGTM, thanks Gilles! @rtavenar

No thank you for again some great feedback! Always a great learning experience to do a PR here! Now next up should be some improvements for the ShapeletModel. I hope I can do it faster than this one ;)

tslearn/neighbors.py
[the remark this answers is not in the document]
GillesVandewiele: So I removed all normalization. It is now up to the user to apply timeseries-normalization before running KNN!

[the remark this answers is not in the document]
rtavenar: OK for me. Could you at least document it in the class definition (something like "SAX is supposed to operate on centered-reduced time series, the use of `TimeSeriesScalerMeanVariance` is strongly encouraged")?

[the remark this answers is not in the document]
GillesVandewiele: Ok, I indeed forgot to re-check if docs are still up-to-date. Let me make that change & check up on all docs in a bit!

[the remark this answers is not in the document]
rtavenar: Or maybe in the class definition + the distance function?

rtavenar:
This looks good to me too. I had a simple comment regarding the documentation and my last comment is that you can document your changes in the changelog before merge.

Great job, once again, thanks @GillesVandewiele !

GillesVandewiele:
Hi Romain,

Thanks! Where/how do you want me to add it to the changelog? Do I add the following just on top?

```
## [v0.3.0]

### Changed
* TimeSeriesScalerMeanVariance and TimeSeriesScalerMinMax are now completely sklearn-compliant
* Bugfix in kneighbors() methods.

### Added
* Nearest Neighbors on SAX representation (with custom distance)
* Calculate pairwise similarity matrix between SAX representations
* PiecewiseAggregateApproximation can now handle variable lengths
```

rtavenar:
Yes you can create a `## [Towards v0.4.0]` section and describe your changes there.

GillesVandewiele:
Alright done! :)

tslearn/metrics.py
```diff
@@ -1418,7 +1418,9 @@ def gamma_soft_dtw(dataset, n_samples=100, random_state=None):
 def cdist_sax(dataset1, breakpoints_avg, size_fitted, dataset2=None,
               n_jobs=None, verbose=0):
     r"""Calculates a matrix of distances (MINDIST) on SAX-transformed data,
-    as presented in [1]_.
+    as presented in [1]_. It is important to note that this function
+    expects dataset1 and dataset2 to be normalized to have zero mean
```
rtavenar: If stated this way, it is not clear which normalization is expected (per time series or dataset wise )

GillesVandewiele: Agreed! Pushing fix now.

Issue #45, opened 2018-05-31
Consider refactoring DBA to Majorize-Minimize Mean Algorithm


unnamedplay-r:
Referencing issue #44, DBA can be expressed in matrix and vectorized terms that might be more suited for numpy's vectorized operations. I think it would be worthwhile to explore this to see if we can get a considerable performance boost by expressing the function in this form.

Instead of removing the internals of the original algorithm, it might be worth considering adding a parameter to the function which takes values such as `'mm'` or `'original'` to differentiate what kind of internal operation it uses: either the form in the **original** paper, or the **MM** expression of the algorithm.

Relevant section in the Schultz & Jain (2017) paper is _4.2. The Majorize-Minimize Mean Algorithm_ and it's respective algorithm: 2.

rtavenar:
It is now the default for DBA on `dev` branch, and will be included in tslearn from version 0.3.
Thanks @unnamedplay-r for the ref, computations are now twice faster in my tests!

unnamedplay-r:
Yay! This is so wonderful @rtavenar, I’m grateful you took the recommendation into consideration.

Issue #44, opened 2018-05-31
Implement (Stochastic) Subgradient Mean


unnamedplay-r:
**Summary:**
Stochastically explore the DTW Averaging Space. It's a useful algorithm for exploring online settings and larger sample sizes. The paper is a bit of work to get through as they rigerously detail a connection between the DTW sample mean to nonsmooth optimization methods for their proposed algorithm. They 
also detail a non-stochastic subgradient method and present a vectorized version of DBA that might provide additional performance benefits (I'll create a seperate issue for refactoring DBA to that).

I think this package would benefit greatly for both the Subgradient (SG) and Stochastic Subgradient (SSG) methods.

**Original Paper:**
_Nonsmooth Analysis and Subgradient Methods for Averaging in Dynamic Time Warping Spaces_
- arxiv pdf: https://arxiv.org/pdf/1701.06393.pdf
- Final version of paper (paywall): https://www.sciencedirect.com/science/article/pii/S0031320317303163

Relevant sections are _4.1. The Subgradient Mean Algorithm_ and _4.3. The Stochastic Subgradient Mean Algorithm_ and their respective algorithms 1 and 3.

rtavenar:
It is now merged to `dev`, and will be included in tslearn from version 0.3.
Thanks @unnamedplay-r for the ref!

PR #155, opened 2019-10-02
[WIP] Implement MM algorithm for DBA

rtavenar:
As suggested (loooong ago) in #45 , there exist a vectorized form of the DBA algorithm. I have implemented it, and I like it because:

* it it strictly equivalent to DBA
* the code is more readable imho
* it is a bit faster:

```python
import numpy as np
import time

from tslearn.barycenters import dtw_barycenter_averaging_mm, dtw_barycenter_averaging


__author__ = 'Romain Tavenard romain.tavenard[at]univ-rennes2.fr'


n, sz, d = 500, 200, 10
data = np.random.randn(n, sz, d)

t0 = time.time()
bar = dtw_barycenter_averaging_mm(X=data)
print("Elapsed time for MM version: %f" % (time.time() - t0))

t0 = time.time()
bar = dtw_barycenter_averaging(X=data)
print("Elapsed time for DBA version: %f" % (time.time() - t0))
```

gives:

```
Elapsed time for MM version: 25.453088
Elapsed time for DBA version: 37.681433
```

on my machine.

In the end, I am not so sure about keeping original DBA code since it seems more likely to be/get buggy than this one.

### DIFF
--- a/tslearn/barycenters.py
+++ b/tslearn/barycenters.py
@@ -5,11 +5,13 @@ barycenter computation.
 
 # Code for soft DTW is by Mathieu Blondel under Simplified BSD license
 
+from __future__ import division
 import numpy
 from scipy.interpolate import interp1d
 from scipy.optimize import minimize
 from sklearn.exceptions import ConvergenceWarning
 import warnings
+from sklearn.utils import check_random_state
 
 from tslearn.utils import to_time_series_dataset, check_equal_size, \
     to_time_series, ts_size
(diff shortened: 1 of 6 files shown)


tslearn/barycenters.py
```diff
@@ -223,6 +223,249 @@ def dtw_barycenter_averaging(X, barycenter_size=None, init_barycenter=None,
...
+
+    list of numpy.array of shape (barycenter_size, sz_k)
+        list of weighted :math:`W^{(k)}` matrices
+
+    References
+    ----------
+
+    .. [1] D. Schultz and B. Jain. Nonsmooth Analysis and Subgradient Methods
+       for Averaging in Dynamic Time Warping Spaces.
+       TODO: where was it published?
```
rtavenar: TODO: change that!

tslearn/barycenters.py
```diff
@@ -223,6 +223,249 @@ def dtw_barycenter_averaging(X, barycenter_size=None, init_barycenter=None,
...
+
+    float
+        Current alignment cost
+
+    References
+    ----------
+
+    .. [1] D. Schultz and B. Jain. Nonsmooth Analysis and Subgradient Methods
+       for Averaging in Dynamic Time Warping Spaces.
+       TODO: where was it published?
```
rtavenar: TODO: change that!

rtavenar:
At the moment, implementations differ when non-uniform weights are used. I will check that later.

rtavenar:
At some point, I had the idea to make a specific implem for the equal-length time series case (since vectorization can go a bit further in this case): I have tried it locally and the benefit is so tiny (in terms of computation time) that I have decided not to include it (to limit maintenance burden).

rtavenar:
In the last commit, I fixed the bug related to weights. Now things should be ok. Next on the TODO list for this PR could be (not sure yet) to consider implementing other variants for barycenter computations (cf #44).

PR #57, opened 2018-08-06
Added save model functionality for shapelets

gio8tisu:
Added get_config in LocalSquaredDistanceLayer class and a couple wrappers to keras' save/load_model functionalities.

```python
"""Train shapelet model, save, load and predict."""

import pickle
from tslearn import shapelets

# Create mock model.
shp_clf = shapelets.ShapeletModel({10:5}, max_iter=1)
shp_clf.fit(X_train, y_train)

# Save it using custom method.
shp_clf.save('shapelet_model.h5')
# You probably also want to save/pickle the label_binarizer.
with open('label_binarizer.pkl', 'wb') as f:
    pickle.dump(shp_clf.label_binarizer, f)

# Delete trained model and load it from file.
del shp_clf
new_shp_clf = shapelets.load_model('shapelet_model.h5')

# Predict class probabilities.
X = new_data
y_probs = new_shp_clf.predict(X) 

# Load the label_binarizer to get the predicted label.
with open('label_binarizer.pkl', 'rb', f):
    label_binarizer = pickle.load(f)

y_pred =  label_binarizer.inverse_transform(y_probs)  # predicted labels for input data X
```

PR #160, opened 2019-10-15
[WIP] Save models to hdf5 and other formats

kushalkolar:
Hi,

I thought it would be useful to save the KShape model without pickling. I implemented a simple **to_hdf5()** method for saving a KShape model to an hdf5 file and **from_hdf5()** for reloading it so that predictions can be done with the model.

Changes to the KShape class:

- the class attribute "model_attrs" is a list of attributes that are sufficient to describe the model.
- **to_dict()** method packages the model attributes and params to a dict.
- **to_hdf5()** and **from_hdf()** can be used to save/load the model to/from hdf5 files.
- put instance attributes in constructor

An hdftools module is added to handle saving a dict of numpy arrays to an hdf file.

Usage:

```
ks.to_hdf5('/path/to/file.h5')
model = KShape.from_hdf5('path/to/file.h5')
```


kushalkolar:
I think it would be useful for any tslearn model where it makes sense.

Using the structure shown here for KShape, I can create a base class that uses a `model_attrs` list with a `to_dict()` method for packaging models . The dict can then be used to export to various formats, hdf, json etc. This should work for all tslearn models where a collection of instance attributes are sufficient to describe the model.

pickling is the only answer I could find for how sklearn does it (and that's what their docs say too https://scikit-learn.org/stable/modules/model_persistence.html). Thought there are issues & limitations with pickling of course (some are outlined in their docs).

rtavenar:
Yes, it could be a great idea to have a base class for serializable models so that code does not have to be duplicated everywhere!

rtavenar:
Ideally, this could be a more general solution than #57 for example.

kushalkolar:
Sorry I haven't had time to work on implementing this yet. Yes I will definitely try to make a generalized solution to this.

rtavenar:
Hi @kushalkolar 

Any update on that project? Or should I take the lead on this?

tslearn/clustering.py
```diff
@@ -915,8 +918,12 @@ class KShape(BaseEstimator, ClusterMixin,
     .. [1] J. Paparrizos & L. Gravano. k-Shape: Efficient and Accurate
        Clustering of Time Series. SIGMOD 2015. pp. 1855-1870.
     """
+
+    model_attrs = ['inertia_', 'cluster_centers_', '_X_fit', '_norms', '_norms_centroids']
```
kushalkolar: Define a class attribute which is the list of model attributes that are sufficient to describe the model. BaseModelPackage will throw an exception if it doesn't find this class attribute.

kushalkolar:
Hi @rtavenar I've updated the pull request with what I have so far. There's a base class for packaging models that is currently implemented by the KShape class, however I'm not familiar with other tslearn models so I'm not exactly sure which attributes need to be packaged.

To implement the BaseModelPackage in other models you'll need to do this (also highlighted in the code comments):

1. Inherit from tslearn.bases.BaseModelPackage instead of BaseEstimater (BaseModelPackage subclasses BaseEstimator)

2. Define a class attribute which is the list of model attributes that are sufficient to describe the model. BaseModelPackage will throw an exception if it doesn't find this class attribute.

3. Define all parameter and model instance attributes within __init__() so that they can be easily set with **kwargs by the BaseModelPackage.

4. Have a simple method that checks if the model is fitted. Usually check_is_fitted() with some args.

This shows how it can be used, just Model.to_<format> and Model.from_<format> :
https://nbviewer.jupyter.org/github/kushalkolar/tslearn/blob/master/notebooks/save_load_kshape.ipynb

Not sure if this the most elegant way to do it so feel free to make changes!

rtavenar:
Hi @kushalkolar 

This looks really great!

My first question would be: why do you rely on a class attribute `model_attrs` rather than relying on the [`get_params`](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.get_params) method defined in `BaseEstimator`?
Or maybe we could have something in-between like: if `model_attrs` is defined, then use it, otherwise, rely on what is returned by `get_params`?

Also, your point number 3: "Define all parameter and model instance attributes within init() so that they can be easily set with **kwargs by the BaseModelPackage." is not in line with `scikit-learn` basics that forces the following:
- in `__init__`: all hyper-parameters are set and their name are in the form `param_name`
- in `fit`: all model parameters are set and their names are in the form `param_name_` 

Maybe we could use [`set_params`](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.set_params) here, and do something like:
1. Initialize a model with all parameters set to default
2. Use `set_params` to set all parameters to the values loaded from disk

What do you think about these options?

kushalkolar:
Hi @rtavenar Thanks for the feedback!

`get_params()` doesn't always return model parameters that are necessary for using the model in future predictions.

For example, with **KShape**  `_norms`, `_norms_centroids` and `cluster_centers_` are necessary for predictions but aren't returned by `get_params()`

Perhaps we can use `get_params()` to get all hyper-parameters and have another way to get model parameters which are necessary & sufficient to use that model for predictions?

rtavenar:
Hi @kushalkolar 

You are right. We could use get params to build the model, but we need something else for parameters set at fit time.

kushalkolar:
@rtavenar 

I added a mandatory `get_model_params()` method which returns a **dict** of all model parameters, which are combined with what's returned by `get_params()` and saved in a dict.

To load the model, the class is instantiated with the hyper-parameters and then all model parameters are set.

In the KShape class I also changed the name of `_norms` and `_norm_centroids` to `norms_` and `norms_centroids_` to conform to sci-kit learn.

Let me know what you think.

rtavenar:
This is great! I would assume that now, you do not need to add `model` as an extra parameter to `KShape` class init, do you?

I think we are getting close to a very easy-to-use version of your suggestion, so I could probably start an actual code review as soon as soon as you tell me your code is ready for review.

kushalkolar:
@rtavenar yup exactly. You just need to inherit from `BaseModelPackage` and add  `get_model_params()` and `is_fitted()` methods.

I'm having a weird issue with json, as soon as that's fixed I'll let you know and you can review!

kushalkolar:
@rtavenar 
So all formats work now including json.

You could review `BaseModelPackage` in `tslearn.bases` and `tslearn.hdftools`.

I have implemented it for KShape and I'll make the necessary modifications to other models after you're fine with `BaseModelPackage` and `hdftools` if that's ok for you? Implementation in other models just requires adding `if_fitted()` and `get_model_params()` methods, and renaming model parameters according to sklearn standards if necessary (trailing underscores).

tslearn/bases.py
```diff
@@ -0,0 +1,217 @@
...
+from sklearn.base import BaseEstimator
+from . import hdftools
+import json
+import pickle
+import numpy as np
+
+
+class BaseModelPackage(BaseEstimator, metaclass=ABCMeta):
+    @abstractmethod
+    def is_fitted(self) -> bool:
```
rtavenar: Type annotations would break Python2 compatibility. Though I know Python2 is not very important, I would prefer we keep compatibility when possible.

rtavenar: Also, as the method will be implemented by all models to be serialized, it would be nice to make those function protected members of the class so that they do not appear in their docs (they are useless for the end user).

Would it be possible to rename it to `_is_fitted`?

tslearn/bases.py
```diff
@@ -0,0 +1,217 @@
...
+        sklearn.utils.validation.check_is_fitted
+
+        Returns
+        -------
+        bool
+        """
+        pass
+
+    @abstractmethod
+    def get_model_params(self) -> dict:
```
rtavenar: Same comments as above

tslearn/bases.py
```diff
@@ -0,0 +1,217 @@
...
+        """Get model parameters that are sufficient to recapitulate it."""
+        pass
+
+    def to_dict(self, arrays_to_lists=False) -> dict:
```
rtavenar: type annotation

tslearn/bases.py
```diff
@@ -0,0 +1,217 @@
...
+             'model_params': self.get_model_params()}
+
+        # This is just for json support to convert numpy arrays to lists
+        if arrays_to_lists:
+            d['model_params'] = BaseModelPackage._listify(d['model_params'])
+
+        return d
+
+    @staticmethod
+    def _listify(model_params) -> dict:
```
rtavenar: type annotation

tslearn/bases.py
```diff
@@ -0,0 +1,217 @@
+from abc import ABCMeta, abstractmethod
+from sklearn.base import BaseEstimator
+from . import hdftools
```
rtavenar: Should it be added to `requirements.txt`?

rtavenar: Up to now, we use absolute imports (we should maybe change it at some point)

tslearn/clustering.py
```diff
@@ -23,7 +23,7 @@
 from tslearn.utils import (to_time_series_dataset, to_time_series,
                            ts_size, check_dims)
 from tslearn.cycc import cdist_normalized_cc, y_shifted_sbd_vec
-
+from .bases import BaseModelPackage
```
rtavenar: Please use absolute import

rtavenar:
Hi @kushalkolar 

This is great work! I have added a few remarks in the code that should not take too long to take into account.

Four last things:

* [ ] Your documentation of this feature should be added as a new example in the gallery of examples (cf `tslearn/docs/examples/` folders for other gallery examples)
* [ ] The serialization functionality should be tested in a dedicated file in `tslearn/tests`
* [ ] Add this functionality to new models (and test the functionality for all models): for this, I can provide a list of models that should not be too complicated to deal with (and we could keep more complicated ones for a later `tslearn` version) -> let me know
* [ ] Once all this done, you should update the `CHANGELOG.md` to document your changes in the section `## [Towards v0.3.0]`

rtavenar:
Ah, and also: you should aim at merging to `dev`, not `master`, so that you can see the updated code (the one that will be integrated to tslearn v0.3)

tslearn/bases.py
johannfaouzi: Using relative imports is better practice.

rtavenar: I know but my point is that we should change all imports at once, one day. But OK, this point is probably not important / counter-productive.

johannfaouzi:
I only looked briefly at the PR but what I can say is that:
* `BaseModelPackage` has a lot of public methods, I'm afraid that they will all be displayed in the API documentation (not totally sure).
* For scikit-learn estimators, the common practice is to use `joblib.dump` and `joblib.load` to save and load estimators. Joblib has been optimized to save and load Python objects that contain `numpy.array` objects. I would stick to these two functions for pure numpy/scikit-learn estimators.
* For tensorflow/keras estimators, I don't know what would be the best way to handle this: either adding specific methods for these estimators (and have private functions that help if there is common code to do so), or adding two public functions to save and load tensorflow/keras estimators.

kushalkolar:
@rtavenar and @johannfaouzi Thanks for your comments!

@rtavenar 
1. I modified as per your comments.
2. Yes it'd be great if you could provide a list of models for which I can add this functionality to and I'll work on the tests and docs.

@johannfaouzi 
1. The only methods which would now appear in the docs are the `to_<format>` and `from_<format>`.
2. `joblib.dump` and `joblib.load` use pickle in the end as far as I know, I could add helper methods to simplify the saving/loading of models using joblib's pickle instead. 
Having the option to serialize to other formats can be useful for other users (hdf5 is quite common in our field)

rtavenar:
> @rtavenar and @johannfaouzi Thanks for your comments!
> 
> @rtavenar
> 
> 1. I modified as per your comments.

Some remarks left:

* `notebooks/save_load_kshape.ipynb` should be removed and a gallery example with similar content should be added to `tslearn/docs/examples`: have a look at other examples there to see how to write those
* you should test that dumping + loading a model gives equivalent results than the original model for each class that inherits from `BaseModelPackage`: add this test file to `tslearn/tests`

> 2. Yes it'd be great if you could provide a list of models for which I can add this functionality to and I'll work on the tests and docs.

Here is a tentative list:

* all 3 classes from the `tslearn.clustering` module
*  `KNeighborsTimeSeries` from `tslearn.neighbors` and check that `KNeighborsTimeSeriesClassifier` also gets the feature through inheritance
* all estimators from `tslearn.piecewise`

The other classes might require more work. I'd like to work on the Shapelet class since it is a feature that is often asked for in GitHub Issues, but I'm afraid I won't have time to do so in a short term perspective, and I'd prefer if we could merge this PR soon, so let's keep that one for a later `tslearn` version (0.4 maybe?).

> 
> @johannfaouzi
> 
> 1. The only methods which would now appear in the docs are the `to_<format>` and `from_<format>`.

This is great!

> 2. `joblib.dump` and `joblib.load` use pickle in the end as far as I know, I could add helper methods to simplify the saving/loading of models using joblib's pickle instead.
>    Having the option to serialize to other formats can be useful for other users (hdf5 is quite common in our field)

And it is probably necessary for keras models (and hence for `Shapelet` models as they are implemented now).

Thanks for the great work @kushalkolar !

kushalkolar:
@rtavenar Thanks! BTW is there a timeline for v0.3.0? I'll aim to get this done at the end of the week.

rtavenar:
That would be great ! Have you seen my message on the other repo ? Could you answer there ?

rtavenar:
By the way @kushalkolar , you should have a look at Travis CI logs for your PR, there are a few errors there: https://travis-ci.org/rtavenar/tslearn/builds/634908936

* [ ] Code should be valid for Python 2.7, 3.5, 3.6 (only valid for 3.6 at the time)
* [ ] Even for 3.6, tests break because estimators should be valid `sklearn` estimators, but tests fail since some parameters are set during `__init__` while they should be set during `fit`

rtavenar:
@kushalkolar : note that you still have issues with tests:

* failing test in `TimeSeriesScalerMinMax` is not your fault and has been fixed in `dev`, so no worry for this one
* you seem to have a Python2.7 compatibility issue
* float128 seem to be only accessible with python 3.6+

https://travis-ci.org/rtavenar/tslearn/builds/639309691

tslearn/hdftools.py
```diff
@@ -0,0 +1,152 @@
...
+                # otherwise h5py doesn't restore the right dtype for str types
+                h5file[path + key].attrs['dtype'] = item.dtype.str
+
+            # other types
+            else:
+                h5file[path + key] = item
+                # h5file[path + key].attrs['dtype'] = item.dtype.str
+
+        # single pieces of data
+        elif isinstance(item, (str, bytes, int, float, np.int, np.int8,
```
johannfaouzi: You can use parent classes if you don't want to add them all manually (e.g. replacing np.int, ..., np.int64 with np.integer), but there are some differences (for instance unsigned integers):
https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html

tslearn/tests/test_serialize_models.py
```diff
@@ -0,0 +1,92 @@
...
+from tslearn import hdftools
+from tslearn.clustering import KShape
+from tslearn.datasets import CachedDatasets
+from tslearn.preprocessing import TimeSeriesScalerMeanVariance
+
+
+tmp_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'tmp')
+
+
+os.makedirs(tmp_dir, exist_ok=True)
```
johannfaouzi: `exist_ok` has been added in Python 3.2

johannfaouzi:
@rtavenar Do you know why the CI is displayed as successful, although it is not? #177 was merged although there are some errors...
https://travis-ci.org/rtavenar/tslearn/builds/634449123

rtavenar:
@johannfaouzi Yep, I noticed that this morning and updated `dev` branch accordingly, this is why I told @kushalkolar not to worry about these `TimeSeriesScalerMinMax` errors (they will be fixed by merging to dev).

tslearn/bases.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # set all model params
+        for p in model_params.keys():
+            setattr(inst, p, model_params[p])
+
+        return inst
+
+    def to_hdf5(self, path):
+        """
+        Save this model as an HDF5 file.
```
rtavenar: Let's use a uniform phrasing for all `to_*` and `from_*` methods. I would suggest:
* Save model to a XXX file
* Load model from a XXX file

tslearn/bases.py
```diff
@@ -0,0 +1,236 @@
...
+    def to_pickle(self, path):
+        """
+        Save as a pickle. Not recommended for interoperability.
+
+        Parameters
+        ----------
+        path : str
+            Full file path.
+
+        Returns
```
rtavenar: For all `to_*` methods, do not have a `Returns` section since they do not return anything.

tslearn/tests/test_serialize_models.py
```diff
@@ -0,0 +1,178 @@
...
+
+    d2 = hdftools.load_dict(fname, 'data')
+
+    for k in d2.keys():
+        numpy.testing.assert_equal(d[k], d2[k])
+
+
+def _check_not_fitted(model):
+    # not serializable if not fitted
+    with pytest.raises(NotFittedError):
```
rtavenar: Here, shouldn't the `with` block be located inside the loop such that we ensure all formats would raise an exception if the model is not fitted?

tslearn/tests/test_serialize_models.py
```diff
@@ -0,0 +1,178 @@
...
+        for p in hyper_params .keys():
+            numpy.testing.assert_equal(getattr(model, p), getattr(sm, p))
+
+    clear_tmp()
+
+
+def test_global_alignment_kernel_kmeans():
+    seed = 0
+    numpy.random.seed(seed)
+    X_train, y_train, X_test, y_test = CachedDatasets().load_dataset("Trace")
```
rtavenar: Here, it would be better practice to rely on fully synthetic data (generated from `numpy.random.rand` for example, with adequate dimensions) so that you fully control the shapes and do not rely on other parts of the code for the test.

Same comment holds for tests related to all estimators.

kushalkolar:
Thanks for the help! 

Almost finished with this, a few questions:

If I understand `GlobalAlignmentKernelKMeans` correctly, `_X_fit` is required for the model in order to make future predictions? Unlike `TimeSeriesKMeans` and `KShape`.

`KNeighborsTimeSeries` also doesn't seem to make sense without the original `_X_fit`.

Now `TimeSeriesKMeans` and `KShape` do not store `_X_fit` in the serialized model since some users might have used a very large training dataset.

Continuing to add this to all the models right now.

rtavenar:
Yes, for these two models, the training data needs to be known.

But your point is important: could you, for these models, have a `Note` section in the class docstring that states that saving these models to disk can be space-consuming due to the need to store training data as a model attribute?

kushalkolar:
I'm having a weird issue with `KNeighborsTimeSeries` https://travis-ci.org/rtavenar/tslearn/jobs/639667444#L797

`KNeighborsTimeSeries` does not define a `_ts_fit` anywhere, and line 250 of `tslearn.neighbors` is a docstring 

I'm not able to reproduce this with pytest on my local machine.

rtavenar:
Hi,

The thing is that `_ts_fit` is only set in some cases (depending on the metric). Maybe one way to test if the model has been fit is to have a check similar to what is done in `kneighbors`:

```python
        if self.metric in VARIABLE_LENGTH_METRICS:
            check_is_fitted(self, '_ts_fit')
        else:
            check_is_fitted(self, '_X_fit')
```

Similarly, the list of attributes to be saved should depend on the same condition

johannfaouzi:
I can't reproduce this either locally.

But looking at the error, we see that the value of the `metric` is `'precomputed'`
https://travis-ci.org/rtavenar/tslearn/jobs/639667446#L827-L828

but `knn` is defined here:
https://github.com/kushalkolar/tslearn/blob/e853138624a9c77f42976ee9df6b6ae22a2b226f/tslearn/tests/test_serialize_models.py#L197
so the `metric` parameter should have the default value, that is `'dtw'`.

johannfaouzi:
Well I'm confused, the loaded estimator should not have the `'precomputed'` value for `metric` but `'dtw'` instead and this is why this error is raised.

johannfaouzi:
Can't promise it will work but you can try to change the code so that `self.metric` is never changed:
https://github.com/kushalkolar/tslearn/blob/e853138624a9c77f42976ee9df6b6ae22a2b226f/tslearn/neighbors.py#L342-L344

https://github.com/kushalkolar/tslearn/blob/e853138624a9c77f42976ee9df6b6ae22a2b226f/tslearn/neighbors.py#L359-L360

However, this is only for Classifier and Regressor, which does not seem to be concerned with the current failing test...

Though I don't get why the error is raised from either Classifier and Regressor although KneighborsTimeSeries is the only estimator used.

johannfaouzi:
Maybe a hint about what this issue comes from:
https://github.com/rtavenar/tslearn/blob/7f46f489475c8bc4e6d9289763103bccbe2c235e/tslearn/neighbors.py#L250-L254

It looks like the executed code comes from the `dev` branch. I don't know which branch is ahead/behind, but doing a `git pull` or a `git rebase` may solve this, because the number of modified files is different from this PR only.

kushalkolar:
@johannfaouzi Thanks! I tried a git rebase and it got quite messy. I think I'll just move my changes to a new PR that branches from `dev`.

kushalkolar:
Closing because #183

Issue #132, opened 2019-08-24
pip install fails


daltonfury42:
Looks like setup.py is importing cython, which is a dependency:

> Collecting tslearn
>   Downloading https://files.pythonhosted.org/packages/af/f9/7225f4b1e7ba482b10b544cecf50a0a92bfbbde4ec30fdce824a867ad9ab/tslearn-0.1.29.tar.gz (587kB)
>     100% |████████████████████████████████| 593kB 766kB/s 
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>       File "/tmp/pip-build-4bo4i_6k/tslearn/setup.py", line 5, in <module>
>         from Cython.Distutils import build_ext as _build_ext
>     ModuleNotFoundError: No module named 'Cython'
>     
>     ----------------------------------------
> Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-4bo4i_6k/tslearn/
>

GillesVandewiele:
This can indeed happen sometimes... Could you try `pip install Cython` first and see if it works?

daltonfury42:
There is a way around this, though: https://github.com/LeeKamentsky/python-javabridge/pull/131/files#diff-2eeaed663bd0d25b7e608891384b7298

rtavenar:
`cython` and `numpy` are now declared as build time requirements for PyPI and github-based installs (basically source-based installs) in the docs.

`conda` install should make things easier for those who can use it.

And #150 will be used to advance on the question of releasing wheels for a wide range of platforms on PyPI.

Issue #84, opened 2019-02-05
No module named 'numpy'


jen0liu:
I'm using tslearn to launch an app on heroku. When installing packages via requirements.txt, it always gave this error message "No module named 'numpy'", corresponding to this line(https://github.com/rtavenar/tslearn/blob/master/setup.py#L2).  I'm able to install numpy first then add tslearn with heroku push. I guess these environments don't have numpy pre-installed so we probably need to handle dependencies differently?

rtavenar:
Hi @jen0liu and thanks for your feedback.

Hum, you're right, it seems like a real issue. I am not an expert (at all) in setups, so if anyone can help on that one, that would be great. 

Is there a way to remove the `numpy` import in `setup.py` while still having the same behaviour?

rth:
Well numpy is a build dependency, so it should be installed in the environment when `setup.py` is run. There is no real a way around it, short of building binary wheels and adding some conditional logic in `setup.py` on whether there package is built from source or not, but we are not there.

Overall it feels more like an issue with `heroku` configuration, rather than a bug in tslearn.

rtavenar:
So, @rth , if I understand you well, the way we deal with the `numpy` dependency in `setup.py` is OK? What do you think about the related discussion about `cython` in #132 ?

rth:
Well there are two choices,
 - say that numpy (and Cython when building from sources) needs to be explicitly installed before installing tslearn (and fail if it's not present). This is what scikit-learn is [currently doing](https://github.com/scikit-learn/scikit-learn/blob/c4f372c2de33db77e4dbc5a136d1443a20a3cb45/setup.py#L278). 
 - Import numpy and scipy in a custom build_ext class as suggested in that other issue https://github.com/LeeKamentsky/python-javabridge/pull/131/files#diff-2eeaed663bd0d25b7e608891384b7298 . Though this whole part is non trivial and a bit hacky https://github.com/scikit-learn/scikit-learn/issues/7867#issuecomment-261488725. The real solution for this is [PEP 518](https://www.python.org/dev/peps/pep-0518/) with defining build dependencies in `pyproject.toml`, but it's not fully adopted in scientific python packages yet.

So overall, there may be ways to make this smoother but it's not that straightforward and anyway saying that numpy, cython need to be installed when building from sources is not a huge constraint I think. The fact that most people on windows won't have a working compiler by default, is a bigger concern IMO.

[...]

PR #164, opened 2019-10-30
Add pyproject.toml

FlorisHoogenboom:
In the past there have been some issues raised on not being able to easilly install `tslearn` via `pip` (see e.g. #84 and #132). Even though these issues are closed there is an easy fix for this given by [PEP-518](https://www.python.org/dev/peps/pep-0518/). This PR adds a `pyproject.toml` file which makes it possible to install `tslearn` from source by simply using `pip`.

I'm curious to hear any feedback on the idea of including this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,9 +121,13 @@ If you use `tslearn` in a scientific publication, we would appreciate citations:
 
 ```bibtex
 @misc{tslearn,
- title={tslearn: A machine learning toolkit dedicated to time-series data},
- author={Tavenard, Romain and Faouzi, Johann and Vandewiele, Gilles},
- year={2017},
- note={\url{https://github.com/rtavenar/tslearn}}
+      title={tslearn: A machine learning toolkit dedicated to time-series
+             data},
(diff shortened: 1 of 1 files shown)


rth:
+1 to use `pyproject.toml` to imporve installation experience.

rth:
Generally LGTM, but let's fix CI first in https://github.com/rtavenar/tslearn/pull/190 then merge master into this PR to check that all tests work as expected.

rtavenar:
> Generally LGTM, but let's fix CI first in #190 then merge master into this PR to check that all tests work as expected.

OK, thanks for your feedback on this point, and I agree that merging #190 is the first thing to do: will do asa Travis passes for Python3*

rtavenar:
Is there something to be added to the docs so that this `pyproject.toml` is taken into account at build time?

rth:
> Is there something to be added to the docs so that this pyproject.toml is taken into account at build time?

No it should be automatic with recent pip versions.

PR #175, opened 2020-01-07
Disregard NaNs in TimeSeriesScalerMeanVariance

eliwoods:
If a time series with any `NaN` values is passed to `TimeSeriesScalerMeanVariance.fit_transform()` the transformed time series returns as all `NaN`

E.g:
```
In [1]: from tslearn.preprocessing import TimeSeriesScalerMeanVariance                                                                

In [2]: from numpy import nan                                                                                                         

In [3]: TimeSeriesScalerMeanVariance().fit_transform([0, nan, 6])                                                                     
Out[3]: 
array([[[nan],
        [nan],
        [nan]]])
```

This could be fixed by using `numpy.nanmean()` and `numpy.nanstd()` in `TimeSeriesScalerMeanVariance.transform` instead of the current implementation. This also makes tslearn in line with sklearn in terms of NaN's and preprocessing: https://github.com/scikit-learn/scikit-learn/issues/10404

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ install:
 - pip install pytest
 - if [[ "$NUMBA_DISABLE_JIT" == 1 ]]; then pip install coverage==4.3 "pytest-cov<2.6"; fi
 - pip install argparse
-- pip install codeclimate-test-reporter
+- pip install codeclimate-test-reporter codecov
 - rm -f tslearn/*.c
 - python setup.py build_ext --inplace
 script:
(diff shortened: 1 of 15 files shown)


rtavenar:
Hi @eliwoods 

Thanks for the suggestion, this makes a lot of sense.

I would suggest the following improvements:

* [ ] Do the same for `TimeSeriesScalerMinMax`
* [ ] Add doctest examples to make the behaviour explicit for end-users
* [ ] Change the destination branch to `dev` such that your work can be included in the next `tslearn` version to come

Once all this is done, one last thing would be to document your contribution in `CHANGELOG.md`.

PR #190, opened 2020-03-03
Azure pipelines setup

rth:
Working on setting up CI with Azure Pipelines

### DIFF
--- a/tslearn/tests/test_utils.py
+++ b/tslearn/tests/test_utils.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pickle
 import os
+from numpy.testing import assert_allclose
 
 import pytest
 
@@ -11,9 +12,8 @@ __author__ = 'Romain Tavenard romain.tavenard[at]univ-rennes2.fr'
 
 def test_arraylike_copy():
     X_npy = np.array([1, 2, 3])
-    np.testing.assert_allclose(tslearn.utils._arraylike_copy(X_npy), X_npy)
-    np.testing.assert_allclose(tslearn.utils._arraylike_copy(X_npy) is X_npy,
-                               False)
+    assert_allclose(tslearn.utils._arraylike_copy(X_npy), X_npy)
+    assert_allclose(tslearn.utils._arraylike_copy(X_npy) is X_npy, False)
 
 
 def test_save_load():
@@ -23,15 +23,15 @@ def test_save_load():
     tslearn.utils.save_timeseries_txt("tmp-tslearn-test.txt", dataset)
     reloaded_dataset = tslearn.utils.load_timeseries_txt(
         "tmp-tslearn-test.txt")
-    np.testing.assert_allclose(dataset, reloaded_dataset)
+    assert_allclose(dataset, reloaded_dataset)
 
     dataset = tslearn.utils.to_time_series_dataset([[1, 2, 3, 4], [1, 2, 3]])
     tslearn.utils.save_timeseries_txt("tmp-tslearn-test.txt", dataset)
     reloaded_dataset = tslearn.utils.load_timeseries_txt(
         "tmp-tslearn-test.txt")
     for ts0, ts1 in zip(dataset, reloaded_dataset):
-        np.testing.assert_allclose(ts0[:tslearn.utils.ts_size(ts0)],
-                                   ts1[:tslearn.utils.ts_size(ts1)])
+        assert_allclose(ts0[:tslearn.utils.ts_size(ts0)],
+                        ts1[:tslearn.utils.ts_size(ts1)])
 
 
 def test_label_categorizer():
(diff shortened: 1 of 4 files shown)


rth:
So there are several issues while trying tests,
 - Python 2.7 the dependencies (and in particular numba) won't install, because of version mismatch https://github.com/rtavenar/tslearn/issues/191 I'm not sure it's worth spending time on this, dropping python 2.7 might be a solution.
 - Python 3.5: some tests are very slow in CI. Locally tests run for me in a minute or so, while in CI some individual tests timeout after 90s. I believe this is related to TensorFlow, maybe due to CPU oversubscription: in other CI the detected number of CPU used to be much larger than the number of available CPU i.e  2. This is usually addressed by setting environement variables, but it looks like [Tensorflow 1.15  dropped support for OMP_NUM_THREADS](https://github.com/tensorflow/tensorflow/releases/tag/v1.15.0). This needs more investigation, meanwhile maybe 

I'm planning to add other OSs once Linux test pass.

rtavenar:
Hi @rth 

Thanks for this. Yeah I think for the moment you can move forward with Python 2 disabled and skip `ShapeletModel`-related tests. I can try to have a deeper look at the specific problem related to TF.

GillesVandewiele:
Some of the tests in check_estimators (especially those with ShapeletModels as @rtavenar pointed out) are indeed rather slow. This is because the default hyper-parameters of `ShapeletModel` will have a long runtime (e.g. max_iter of 10000...). Perhaps we should look to extend it with early stopping or smth like that, as it converges quickly on the benchmark datasets.

rth:
>  This is because the default hyper-parameters of ShapeletModel will have a long runtime (e.g. max_iter of 10000...).

Thanks! I see, maybe it could be worth setting it to a lower value in tests similarly to what is done in https://github.com/scikit-learn/scikit-learn/blob/c3969f43d9e4ba66d7f211d3835bb4cbe616c12b/sklearn/utils/estimator_checks.py#L481

For now I will skip them.

rth:
OK so I skipped ShapeletModel tests for now and all tests run in less than 4 min. We can add wheel building (and Python 3.8 https://github.com/rtavenar/tslearn/issues/193) in a follow up PR. Current setup is a bit repetitive but it's a start.

GillesVandewiele:
That is a great idea, and that reminded me that we actually tried to do something similar:
https://github.com/rtavenar/tslearn/blob/master/tslearn/tests/test_estimators.py#L168

But it seems to not be working as it should?

rth:
> that reminded me that we actually tried to do something similar:

Indeed, I opened https://github.com/rtavenar/tslearn/issues/194 to discuss how to do that without redefining check_estimator, but generally yes that should have worked.

rtavenar:
OK, so the nice thing (correct me if I'm wrong) is that here you discarded Shapelet tests only for Azure pipelines, so we can still use Travis in parallel for the moment until we have found a fix for Shapelet timeouts, right?

If this is true, I would merge and we can have a look at that #194 you opened.

tslearn/tests/test_utils.py
```diff
@@ -46,6 +48,7 @@ def test_label_categorizer():
 
 
 def test_conversions():
+    pytest.importorskip('pandas')
```
rtavenar: Hum, I have a concern here: I would prefer that we skipped only pandas-related conversion functions (ie. those involving sktime, pyflux and tsfresh) only. Maybe we could split the tests into two functions (one with pandas required, one that should pass even without pandas)?

tslearn/utils.py
```diff
@@ -770,471 +770,483 @@ def transpose_or_expand(ts):
     return to_time_series_dataset([transpose_or_expand(Xi) for Xi in X])
 
 
-try:  # Ugly hack, not sure how to to it better
-    import pandas as pd
+def to_sktime_dataset(X):
+    """Transform a tslearn-compatible dataset into a sktime dataset.
 
-    def to_sktime_dataset(X):
```
rth: Having the `try: ... except ImportError:` inside the functions is a more standard way of doing it. Done that and removed 1 indentation from all of these functions.

rth:
>  would prefer that we skipped only pandas-related conversion functions (ie. those involving sktime, pyflux and tsfresh) only. Maybe we could split the tests into two functions (one with pandas required, one that should pass even without pandas)?

OK, I split that test in multiple functions depending on dependencies.

I would be good to merge this, if only to avoid CI failures in other PRs.

tslearn/utils.py
rtavenar: OK, but now it seems that on Travis doctests fail if the dependencies are not properly installed.

rtavenar: The question is: why do these pass on Azure? Are doctests executed at all?

rth: I saw, working on a fix. No I need to add the corresponding pytest option.

rth:
OK, I added a ` conftest.py` to skip doctests that require optional dependencies (as it is done in scikit-learn). Hopefully that would fix Travis.

For now Azure doesn't run doctests indeed, I would rather add it in a follow up PR. For one thing, it would also be good to move the `doc/` folder one level up, to avoid excluding it from tests.

rth:
What's the difference between dev and master branches BTW? This is a PR into master, should it be done to dev instead?. I guess the naming of branches doesn't matter much, but it would be good to have the default branch on github to be the one in active development, so that PRs are made to the correct one..

rtavenar:
> What's the difference between dev and master branches BTW? This is a PR into master, should it be done to dev instead?. I guess the naming of branches doesn't matter much, but it would be good to have the default branch on github to be the one in active development, so that PRs are made to the correct one..

You're right. What we have right now is:

* `master` is the branch from which wheels, conda packages, and docs are built
* `dev` is a branch used to prepare next release (contains new features to appear, etc)

PR #198, opened 2020-03-24
[MRG] Early Classification for Time Series

Francois-33:
This PR implements the early classifier from [1].

François

[1] A. Dachraoui, A. Bondu & A. Cornuejols. Early classification of time
       series as a non myopic sequential decision making problem.
       ECML/PKDD 2015


tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+    def graphic(self, PATH, name, prediction_time=None):
+        plt.clf()
+        for current_cluster in range(0, self.number_cluster):
+            for current_class in range(0, self.number_classes_):
+                for current_hat_class in range(0, self.number_classes_):
+                    if current_class != current_hat_class:
+                        sum_prob = [self.pyck_[current_class][current_cluster] * item[current_class][current_hat_class]
+                                    for item in self.pyhatyck_["pyhatycks{0}".format(current_cluster)]]
+                        lab = \
+                            "y=" + str(current_class) + "yhat=" + str(current_hat_class) + "ck=" + str(current_cluster)
```
rth: Better to use string formatting,
```
lab = "y={} yhat={} ck={}".format(
        current_class, current_hat_class, current_cluster
)

File: tslearn/nonmyopic.py [hunk 2]
```diff
@@ -0,0 +1,450 @@
...
+            time_predictions.append(new_time)
+            pred_proba.append(new_proba)
+            cost_average = cost_average + new_cost
+        average_pred, variance_pred = self.avg_prediction(time_predictions)
+        pred_proba = np.reshape(np.asarray(pred_proba), (X_test.shape[0], 2))
+        auc = roc_auc_score(Y_test, pred_proba[:, 1])
+        cost_average = cost_average / X_test.shape[0]
+        return average_pred, variance_pred, auc, cost_average
+
+    def graphic(self, PATH, name, prediction_time=None):
```
rth:
```suggestion
    def graphic(self, path, name, prediction_time=None):
```

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+                    if current_class != current_hat_class:
+                        sum_prob = [self.pyck_[current_class][current_cluster] * item[current_class][current_hat_class]
+                                    for item in self.pyhatyck_["pyhatycks{0}".format(current_cluster)]]
+                        lab = \
+                            "y=" + str(current_class) + "yhat=" + str(current_hat_class) + "ck=" + str(current_cluster)
+                        plt.plot(sum_prob, label=lab)
+        plt.legend()
+        if prediction_time is not None:
+            plt.title("The prediction time is " + str(prediction_time))
+        plt.savefig(PATH + name + ".png", dpi=500)
```
rth:
```suggestion
        plt.savefig(os.path.join(path, name + ".png"), dpi=500)
```

+ add `import os.path` at the top.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+    @staticmethod
+    def accuracy_prediction(Y_test, Y_pred_test):
+
+        good_prediction = 0
+        for class_index in range(0, len(Y_test)):
+            if Y_test[class_index] == Y_pred_test[class_index]:
+                good_prediction = good_prediction + 1
+        return good_prediction / len(Y_test)
+
+    def predict(self, X_test):
```
rth: The scikit-learn conventions is,

```suggestion
    def predict(self, X):
```

Unless one cannot run this on other thing than the test dataset?

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+        Returns
+        ----------------
+        Float: the expected euclidean distance
+        """
+
+        euclidean_dist = np.linalg.norm(truncated_xt - self.cluster_.cluster_centers_[wanted_cluster, :len(truncated_xt)])
+
+        return euclidean_dist
+
+    def get_cluster_prob(self, truncated_xt):
```
rth:
```suggestion
    def get_cluster_proba(self, truncated_xt):
```
to be consistent with scikit-learn's 'predict_proba'

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+
+         Float : the silhouette score of the fitting of the clustering
+        """
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.number_cluster)
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X_train)))
+        silhouette = silhouette_score(to_time_series_dataset(X_train), c_k, metric="euclidean")
+
+        return c_k, silhouette
+
+    def fit(self, X_train, Y_train):
```
rth:
```suggestion
    def fit(self, X, y):
```

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+        self.minimum_time_stamp = minimum_time_stamp
+        self.lamb = lamb
+        self.cost_time_parameter = cost_time_parameter
+
+    def clustering(self, X_train):
+        """
+        This function compute and fit clusters for a training dataset of time series using K-means and euclidean
+        distances
+        Parameters
+        -----------------
```
rth: The length of the underline should be exactly the same as the title above, otherwise you would get sphinx warnings. Here and everywhere else.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,450 @@
...
+from tslearn.clustering import TimeSeriesKMeans, silhouette_score
+import matplotlib.pyplot as plt
+import numpy as np
+import math
+from sklearn.metrics import roc_auc_score
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--cost_time_parameter', type=float, default=0.01)
+args = parser.parse_args()
```
rth: This can probably be removed

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+            time_predictions.append(new_time)
+            pred_proba.append(new_proba)
+            cost_average = cost_average + new_cost
+        average_pred, variance_pred = self.avg_prediction(time_predictions)
+        pred_proba = np.reshape(np.asarray(pred_proba), (X.shape[0], 2))
+        auc = roc_auc_score(Y, pred_proba[:, 1])
+        cost_average = cost_average / X.shape[0]
+        return average_pred, variance_pred, auc, cost_average
+
+    def graphic(self, path, name, prediction_time=None):
```
rtavenar, line 485: This should not be included in the class. This is code you use for visualization but the class code should just consist in what is required to build your estimator, fit it to data, and predict outcome for new data.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        Y_pred = []
+
+        for t in range(0, X.shape[0]):
+            new_classif, new_time, new_proba, new_cost = self.classification(
+                X[t], self.cost_function
+            )
+            Y_pred.append(new_classif)
+        return Y_pred
+
+    def predict_completed(self, X, Y):
```
rtavenar, line 447: This should not be included here (same reason as for `graphic`).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        """
+
+        Y_pred = []
+
+        for t in range(0, X.shape[0]):
+            new_classif, new_time, new_proba, new_cost = self.classification(
+                X[t], self.cost_function
+            )
+            Y_pred.append(new_classif)
+        return Y_pred
```
rtavenar, line 445: This one should also return the time at which prediction is made, otherwise one cannot retrieve that information.
Also, `t` is a **VERY** bad variable name for something that does not correspond to a time index but rather a series index.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+
+    @staticmethod
+    def avg_prediction(time_predictions):
+
+        avgerage_time = sum(time_predictions) / len(time_predictions)
+        variance_time = np.var(time_predictions)
+        return avgerage_time, variance_time
+
+    @staticmethod
+    def accuracy_prediction(Y_test, Y_pred_test):
```
rtavenar, line 416: This should not be included here (same reason as for `graphic`).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+                    time_prediction - self.minimum_time_stamp
+                ].predict_proba(xt[:, :time_prediction])
+                loss_exit = minimum_loss
+                stop_criterion = True
+            else:
+                time_prediction = time_prediction + 1
+        return result, time_prediction, result_proba, loss_exit
+
+    @staticmethod
+    def avg_prediction(time_predictions):
```
rtavenar, line 409: This should not be included here (same reason as for `graphic`).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        tau_star = 0
+        so_far_minimum_cost = function_to_minimize(xt, tau_star)
+        for tau in range(1, end_of_time):
+            current_cost = function_to_minimize(xt, tau)
+            if current_cost < so_far_minimum_cost:
+                so_far_minimum_cost = current_cost
+                tau_star = tau
+        return tau_star + xt.shape[-1], so_far_minimum_cost
+
+    def classification(self, xt, cost_function):
```
rtavenar, line 358: Do you sometimes use a `cost_function` that is different from `self.cost_function`? If not, then it should not be an argument of the function.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+                                truncated_t + tau - self.minimum_time_stamp
+                            ][y][y_hat]
+                        )
+                cost_y = cost_y + self.pyck_[y, current_cluster] * cost_y_hat
+            cost = cost + proba_clusters[current_cluster] * cost_y
+        cost = cost + self.cost_time(t=truncated_t + tau)
+
+        return cost
+
+    @staticmethod
```
rtavenar, line 331: This one should probably not be static, and then `function_to_minimize` would not be a required argument anymore (accessible through `self.cost_function`).

Also the name of this function should be more explicit (something like `expected_optimal_prediction_time`).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+                math.exp(self.lamb * minimum_distance_to_average)
+                + math.exp(self.lamb * (minimum_distance_to_average - Delta_k))
+                + 10e-6
+            )
+            sk.append(proba_current_cluster)
+            sum_sk = sum_sk + sk[-1]
+        final_sk = [(x / sum_sk) for x in sk]
+        return final_sk
+
+    def cost_function(self, X, tau):
```
rtavenar, line 294: I would rather call this function `expected_cost` and make it private (call it `_expected_cost`) to better reflect what it does.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        Float: the expected euclidean distance
+        """
+
+        euclidean_dist = np.linalg.norm(
+            X - self.cluster_.cluster_centers_[wanted_cluster, : len(X)]
+        )
+
+        return euclidean_dist
+
+    def get_cluster_prob(self, X):
```
rtavenar, line 262: A more standard name would be `get_cluster_probas`.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        Parameters
+        ----------
+        X: vector
+            a time series that may be truncated
+
+        Returns
+        -------
+        vector : the probabilities of the given time series to be in the clusters of the model
+        """
+        average_distance, minimum_distance = self.get_avg_dist(X)
```
rtavenar, line 275: In this call you compute distances to all cluster centers, and you will recompute it later in the loop, that's a pity. I'd rather first get all distances, and from that list extract all information (`avg_dist`, `min_dist`, and all individual distances).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
+from sklearn.neural_network import MLPClassifier
+from sklearn.metrics import confusion_matrix
+from sklearn.base import BaseEstimator, ClassifierMixin
+from tslearn.utils import to_time_series_dataset
+from tslearn.clustering import TimeSeriesKMeans, silhouette_score
+import matplotlib.pyplot as plt
+import numpy as np
+import math
```
rtavenar, line 8: Probably useless, use the corresponding functions from `numpy` instead. Also, please clean the imports once you have removed methods that should not belong here.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+
+class NonMyopicEarlyClassification(BaseEstimator, ClassifierMixin):
+    """Early Classification modelling for time series using the model presented in [1]
+
+    Parameters
+    ----------
+    number_cluster : int
+        Number of clusters to form.
+
+    solver :
```
rtavenar, line 21: Here, you stick to the original Dachraoui et al's version that rely on MLP classifier. Wouldn't it be more flexible, if you just had a `classifier` parameter that would take an estimator to be used for individual classifications?

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        Seed of the MLP
+
+    maximum_iteration : int
+        Maximum number of iterations performed by the MLP
+
+    minimum_time_stamp : int
+        Earliest time at which a classification can be performed on a time series
+
+    lamb : float
+        Value of the hyper parameter lambda
```
rtavenar, line 37: Maybe your class description should be richer so that it is clear to the end user what `lambda` is.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+
+    cost_time_parameter : float
+        Parameter of the cost function of time
```
rtavenar, line 40: Same remark as above: it should be clear what this `cost_time_parameter` if you had a formula above that would describe the cost function used by this model.

tslearn/nonmyopic.py
rtavenar: Then you could build (at fit time), the attribute `classifier_` by just cloning this one as many times as required (time length of `X`).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+    clusters_ : dictionary
+        Contains the times series by clusters
+
+    pyhatyck_ : dictionary
+        Contains the probabilities of being classified as class y_hat given class y and cluster ck
+
+    indice_ck_ : list
+        Contains for each clusters the indexes of the time series in the dataset
+
+    number_classes_ : int
```
rtavenar, line 59: We do not need to expose this one as a public attribute I think.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        The number of classes in the series
+
+    len_X_ : int
```
rtavenar, line 62: We do not need to expose this one as a public attribute I think.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        The length of the time series
+
+    Examples
+    --------------------
+    >>>
```
rtavenar, line 67: TODO :)

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        Returns
+        -------
+        Vector : a vector as long as the number of series in the training dataset that indicates in which cluster the
+        series at the same index belongs
+
+         Float : the silhouette score of the fitting of the clustering
+        """
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.number_cluster)
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X)))
+        silhouette = silhouette_score(
```
rtavenar, line 116: Do we need to compute silhouette score if we use a pre-defined number of clusters? And then, we probably do not need a method for that since it only calls `TimeSeriesKMeans` build and fit.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+from sklearn.metrics import roc_auc_score
+import os.path
+
+
+class NonMyopicEarlyClassification(BaseEstimator, ClassifierMixin):
+    """Early Classification modelling for time series using the model presented in [1]
+
+    Parameters
+    ----------
+    number_cluster : int
```
rtavenar, line 18: Please use `n_clusters` as in `TimeSeriesKMeans`.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        self.indice_ck_ = []
+        self.number_classes_ = len(classes_y)
+        self.len_X_ = X.shape[1]
+        mid_X = int(X.shape[0] / 2)
+        X1 = X[:mid_X]
+        Y1 = Y[:mid_X]
+        X2 = X[mid_X:]
+        Y2 = Y[mid_X:]
+        c_k2 = c_k[mid_X:]
+        self.pyck_ = np.zeros(shape=(self.number_classes_, self.number_cluster))
```
rtavenar, line 151: pyck_ can be constructed using `scipy.sparse.coo_matrix` as done for `confusion_matrix` in scikit-learn.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.html

One important thing here is the note: 
> By default when converting to CSR or CSC format, duplicate (i,j) entries will be summed together. This facilitates efficient construction of finite element matrices and the like. (see example)

so you could have something like:

```python
weights = np.ones((self.n_classes_, self.n_clusters)) / (self.n_classes_ * self.n_clusters)
self.pyck_ = coo_matrix((weights, (y_classes, y_clusters)),
                    shape=(self.n_classes_, self.n_clusters), dtype=np.float32,
                    ).toarray()
```

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+                value_current_class = classes_y[current_class]
+                if len(Y_current_cluster) == 0:
+                    current_pyck = 0
+                else:
+                    current_pyck = len(
+                        Y_current_cluster[Y_current_cluster == value_current_class]
+                    ) / len(Y_current_cluster)
+                self.pyck_[current_class][current_cluster] = current_pyck
+
+        for i in range(self.minimum_time_stamp, X.shape[1] + 1):
```
rtavenar, line 167: Please use variable names that reflect their use, eg. :
* `i` is for time series index
* `t` is for time index
* `di` is for dimension index
* `k` is for cluster index

tslearn/nonmyopic.py
rtavenar: Also, this method should have a more explicit name, something like `_predict_at_fixed_length` (underscore at the beginning to mark a private function, that should not be called from the outside of the class).

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+                        lab = "y={} yhat={} ck={}".format(
+                            current_class, current_hat_class, current_cluster
+                        )
+                        plt.plot(sum_prob, label=lab)
+        plt.legend()
+        if prediction_time is not None:
+            plt.title("The prediction time is " + str(prediction_time))
+        plt.savefig(os.path.join(path, name + ".png"), dpi=500)
+
+    def cost_time(self, t):
```
rtavenar, line 507: This function should probably be private (call it `_cost_time` then)

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,509 @@
...
+        time_stamp = np.linspace(0, self.len_X_, num=self.len_X_)
+        return time_stamp[t] * self.cost_time_parameter
```
rtavenar, line 509: Isn't the formula simply `t * self.cost_time_parameter`? I do not see what you are doing here.

tslearn/nonmyopic.py
Francois-33: self.pyck_ is now computed using coo_matrix function but it did not remove the for loop just made it lighter.

tslearn/nonmyopic.py
Francois-33: I let the self.cost_function in case one might want to experiment new cost_function in the future. As it is for now it can be removed though.

tslearn/nonmyopic.py
Francois-33: Also this means that the classes must be expressed with the labels {0, 1, 2,...} and not for instance {-1, 1} or otherwise.

rtavenar: Well, that means you probably need a dictionary somewhere that maps actual class labels (that could even be strings, I think, in the sklearn context) to indices 0, 1, 2.

Something like:

```python 
self.labels_ = sorted(set(y_classes))
y_classes_indices = [self.labels_.index(yi) for yi in y_classes]
```

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+                                ][1]
+                                / column_sum[1],
+                            ],
+                        ]
+                    )
+                    self.pyhatyck_["pyhatycks{0}".format(k)].append(
+                        matrix_proba_current_cluster_class
+                    )
+
+    def get_cluster_probas(self, X):
```
rtavenar, line 188: This function could definitely be vectorized. I guess we need no `for` loop nor lists in here.

For distance computations:

```python
diffs = X[numpy.newaxis, :] - centroids
distances_clusters = numpy.linalg.norm(diffs, axis=(1, 2))
```

Then, similarly, `Delta_k` and `probas` can be vectorized:

```python
delta = (average_distance - distances_clusters) / average_distance
probas = 1. / (np.exp(self.lamb * minimum_distance_to_average)
               + np.exp(self.lamb * (minimum_distance_to_average - delta))
               + 1e-6)
return probas / probas.sum()
```

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        Seed of the MLP
+
+    maximum_iteration : int
+        Maximum number of iterations performed by the MLP
+
+    minimum_time_stamp : int
+        Earliest time at which a classification can be performed on a time series
+
+    lamb : float
+        Value of the hyper parameter lambda used during the computation of the cost function to evaluate the probability
```
rtavenar, line 33: Your lines should never be longer than 79 characters, otherwise PEP8 checks will fail.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        ----------
+        X: Array-like
+            a dataset of time series
+        Y: vector
+            the associated classes of the series from X_train
+        """
+        classes_y = np.unique(Y)
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.n_clusters)
+        self.classifier_ = []
+        self.clusters_ = {}
```
rtavenar, line 113: Using variable names that are too similar (`self.cluster_` and `self.clusters_`) should be avoided in order to avoid bugs.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        self.pyhatyck_ = {}
+        self.indice_ck_ = []
+        self.__n_classes_ = len(classes_y)
+        self.__len_X_ = X.shape[1]
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X)))
+        mid_X = int(X.shape[0] / 2)
+        X1 = X[:mid_X]
```
rtavenar, line 120: Here, I'd prefer to have more explicit names, something like `X_past`, `X_future`, so that the rest of the code is easier to read.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        classes_y = np.unique(Y)
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.n_clusters)
+        self.classifier_ = []
+        self.clusters_ = {}
+        self.pyhatyck_ = {}
+        self.indice_ck_ = []
+        self.__n_classes_ = len(classes_y)
+        self.__len_X_ = X.shape[1]
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X)))
+        mid_X = int(X.shape[0] / 2)
```
rtavenar, line 119: Here what you want to do is a train/valid split. You should use `sklearn`'s `train_test_split` to do that properly.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        classes_y = np.unique(Y)
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.n_clusters)
+        self.classifier_ = []
+        self.clusters_ = {}
+        self.pyhatyck_ = {}
+        self.indice_ck_ = []
+        self.__n_classes_ = len(classes_y)
+        self.__len_X_ = X.shape[1]
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X)))
```
rtavenar, line 118: The cast to time series dataset format is probably already done in the clustering estimator, no need to do it twice.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+                    X2_current_cluster = np.squeeze(X2[index_cluster, :t], axis=0)
+                    Y2_current_cluster = Y2[tuple(index_cluster)]
+                    Y2_hat = self.classifier_[-1].predict(X2_current_cluster[:, :t])
+                    self.clusters_["ck_cm{0}".format(k)].append(
+                        confusion_matrix(Y2_current_cluster, Y2_hat, labels=classes_y)
+                    )
+                    column_sum = self.clusters_["ck_cm{0}".format(k)][
+                        -1
+                    ].sum(1)
+                    for current_class in range(0, self.__n_classes_):
```
rtavenar, line 157: Can be vectorized easily: `column_sum = np.max(column_sum, 1)`, no?

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+                        if column_sum[current_class] == 0:
+                            column_sum[current_class] = 1
+                    matrix_proba_current_cluster_class = np.asarray(
```
rtavenar, line 160: This is not readable as is. Probably a limitation of `black`, please reformat it to be more human-eye friendly.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+            gives the future time for which the cost is calculated
+
+        Returns
+        --------
+        float : the computed cost
+        """
+        cost = 0
+        proba_clusters = self.get_cluster_probas(X=X)
+        truncated_t = X.shape[-1]
+        for k in range(0, self.n_clusters):
```
rtavenar, line 248: The formula is a sum of sums. Probably if your data was stored in `numpy` arrays as it should, this function could be vectorized.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        tau_star = 0
+        so_far_minimum_cost = function_to_minimize(xt, tau_star)
+        for tau in range(1, end_of_time):
+            current_cost = function_to_minimize(xt, tau)
+            if current_cost < so_far_minimum_cost:
+                so_far_minimum_cost = current_cost
+                tau_star = tau
+        return tau_star + xt.shape[-1], so_far_minimum_cost
+
+    def _predict_at_fixed_length(self, xt, cost_function):
```
rtavenar, line 295: I insist that `cost_function` should not be passed as an argument to that function: the function is private so it should not be call from outside the class. Also, if one wants to change the cost function for this estimator, the correct way would probably be to inherit from this class and change the `_cost_time` function.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        ----------
+        X_test : Array-like
+            The test dataset
+
+        Returns
+        -------
+        Vector : the predicted classes
+        """
+
+        Y_pred = []
```
rtavenar, line 359: The `sklearn` standard is to write `y` instead of `Y` for label arrays. Please change those in your code (here and in other places too).

tslearn/nonmyopic.py
rtavenar: I also think the name of this function is misleading. It is not performing classification at fixed length, but rather is performing early classification for a single time series. Its prototype should probably be something like:
`def _predict_single_series(self, Xi):`

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,370 @@
...
+        cost_function: funct
+            The cost function that will be use to achieve our classification purpose
+        Returns
+        -------
+        int: the class which is predicted
+        int : the time of the prediction
+        float : the probability used for computing the cost
+        float : the loss when classifying
+        """
+        time_stamp = xt.shape[0]
```
rtavenar, line 312: `time_stamp` is a misleading name, you should use something like `ts_length`

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,312 @@
...
+        self.classifiers_ = {
+            t: clone(self.base_classifier)
+            for t in range(self.minimum_time_stamp, X.shape[1] + 1)
+        }
+        self.pyhatyck_ = {}
+        self.indice_ck_ = []
+        self.__n_classes_ = len(y_classes_indices)
+        self.__len_X_ = X.shape[1]
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X)))
+        X_and_cluster = np.concatenate((X, c_k[:, np.newaxis]), axis=1)
```
rtavenar, line 111: `train_test_split` can take 3 arrays if needed, so there is no need for this ugly hack, just do: 

```python
X_train, X_valid, c_k_train, c_k_valid,  y_train, y_valid = train_test_split(X, c_k, y, test_size=0.5)
```

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,312 @@
...
+        X1, X2, y1, y2 = train_test_split(X_and_cluster, y, test_size=0.5)
+        X1 = X1[:, :-1]
+        c_k2 = X2[:, -1]
+        X2 = X2[:, :-1]
+        vector_of_ones = np.ones((len(X[:]),))
```
rtavenar, line 116: `vector_of_ones = np.ones((X.shape[0], ))` to prevent a useless copy

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,312 @@
...
+        X: vector
+            a time series that may be truncated
+
+        Returns
+        -------
+        vector : the probabilities of the given time series to be in the clusters of the model
+        """
+        diffs = X[np.newaxis, :] - self.cluster_.cluster_centers_[:, len(X)]
+        distances_clusters = np.linalg.norm(diffs, axis=2)
+        distances_clusters = np.asarray(distances_clusters)
```
rtavenar, line 166: No need for that, it's already an array

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,312 @@
...
+
+        Returns
+        --------
+        float : the computed cost
+        """
+        cost = 0
+        proba_clusters = self.get_cluster_probas(X=X)
+        truncated_t = X.shape[-1]
+        for k in range(0, self.n_clusters):
+            sum_pyhatyck = np.sum(self.pyhatyck_["pyhatycks{0}".format(k)][
```
rtavenar, line 200: My comment above means we should have `self.pyhatyck_` that would be an array of shape (n_classes, n_classes, n_clusters) and the operation you are performing in this for loop could then be vectorized for efficiency purpose.

tslearn/nonmyopic.py
```diff
@@ -0,0 +1,312 @@
...
+    clusters_ : dictionary
+        Contains the times series by clusters
+
+    pyhatyck_ : dictionary
+        Contains the probabilities of being classified as class y_hat given class y and cluster ck
+
+    indice_ck_ : list
+        Contains for each clusters the indexes of the time series in the dataset
+
+    __n_classes_ : int private
```
rtavenar, line 48: All that starts with a `_` is private or protected, so it should not be exposed in the docstrings.

tslearn/nonmyopic.py
rtavenar: This comment still holds.

tslearn/nonmyopic.py
rtavenar: My above comment was stupid, `[:]` does not make a copy when numpy arrays are involved, so you can chage it to my suggestion, but just for cosmetic reasons :)

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+    clusters_ : dictionary
+        Contains the times series by clusters
+
+    pyhatyck_ : dictionary
+        Contains the probabilities of being classified as class y_hat given class y and cluster ck
+
+    indice_ck_ : list
+        Contains for each clusters the indexes of the time series in the dataset
+
+    __n_classes_ : int private
```
rtavenar, line 48: This is a private attribute, it should not be documented.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        The number of classes in the series
+
+    __len_X_ : int
```
rtavenar, line 51: Same here.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        Parameters
+        ----------
+        X: vector
+            a time series that may be truncated
+
+        Returns
+        -------
+        vector : the probabilities of the given time series to be in the clusters of the model
+        """
+        diffs = X[np.newaxis, :] - self.cluster_.cluster_centers_[:, len(X)]
```
rtavenar, line 164: This is buggy, please fix and fix the following line also that is not valid (distance should be computed over axes 1 and 2 and one should get one distance per cluster in the end).

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        distances_clusters = np.linalg.norm(diffs, axis=2)
+        distances_clusters = np.asarray(distances_clusters)
```
rtavenar, line 166: Previous line already creates a numpy array, so that this line is a useless copy.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+            gives the future time for which the cost is calculated
+
+        Returns
+        --------
+        float : the computed cost
+        """
+        cost = 0
+        proba_clusters = self.get_cluster_probas(X=X)
+        truncated_t = X.shape[-1]
+        for k in range(0, self.n_clusters):
```
rtavenar, line 199: It looks like this loop cpould be vectorized if `self.pyhatyck_` was a numpy array (no dict or list involved). Please check that.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+            sum_pyhatyck = np.sum(self.pyhatyck_["pyhatycks{0}".format(k)][
+                                truncated_t + tau - self.minimum_time_stamp
+                            ], axis=0)
+            sum_global = np.dot(sum_pyhatyck, self.pyck_[:, k])
+            cost = cost + proba_clusters[:, k] * sum_global
+
+        return cost
+
+    def minimize_integer(self, end_of_time, xt):
```
rtavenar, line 208: Please change the name of this function to something more explicit.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+            the function of interest
+        xt: vector
+            the time series related to the function "funct"
+
+        Returns
+         ------
+         int : The integer in {0,...,stop} that minimizes "funct"
+         float : the so fat minimum cost
+        """
+        tau_star = 0
```
rtavenar, line 227: This function could probably be 3-4 lines of code (does not change efficiency, but seems more readable to me):

```python
costs = [self._expected_cost(xt, tau) for tau in range(end_of_time)]
tau_star = numpy.argmin(costs)
min_cost = costs[tau_star]
return tau_star, min_cost
```

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        int: the class which is predicted
+        int : the time of the prediction
+        float : the probability used for computing the cost
+        float : the loss when classifying
+        """
+        length_Xi = Xi.shape[0]
+        Xi = np.reshape(Xi, (1, length_Xi))
+        time_prediction = self.minimum_time_stamp
+        stop_criterion = False
+        while stop_criterion is False:
```
rtavenar, line 255: I would rather use a for loop (to avoid infinite loops in corner cases) here since you are basically iterating over consecutive possible prediction times, with a `break` instruction if the stopping criterion is met.
Also, it seems that cases where (i) t=max_t and (ii) current_expected_loss <= min future_expected_losses lead to the same actions so they should probably be put together in a single `if` statement to avoid code redundancy.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+            a dataset of time series
+        y: vector
+            the associated classes of the series from X_train
+        """
+        y_classes = np.unique(y)
+        self.labels_ = sorted(set(y_classes))
+        y_classes_indices = [self.labels_.index(yi) for yi in y_classes]
+
+        for idx, current_classe in enumerate(y_classes):
+            y[y == current_classe] = y_classes_indices[idx]
```
rtavenar, line 99: You should never change `X` or `y` here. If you really need to do something like that, you should first copy `y` to a new `y_` and then operate on this new one. This is because the end-user never expects that calling fit for an estimator could change its data.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.n_clusters)
+        self.classifiers_ = {
+            t: clone(self.base_classifier)
+            for t in range(self.minimum_time_stamp, X.shape[1] + 1)
+        }
+        self.pyhatyck_ = {}
+        self.indice_ck_ = []
+        self.__n_classes_ = len(y_classes_indices)
+        self.__len_X_ = X.shape[1]
+        c_k = self.cluster_.fit_predict((to_time_series_dataset(X)))
```
rtavenar, line 110: Here, `fit_predict` already casts the data to time series dataset format, no need to do it twice.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        X_and_cluster = np.concatenate((X, c_k[:, np.newaxis]), axis=1)
+        X1, X2, y1, y2 = train_test_split(X_and_cluster, y, test_size=0.5)
```
rtavenar, line 112: `train_test_split` accepts as many arrays as inputs as you want, so you can pass X, y and c_k in and get all of them (doubled) as outputs, instead of this (rather ugly :) hack.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        vector_of_ones = np.ones((len(X[:]),))
+        self.pyck_ = coo_matrix(
+            (vector_of_ones, (y, c_k)),
+            shape=(self.__n_classes_, self.n_clusters),
+        ).toarray()
+        for k in range(0, self.n_clusters):
+            self.pyhatyck_["pyhatycks{0}".format(k)] = []
+            self.indice_ck_.append(np.where(c_k == k))
+            current_sum = self.pyck_[:, k].sum()
+            for current_classe in range(0, self.__n_classes_):
```
rtavenar, line 125: You have already filled `self.pyck_` so you just need to scale it right?

```python
self.pyck_ /= self.pyck_.sum(axis=0, keepdims=True)
```

rtavenar: So, typically, not in the for loop :)

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        for idx, current_classe in enumerate(y_classes):
+            y[y == current_classe] = y_classes_indices[idx]
+
+        self.cluster_ = TimeSeriesKMeans(n_clusters=self.n_clusters)
+        self.classifiers_ = {
+            t: clone(self.base_classifier)
+            for t in range(self.minimum_time_stamp, X.shape[1] + 1)
+        }
+        self.pyhatyck_ = {}
+        self.indice_ck_ = []
```
rtavenar, line 107: Seems this is not used anymore, to be removed I think.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+        X1 = X1[:, :-1]
+        c_k2 = X2[:, -1]
+        X2 = X2[:, :-1]
+        vector_of_ones = np.ones((len(X[:]),))
+        self.pyck_ = coo_matrix(
+            (vector_of_ones, (y, c_k)),
+            shape=(self.__n_classes_, self.n_clusters),
+        ).toarray()
+        for k in range(0, self.n_clusters):
+            self.pyhatyck_["pyhatycks{0}".format(k)] = []
```
rtavenar, line 122: `self.pyhatyck_` should be a multidimensional numpy array with adequate dimensions, no need to use dicts nor lists, hence no need for a for loop here.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+            self.classifiers_[t].fit(X1[:, :t], y1)
+            for k in range(0, self.n_clusters):
+                index_cluster = np.where(c_k2 == k)
+
+                if len(index_cluster[0]) != 0:
+                    X2_current_cluster = np.squeeze(
+                        X2[index_cluster, :t], axis=0
+                    )
+                    y2_current_cluster = y2[tuple(index_cluster)]
+                    y2_hat = self.classifiers_[t].predict(
```
rtavenar, line 140: This should be done once and for all, not re-computed for all `k`

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+            current_sum = self.pyck_[:, k].sum()
+            for current_classe in range(0, self.__n_classes_):
+                self.pyck_[current_classe, k] = (
+                    self.pyck_[current_classe, k] / current_sum
+                )
+
+        for t in range(self.minimum_time_stamp, X.shape[1] + 1):
+            self.classifiers_[t].fit(X1[:, :t], y1)
+            for k in range(0, self.n_clusters):
+                index_cluster = np.where(c_k2 == k)
```
rtavenar, line 133: The fact that you sometimes have to make this a tuple, sometimes you need to squeeze, means that this indexing does not seem appropriate.
It looks to me that the following should work (assuming `c_k2` is of shape `(n, )`), but I have not tested it:

```python
indices = (c_k2 == k)
[...]
X2_cluster = X2[indices, :t]
y2_cluster = y2[indices]
```

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+                        X2[index_cluster, :t], axis=0
+                    )
+                    y2_current_cluster = y2[tuple(index_cluster)]
+                    y2_hat = self.classifiers_[t].predict(
+                        X2_current_cluster[:, :t]
+                    )
+                    conf_matrix = confusion_matrix(
+                        y2_current_cluster, y2_hat, labels=y_classes_indices, normalize="true"
+                    )
+                    np.fill_diagonal(conf_matrix, 0)
```
rtavenar, line 146: Please document why this is necessary (with a comment), because it looks weird at first sight.

tslearn/early_classification.py
```diff
@@ -0,0 +1,312 @@
...
+
+    __len_X_ : int
+        The length of the time series
+
+    Examples
+    --------------------
+
+    References
+    --------------------
+    [1] A. Dachraoui, A. Bondu & A. Cornuéjols. Early classification of time series as a non myopic sequential decision
```
rtavenar, line 59: Please do not use characters with accents (write Cornuejols for once) in python files.

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+        self.classifiers_ = {
+            t: clone(self.base_classifier)
+            for t in range(self.minimum_time_stamp, X.shape[1] + 1)
+        }
+        self.__n_classes_ = len(y_classes_indices)
+        self.__len_X_ = X.shape[1]
+        self.pyhatyck_ = np.empty((self.__len_X_ - self.minimum_time_stamp + 1,
+                                   self.n_clusters, self.__n_classes_, self.__n_classes_))
+        c_k = self.cluster_.fit_predict((X))
+        X_and_cluster = np.concatenate((X, c_k[:, np.newaxis]), axis=1)
```
rtavenar, line 105: No! you should have just `X` as the first argument of `train_test_split`.

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+        ----------
+        X: Array-like
+            a dataset of time series
+        y: vector
+            the associated classes of the series from X_train
+        """
+        y_classes = np.unique(y)
+        self.labels_ = sorted(set(y_classes))
+        y_classes_indices = [self.labels_.index(yi) for yi in y_classes]
+        y_ = y
```
rtavenar, line 91: This does not do a copy, but just creates a new pointer to the same data.

https://numpy.org/doc/1.18/reference/generated/numpy.copy.html

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+            a time series that may be truncated for which we want to have the cost
+        tau: int
+            gives the future time for which the cost is calculated
+
+        Returns
+        --------
+        float : the computed cost
+        """
+        proba_clusters = self.get_cluster_probas(X=X)
+        truncated_t = X.shape[-1]
```
rtavenar, line 177: No! Last dimension in `X` is the number of features.

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+        sum_pyhatyck = np.sum(self.pyhatyck_[truncated_t + tau - self.minimum_time_stamp], axis=1)
```
rtavenar, line 178: So in `self.pyhatyck_`, the dimension corresponding to yhat is after the one that corresponds to y? It's a bit misleading, at least you should have a comment that documents this here because it seems strange given the variable name.

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+        sum_pyhatyck = np.transpose(sum_pyhatyck)
```
rtavenar, line 179: Same remark here: if you need to transpose, it means that the ordering is not the same as in `self.pyck_` which is not natural, so it should be documented.

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+        time_prediction = self.minimum_time_stamp
+        for t in range(self.minimum_time_stamp, self.__len_X_ + 1):
+            minimum_tau, minimum_loss = self.cost_function_minimizer(
+                end_of_time=length_Xi - t, xt=Xi[:, :t])
+            if (time_prediction == length_Xi) or (minimum_tau == t):
+                result = self.classifiers_[t].predict(Xi[:, :t])
+                result_proba = self.classifiers_[t].predict_proba(Xi[:, :t])
+                loss_exit = minimum_loss
+                break
+            else:
```
rtavenar, line 234: This else is now useless (for loop) :)

tslearn/early_classification.py
```diff
@@ -0,0 +1,266 @@
...
+            a time series that is probably incomplete but that nonetheless we want to classify
+        Returns
+        -------
+        int: the class which is predicted
+        int : the time of the prediction
+        float : the probability used for computing the cost
+        float : the loss when classifying
+        """
+        length_Xi = Xi.shape[0]
+        Xi = np.reshape(Xi, (1, length_Xi))
```
rtavenar, line 224: You should not do that, because the data can be multi dimensional and then it won't work.
I think the only problem you might have is when you call `predict` and `predict_proba`, but then you should just pass `[Xi[:t]]` as input (a list made of a single series).

tslearn/early_classification.py
Francois-33: Actually the dimension corresponding to yhat is after the one of the clusters and before the one of y as self.pyhatyck is of shape (length_of_time_series - min_time_stamp, n_cluster, n_class, n_class).
From the way the model is built I thought more intuitive to order the dimensions like this.

tslearn/early_classification.py
Francois-33: I think the same answer applies here, for clarity purposes in the building of self.pyhatyck the dimensions are ordered in a way that forces here the transposition to match with those of self.pyck_. However I will investigate whether self.pyck_ could be built another way to prevent the transposition.

tslearn/early_classification.py
```diff
@@ -0,0 +1,269 @@
...
+        """
+        proba_clusters = self.get_cluster_probas(X=X)
+        truncated_t = X.shape[0]
+        sum_pyhatyck = np.sum(self.pyhatyck_[truncated_t + tau - self.minimum_time_stamp], axis=1)
+        sum_pyhatyck = np.transpose(sum_pyhatyck)
+        sum_global = np.sum((sum_pyhatyck * self.pyck_), axis=0)
+        cost = np.dot(proba_clusters, sum_global)
+        return cost + self._cost_time(truncated_t + tau)
+
+    def cost_function_minimizer(self, end_of_time, xt):
```
rtavenar, line 185: One thing I just noticed is that in this code there are several different notation for similar objects. Typically, I would tend to use `Xi` to denote a single time series in all methods, so that going through the code is easier. And use `X` only when this is a dataset (not when it is a single series), which would require to change the variable name in `get_cluster_proba` and in `_expected_cost`.

tslearn/early_classification.py
```diff
@@ -0,0 +1,269 @@
...
+            the function of interest
+        xt: vector
+            the time series related to the function "funct"
+
+        Returns
+         ------
+         int : The integer in {0,...,stop} that minimizes "funct"
+         float : the so fat minimum cost
+        """
+        costs = [self._expected_cost(xt, tau) for tau in range(end_of_time + 1)]
```
rtavenar, line 204: In fact, this seems to be the only place where `_expected_cost` is called. This means we need expected costs for all future timestamps, so probably a future step will be to have `_expected_cost` take a single input argument and return an array of expected costs (and vectorize it :). Can be kept for later, clearly, but it would make sense in my opinion.

tslearn/early_classification.py
```diff
@@ -0,0 +1,269 @@
...
+
+        Returns
+        -------
+        Vector : the predicted classes
+        """
+
+        X = check_dims(X)
+        y_pred = []
+        time_prediction = []
+        print(X.shape)
```
rtavenar, line 255: to be removed at some point :)

tslearn/early_classification.py
```diff
@@ -0,0 +1,269 @@
...
+        for i in range(0, X.shape[0]):
+            (
+                new_classif,
+                new_time,
+                new_proba,
+                new_cost,
+            ) = self._predict_single_series(X[i])
+            y_pred.append(new_classif)
+            time_prediction.append(new_time)
+        print(y_pred)
```
rtavenar, line 265: to be removed at some point :)

tslearn/early_classification.py
```diff
@@ -0,0 +1,269 @@
...
+            minimum_tau, minimum_loss = self.cost_function_minimizer(
+                end_of_time=self.__len_X_ - t, xt=Xi[:t])
+            if (t == self.__len_X_) or (minimum_tau == t):
+                #Without the transpose here the predict fails miserably
+                Xi = np.transpose(Xi)
+                result = self.classifiers_[t].predict(Xi[:t])
+                result_proba = self.classifiers_[t].predict_proba(Xi[:t])
+                loss_exit = minimum_loss
+                break
+        return result, time_prediction, result_proba, loss_exit
```
rtavenar, line 236: It seems you do not really need to return `result_proba` and `loss_exit` since they are never used by the calling functions.
In fact, maybe `result_proba` could be kept, since it could be useful if we decide to have a `predict_proba` method at some point, but the last one seems useless.

tslearn/early_classification.py
rtavenar: Also, it seems that `end_of_time` argument for `cost_function_minimizer` would no longer be required if this code was vectorized.

rtavenar:
Not sure why my last commit is not reported here (it can be seen on your repo though, and GitHub was down yesterday evening, so it may just be a consequence of that), but now the TODO list is something like:
* [ ] Add proper docstring to all methods (I will take care of the MLP classes)
* [ ] Add doctests for all methods (I will take care of the MLP classes)
* [ ] Add deeper tests for early classification class (and adapt automatic tests?)
* [ ] Add an example use-case in the gallery

@Francois-33

rtavenar:
One of the things we will have to decide is which of the common tests are expected to pass for the early classification estimators (since predict, for example does not return a single prediction array, we cannot have all the tests from `check_estimator` passing). I can try to have a look at that.

rtavenar:
For the moment, I have edited the tests a bit so that some of them are skipped for early classifiers, but now that I think about it, it seems weird and unsafe, so maybe an alternative option would be to have dedicated `predict_early` and `predict_proba_early` methods that would return both predictions and timestamps and let the usual `predict` and `predict_proba` retuen the class information alone for better compatibility with sklearn API.

rtavenar:
We have several issues at the moment:

1. The `TimeSeriesMLPClassifier` seems not to be very good on the automatic toy classification example, hence the check fails
2. There is a failing call to `np.linalg.norm` in the early classifier that is spotted by automatic checks
3. The docs won't build because the gallery example that involves kNN fails

All 3 should be investigated.

rtavenar:
Good news: it seems the upgrade from PR #232 has fixed the failing tests here (Travis tests for Python >3.5 might fail too but this will be investigated as a follow-up of #233 ).

So it seems we only need two things here:
* [ ] Update the CHANGELOG.md to document the new features introduced here
* [ ] Add a dedicated example in the gallery

rtavenar:
Woops I was too optimistic, there are still issues...

rtavenar:
This is now ready for merge to `master`. Congrats @Francois-33 !

PR #199, opened 2020-03-25
[MRG] Use python -m pip in installation instructions and add integration with other software page

johannfaouzi:
This PR:
* replaces `pip install` with `python -m pip install` in README and in the documentation
* adds a new page in the documentation illustrating the integration with other Python packages (how to transform time series from ``tslearn`` to another package and vice versa).

### DIFF
--- a/tslearn/docs/index.rst
+++ b/tslearn/docs/index.rst
@@ -80,3 +80,4 @@ From here, you can navigate to:
     reference
     auto_examples/index
     contributing
+    integration_other_software
--- a/tslearn/docs/index.rst
+++ b/tslearn/docs/index.rst
@@ -34,7 +34,7 @@ Using ``pip`` should also work fine:
 
 .. code-block:: bash
 
-    pip install tslearn
+    python -m pip install tslearn
 
 In this case, you should have ``numpy``, ``cython`` and C++ build tools
 available at build time.
@@ -48,7 +48,7 @@ repository hosted at github:
 
 .. code-block:: bash
 
-    pip install git+https://github.com/rtavenar/tslearn.git
+    python -m pip install git+https://github.com/rtavenar/tslearn.git
 
 In this case, you should have ``numpy``, ``cython`` and C++ build tools
 available at build time.
@@ -59,7 +59,7 @@ If you experiment such an issue, try installing it with the following command:
 
 .. code-block:: bash
 
-    pip install cython
+    python -m pip install cython
 
 
 before you start installing ``tslearn``.
--- a/tslearn/docs/index.rst
+++ b/tslearn/docs/index.rst
@@ -77,7 +77,7 @@ From here, you can navigate to:
     gettingstarted
     variablelength
     dtw
+    integration_other_software
     reference
     auto_examples/index
     contributing
-    integration_other_software
--- a/tslearn/docs/integration_other_software.rst
+++ b/tslearn/docs/integration_other_software.rst
@@ -1,0 +1,226 @@
+Integration with other Python packages
+--------------------------------------
+
+``tslearn`` is a general-purpose Python machine learning library for time
+series that offers tools for pre-processing and feature extraction as well as
+dedicated models for clustering, classification and regression.
+To ensure compatibility with more specific Python packages, we provide utilities
+to convert data sets from and to other formats.
+
+:func:`tslearn.utils.to_time_series_dataset` is a general function that
+transforms an array-like object into a three-dimensional array of shape
+``(n_ts, sz, d)`` with the following conventions:
+
+- the fist axis is the sample axis, ``n_ts`` being the number of time series;
+- the second axis is the time axis, ``sz`` being the maximum number of time points;
+- the third axis is the dimension axis, ``d`` being the number of dimensions.
+
+This is how a data set of time series is represented in ``tslearn``.
+
+The following sections briefly explain how to transform a data set from
+``tslearn`` to another supported Python package and vice versa.
+
+
+scikit-learn
+^^^^^^^^^^^^
+
+`scikit-learn <https://scikit-learn.org>`_ is a popular Python package for
+machine learning.
+:func:`tslearn.utils.to_sklearn_dataset` converts a data set from ``tslearn``
+format to ``scikit-learnlearn`` format. To convert a data set from
+``scikit-learn``, you can use :func:`tslearn.utils.to_time_series_dataset`.
+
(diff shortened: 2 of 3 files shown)


tslearn/docs/integration_other_software.rst
```diff
@@ -0,0 +1,226 @@
...
+``tslearn`` to another supported Python package and vice versa.
+
+
+scikit-learn
+^^^^^^^^^^^^
+
+`scikit-learn <https://scikit-learn.org>`_ is a popular Python package for
+machine learning.
+:func:`tslearn.utils.to_sklearn_dataset` converts a data set from ``tslearn``
+format to ``scikit-learnlearn`` format. To convert a data set from
```
rtavenar: typo: ``scikit-learnlearn``

rtavenar:
This is great!

Also, `pandas` is a required dependency to cast to / from the following packages:

* `pyflux`
* `sktime`
* `tsfresh`

I guess this information could be added to the page, no?

johannfaouzi:
Yes you're right! I'm going to add this information.

tslearn/docs/integration_other_software.rst
```diff
@@ -0,0 +1,230 @@
...
+
+tsfresh
+^^^^^^^
+
+`tsfresh <https://tsfresh.readthedocs.io>`_ iis a python package automatically
+calculating a large number of time series characteristics.
+:func:`tslearn.utils.to_tsfresh_dataset` and :func:`tslearn.utils.from_tsfresh_dataset`
+allow users to convert a data set from ``tslearn`` format to ``tsfresh`` format
+and vice versa.
+``pandas`` is a required dependency to use these functions.
```
GillesVandewiele: tsfresh is a dependency as well?

tslearn/docs/integration_other_software.rst
```diff
@@ -0,0 +1,230 @@
...
+
+pyflux
+^^^^^^
+
+`pyflux <https://pyflux.readthedocs.io>`_ is a library for time series analysis
+and prediction.
+:func:`tslearn.utils.to_pyflux_dataset` and :func:`tslearn.utils.from_pyflux_dataset`
+allow users to convert a data set from ``tslearn`` format to ``pyflux`` format
+and vice versa.
+``pandas`` is a required dependency to use these functions.
```
GillesVandewiele: pyflux is a dependency as well?

tslearn/docs/integration_other_software.rst
```diff
@@ -0,0 +1,230 @@
...
+
+sktime
+^^^^^^
+
+`sktime <https://alan-turing-institute.github.io/sktime/>`_ is a ``scikit-learn``
+compatible Python toolbox for learning with time series.
+:func:`tslearn.utils.to_sktime_dataset` and :func:`tslearn.utils.from_sktime_dataset`
+allow users to convert a data set from ``tslearn`` format to ``sktime`` format
+and vice versa.
+``pandas`` is a required dependency to use these functions.
```
GillesVandewiele: sktime is a dependency as well?

tslearn/docs/integration_other_software.rst
rtavenar: I do not think so. Why do you think it would?

tslearn/docs/integration_other_software.rst
```diff
@@ -0,0 +1,230 @@
...
+
+
+cesium
+^^^^^^
+
+`cesium <http://cesium-ml.org>`_ is an open-source platform for time series inference.
+:func:`tslearn.utils.to_cesium_dataset` and :func:`tslearn.utils.from_cesium_dataset`
+allow users to convert a data set from ``tslearn`` format to ``cesium`` format
+and vice versa.
+``cesium`` is a required dependency to use these functions.
```
GillesVandewiele: Bcs of this sentence I added the previous comments (Here you explicitly say that the package itself is a dependency, but nowhere else.).

rtavenar: Yep, but this is because `cesium` has its own format, hence having the lib installed is required to cast to its format. This does not apply to other libs.

GillesVandewiele: Ok loud & clear. Ignore my other comments then!

tslearn/docs/index.rst
```diff
@@ -80,3 +80,4 @@ From here, you can navigate to:
     reference
     auto_examples/index
     contributing
+    integration_other_software
```
rtavenar: I would rather add that one between DTW and API Reference. Such that we have a structure with: "important tutorials", then "api ref", then gallery and finally details for advanced users / contributors, how to cite, etc.

By the way, maybe at some point we could have a proper "Tutorials" section, not sure if that would help or not. WDYT @GillesVandewiele @johannfaouzi ?

GillesVandewiele: Yes that would most definitely be helpful! I have this in mind for the updated README (on which I'll try to work soon).

tslearn/docs/integration_other_software.rst
johannfaouzi: ``cesium`` is the only package which is also a dependency to use it because they have their own object to represent time series: `cesium.data_management.TimeSeries`. The other three rely on pandas, the first ones on numpy.

tslearn/docs/index.rst
johannfaouzi: I will change the order!

Some packages have a user guide:
* scikit-learn: http://scikit-learn.org/stable/user_guide.html
* imbalanced-learn: https://imbalanced-learn.readthedocs.io/en/stable/user_guide.html
* pyts: https://pyts.readthedocs.io/en/stable/user_guide.html
If you look for some inspiration!

rtavenar: Yes, this is the kind of things I have in mind, except that I do not think we would need a dedicated page that would just serve as a table of contents for the user guide (but I'm not sure if it is easy to trick Sphinx without a dedicated TOC page...)

johannfaouzi: You don't have to create a dedicated TOC page, but you need to reference each page somewhere so that users can find them (unless there is only one page for all the tutorials). The dedicated TOC page is just for convenience!

rtavenar:
Hey, by the way, if we want this PR to be included in 0.3.1, we'd better merge it into master, I think. @johannfaouzi could you change that?

rtavenar:
I guess we should close this one, no?

PR #201, opened 2020-03-26
[MRG] Use python -m pip in installation instructions and add integration with other software page

johannfaouzi:
This PR:
* replaces pip install with `python -m pip install` in README and in the documentation
* adds a new page in the documentation illustrating the integration with other Python packages (how to transform time series from `tslearn` to another package and vice versa).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,13 +35,13 @@ conda install -c conda-forge tslearn
 
 ## Using PyPI
 
-### Pre-requisites 
+### Pre-requisites
 
 When using PyPI, C++ build tools should be available to perform installation.
 
 Using `pip` should also work fine:
 ```bash
-pip install tslearn
+python -m pip install tslearn
 ```
 
 In this case, you should have `numpy`, `cython` and C++ build tools available
@@ -49,10 +49,10 @@ at build time.
 
 ## Using latest github-hosted version
 
-If you want to get `tslearn`'s latest version, you can refer to the repository 
+If you want to get `tslearn`'s latest version, you can refer to the repository
 hosted at github:
 ```bash
-pip install git+https://github.com/rtavenar/tslearn.git
+python -m pip install git+https://github.com/rtavenar/tslearn.git
 ```
 
 ## Troubleshooting
@@ -61,7 +61,7 @@ It seems on some platforms `Cython` dependency does not install properly.
 If you experiment such an issue, try installing it with the following command:
 
 ```bash
-pip install cython
+python -m pip install cython
 ```
 
 before you start installing `tslearn`.
@@ -69,7 +69,7 @@ If it still does not work, we suggest you switch to `conda` installation.
 
 # Documentation and API reference
 
-The documentation, including a gallery of examples, is hosted at 
+The documentation, including a gallery of examples, is hosted at
 [readthedocs](http://tslearn.readthedocs.io/en/latest/index.html).
 
 # Already available
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ at build time.
 If you want to get `tslearn`'s latest version, you can refer to the repository
 hosted at github:
 ```bash
-python -m pip install git+https://github.com/rtavenar/tslearn.git
+python -m pip install https://github.com/rtavenar/tslearn/archive/master.zip
 ```
 
 ## Troubleshooting
--- a/tslearn/docs/index.rst
+++ b/tslearn/docs/index.rst
@@ -48,7 +48,7 @@ repository hosted at github:
 
 .. code-block:: bash
 
-    python -m pip install git+https://github.com/rtavenar/tslearn.git
+    python -m pip install https://github.com/rtavenar/tslearn/archive/master.zip
 
 In this case, you should have ``numpy``, ``cython`` and C++ build tools
 available at build time.
--- a/tslearn/docs/index.rst
+++ b/tslearn/docs/index.rst
@@ -33,8 +33,8 @@ Using PyPI
 Using ``pip`` should also work fine:
 
 .. code-block:: bash
-  
-    pip install tslearn
+
+    python -m pip install tslearn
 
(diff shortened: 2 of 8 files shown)


johannfaouzi:
It feels weird to see all these checks failing while I didn't change any Python code XD

Gotta fix the supported dependencies one day...

README.md
```diff
@@ -35,24 +35,24 @@ conda install -c conda-forge tslearn
...
 at build time.
 
 ## Using latest github-hosted version
 
-If you want to get `tslearn`'s latest version, you can refer to the repository 
+If you want to get `tslearn`'s latest version, you can refer to the repository
 hosted at github:
 ```bash
-pip install git+https://github.com/rtavenar/tslearn.git
+python -m pip install git+https://github.com/rtavenar/tslearn.git
```
rth:
```suggestion
python -m pip install https://github.com/rtavenar/tslearn/archive/master.zip
```

to avoid requiring git.

tslearn/docs/index.rst
```diff
@@ -47,8 +47,8 @@ If you want to get ``tslearn``'s latest version, you can refer to the
 repository hosted at github:
 
 .. code-block:: bash
-  
-    pip install git+https://github.com/rtavenar/tslearn.git
+
+    python -m pip install git+https://github.com/rtavenar/tslearn.git
```
rth:
```suggestion
    python -m pip install https://github.com/rtavenar/tslearn/archive/master.zip
```

tslearn/docs/integration_other_software.rst
```diff
@@ -0,0 +1,230 @@
...
+    >>> from_cesium_dataset([TimeSeries(m=[1, 2]), TimeSeries(m=[1, 4, 3])])
+    array([[[ 1.],
+            [ 2.],
+            [nan]],
+
+           [[ 1.],
+            [ 4.],
+            [ 3.]]])
+    >>> len(to_cesium_dataset([[[1], [2], [None]], [[1], [4], [3]]]))
+    2
```
rth: Nice overview!

rtavenar:
@johannfaouzi could you merge master into this to see if tests still fail?

johannfaouzi:
One test fails because there is a tie in `rankdata` using the predicted probabilities which is not present using the decision functions. And it fails for only one job.

GillesVandewiele:
This is indeed a really weird one, I have never encountered it before. As you mentioned, `predict_proba` results in a tie between the probability for `X_test[0]` and `X_test[4]`, while on for `decision_function` the probability for `X_test[0]` is slightly smaller than `X_test[4]`

GillesVandewiele:
It's strange that this test (to have perfect rank correlation) is included for `non_deterministic`  estimators in sklearn (which the SVC in tslearn apparently is). Either the points should be really easy to rank, or the test should be skipped.

tslearn non_deterministic tag:  https://github.com/rtavenar/tslearn/blob/master/tslearn/svm.py#L291
sklearn: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/estimator_checks.py#L147

GillesVandewiele:
There are two solutions that can be patched in (but perhaps there are better solutions):
* Patch `check_decision_proba_consistency` function to skip the test if the estimator is non-deterministic:
```
from sklearn.utils.estimator_checks import check_decision_proba_consistency
def _check_decision_proba_consistency(name, estimator_orig):
  tags = _safe_tags(classifier)
  if 'non_deterministic' in tags and tags['non_deterministic']:
    return True
  check_decision_proba_consistency(name, estimator_orig)
```
And then in test_estimators
```
from tslearn.tests.sklearn_patches import _check_decision_proba_consistency
checks.check_decision_proba_consistency = _check_decision_proba_consistency
```
* Or patch the entire function (copy-paste) and replace the data generation by a timeseries dataset.

rtavenar:
> It's strange that this test (to have perfect rank correlation) is included for `non_deterministic` estimators in sklearn (which the SVC in tslearn apparently is). Either the points should be really easy to rank, or the test should be skipped.
> 
> tslearn non_deterministic tag: https://github.com/rtavenar/tslearn/blob/master/tslearn/svm.py#L291
> sklearn: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/estimator_checks.py#L147

Well, not so strange: it tests that both functions are indeed related, no matter how good the model is.

Then, I would prefer to patch the function with a time series dataset.

GillesVandewiele:
> Well, not so strange: it tests that both functions are indeed related, no matter how good the model is.

But if the estimator is non-deterministic, you could expect the output of it to follow some kind of distribution as opposed to the same value. If the samples in the data are then too closely related, this can cause these distributions to overlap.

rtavenar:
The point is that the relationship between decision function and predict_probas should be deterministic

GillesVandewiele:
Hmm ok, I still don't entirely agree but that's not the important point here :). The easiest solution would be to patch the `make_blobs` function of sklearn to `random_walk_blobs` instead. I cannot seem to create commits to this PR. 

This would actually make all tests performed by sklearn using `make_blobs` to use time series datasets, which makes a lot more sense. Moreover, some of the patches in `sklearn_patches` can perhaps be removed by that. I didn't think of patching the dataset generator functions of sklearn when writing it...

rtavenar:
And now I re-ran the tests and they pass... I suggest we do not try to fix this issue in this PR since it's perfectly unrelated to the content of the PR.

@johannfaouzi @GillesVandewiele can I merge then?

GillesVandewiele:
Great, we can fix it in a future PR. lgtm!

rth:
Can someone open a follow-up issue about check_decision_proba_consistency failure?

GillesVandewiele:
> Can someone open a follow-up issue about check_decision_proba_consistency failure?

Great idea, on it!

PR #204, opened 2020-03-27
Update a couple of examples

johannfaouzi:
This PR updates two examples: `LB_Keogh` and `DTW`.

Side note: there are two example with the same title for Learning Shapelets.

It could be nice to update the docstrings of most examples so that they are more detailed, if some people have some time to work on this.

### DIFF
--- a/tslearn/docs/examples/plot_barycenter_interpolate.py
+++ b/tslearn/docs/examples/plot_barycenter_interpolate.py
@@ -5,9 +5,21 @@ Soft-DTW weighted barycenters
 
 This example presents the weighted Soft-DTW time series barycenter method.
 
-:math:`X_0, X_1, X_2` and :math:`X_3` are time series from 4 different classes
-in the Trace dataset.
-Other time series are weighted barycenters.
+Soft-DTW [1] is a differentiable loss function for Dynamic Time Warping,
+allowing for the use of gradient-based algorithms. The barycenter corresponds
+to the time series that minimizes the sum of the distances between that time
+series and all the time series from a dataset. It is thus an optimization
+problem and having a differentiable loss function makes find the solution much
+easier.
+
+In this example, we consider four time series :math:`X_0, X_1, X_2` and
+:math:`X_3` from four different classes in the Trace dataset. We compute
+the barycenters for different sets of weights and plot them. The closer to a
+time series the barycenter is, the higher the weight for this time series
+is.
+
+[1] M. Cuturi and M. Blondel, "Soft-DTW: a Differentiable Loss Function for
+Time-Series". International Conference on Machine Learning, 2017.
 """
 
 # Author: Romain Tavenard
--- a/tslearn/docs/examples/plot_dtw.py
+++ b/tslearn/docs/examples/plot_dtw.py
@@ -4,7 +4,7 @@ DTW computation
 ===============
 
 This example illustrates DTW computation between time series and plots the
-optimal alignment path.
+optimal alignment path [1].
 
 The image represents cost matrix, that is the squared Euclidean distance for
 each time point between both time series, which are represented
@@ -13,6 +13,9 @@ at the left and at the top of the cost matrix.
 The optimal path, that is the path that minimizes the total cost to go from
 the first time point to the last one, is represented in white on the image.
 
+[1] H. Sakoe and S. Chiba, "Dynamic programming algorithm optimization
+for spoken word recognition". IEEE Transactions on Acoustics, Speech, and
+Signal Processing, 26(1), 43-49 (1978).
 """
 
 # Author: Romain Tavenard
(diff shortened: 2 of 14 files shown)


tslearn/docs/examples/plot_dtw.py
```diff
@@ -5,6 +5,14 @@
...
 This example illustrates DTW computation between time series and plots the
 optimal alignment path.
+
+The image represents cost matrix, that is the squared Euclidean distance for
+each time point between both time series, which are represented
+at the left and at the top of the cost matrix.
+
+The optimal path, that is the path that minimizes the total cost to go from
+the first time point to the last one, is represented in white on the image.
+
```
rtavenar: Maybe a good thing would be to add scholar references to the examples, WDYT?

rtavenar: Similarly, maybe adding a ref to the LB_Keogh doc would help.

rtavenar:
I can work on updating the example for SVM+GAK and kernel kmeans that are related (should be rather easy I guess).

Overall, here is the list of examples to be updated:

* [x] DTW path (@johannfaouzi )
* [x] LB_Keogh (@johannfaouzi )
* [ ] SVM+GAK (@rtavenar )
* [ ] kNN search (who's in?)
* [ ] kShape (who's in?)
* [x] Model persistence seems OK to me, WDYT?
* [ ] kernel kmeans (@rtavenar )
* [ ] barycenters (who's in?)
* [ ] softDTW weighted barycenters (who's in?)
* [ ] Learning shapelets (x2, who's in?)
* [ ] sDTW multi path matching (seems ok, just do the necessary trick to remove warning in output, I can do)
* [ ] Hyper-parameter tuning (seems ok, but there is a warning, @GillesVandewiele could you update?)
* [ ] Nearest neighbors and kNN search (related, who's in?)
* [ ] PAA, SAX and 1d SAX (I can do it)

tslearn/docs/examples/plot_dtw.py
johannfaouzi: I will change that!

johannfaouzi:
I can take:
* kNN search
* Nearest neighbors and kNN search

GillesVandewiele:
I will now do the shapelet examples and the hyper-parameter tuning one!

johannfaouzi:
* Mode persistence looks great.
* I removed the `plt.tight_layout()` from the sDTW multi path example
* I updated softDTW weighted barycenters

felixdivo:
@rtavenar I'll have a look at the barycenters

tslearn/docs/examples/plot_barycenter_interpolate.py
```diff
@@ -5,9 +5,21 @@
...
+problem and having a differentiable loss function makes find the solution much
+easier.
+
+In this example, we consider four time series :math:`X_0, X_1, X_2` and
+:math:`X_3` from four different classes in the Trace dataset. We compute
+the barycenters for different sets of weights and plot them. The closer to a
+time series the barycenter is, the higher the weight for this time series
+is.
+
+[1] M. Cuturi and M. Blondel, "Soft-DTW: a Differentiable Loss Function for
```
felixdivo: Shouldn't this reference be added to the actual `softdtw_barycenter()` function (too)?

rtavenar: Good idea! Done!

tslearn/docs/examples/plot_dtw.py
```diff
@@ -55,7 +66,7 @@
 ax_s_x.axis("off")
 ax_s_x.set_xlim((0, sz - 1))
 
-ax_s_y.plot(- dataset_scaled[0], numpy.arange(sz)[::-1], "b-", linewidth=3.)
+ax_s_y.plot(- dataset_scaled[0], numpy.arange(sz), "b-", linewidth=3.)
```
rtavenar, line 69: Are you sure of that? I mean, are you sure that the time series will be plotted in the right order wrt the cost matrix?

tslearn/neighbors.py
```diff
@@ -7,7 +7,11 @@
 from sklearn import neighbors
 from sklearn.neighbors import (KNeighborsClassifier, NearestNeighbors,
                                KNeighborsRegressor)
-from sklearn.neighbors.base import KNeighborsMixin
```
rtavenar, line 10: Maybe this should be moved to the discussion in #211 ?

rtavenar:
OK, I dealt with a conflict and left some notes. Apart from that, it seems the only missing example is the kShape one, right? If so, I could work on it later so that we can merge to master soon.

rtavenar:
OK, I did a bit on the kShape example. I think we are close to being ready for merge (once my comments are answered).

tslearn/neighbors.py
johannfaouzi: Yes, I didn't think that we would create another issue just for this, but since it's already done I will revert this change.

tslearn/docs/examples/plot_dtw.py
johannfaouzi: I changed the origin for the image of the cost matrix (see line 59), it is not `'lower'` instead of the default value (`'upper'`). So it looks like this now (which will be more familiar for users, since the cost matrix is usually presented this way in the papers, i.e. going from bottom left to top right):
sphx_glr_plot_dtw_001

rtavenar: OK, did not see that.

PR #205, opened 2020-03-27
Fixed a bug that ignored the user-specified gamma value with TimeSeriesSVC and TimeSeriesSVR and gak kernel

rtavenar:
Hi there,

When improving the doc examples, I found out that gamma was not set properly by `TimeSeriesSVC` and `TimeSeriesSVR` when the "gak" kernel was used.

@GillesVandewiele @johannfaouzi can any of you check that this was indeed a bug?
If it was, I would suggest we merge that PR to master and integrate the bugfix in the next patch release.

### DIFF
--- a/tslearn/tests/test_svm.py
+++ b/tslearn/tests/test_svm.py
@@ -1,22 +1,24 @@
 import numpy as np
 
 from tslearn.metrics import cdist_gak
-from tslearn.svm import TimeSeriesSVC
+from tslearn.svm import TimeSeriesSVC, TimeSeriesSVR
 
 __author__ = 'Romain Tavenard romain.tavenard[at]univ-rennes2.fr'
 
 
-def test_svm_gak():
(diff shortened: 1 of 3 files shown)


tslearn/tests/test_svm.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+
+from tslearn.metrics import cdist_gak
+from tslearn.svm import TimeSeriesSVC
+
+__author__ = 'Romain Tavenard romain.tavenard[at]univ-rennes2.fr'
+
+
+def test_svm_gak():
```
johannfaouzi: Maybe make the name of the test more explicit (`test_gamma_value_svm`). Would it be worth to also test for other kernels and for SVR (it uses the same base class so maybe not)?

rtavenar: OK, Will do

PR #207, opened 2020-03-28
[WIP] update KNN HP + Shapelet examples

GillesVandewiele:
This updates the KNN Hyper-Parameter example and the Shapelet examples

## Changes made in Shapelet (Location)
* Changed some of the hyper-parameters of the ShapeletModel to avoid convergence issues (the discovered shapelets had a completely different range than the data)
* Plot the shapelets on the test time series where they match as opposed to just always the timeseries on index 0.
* Add some more comments

## Changes made in KNN HP

I slightly changed the data that is used within the example, in order to have no ties. I also added another print The output is now:
```
Performing hyper-parameter tuning of KNN classifier...Done!

Got the following accuracies on the test set for each fold:
|n_neighbors |  weights   |score_fold_1|score_fold_2|score_fold_3|
-----------------------------------------------------------------
|           5|     uniform|     0.64706|     0.82353|      0.6875|
|           5|    distance|     0.70588|     0.88235|      0.8125|
|          25|     uniform|     0.64706|     0.64706|       0.625|
|          25|    distance|     0.82353|     0.76471|      0.8125|

Best parameter combination:
weights=distance, n_neighbors=5

```

## Warnings fixed in KNN HP
* FutureWarning: The sklearn.neighbors.base module is  deprecated in version 0.22
--> Replaced `from sklearn.neighbors.base import KNeighborsMixin` with `from sklearn.neighbors._base import KNeighborsMixin`
--> We need to make sure this does not break on other versions? Perhaps with a try/except (but that's not pretty)

* FutureWarning: Passing attributes to check_is_fitted is deprecated and will be removed in 0.23
--> Remove second argument from check_is_fitted. It will automatically look for attributes with a `_` as prefix.

==> I assume both these changes will need to happen in other classes as well.


rtavenar:
Hummm, indeed, that's an issue, if it leads to more constraints on the scikit-learn version.

This is a general issue, at the moment, we do not explicitly state which version of scikit-learn is required, but we should probably do, since it is a critical dependency for tslearn.

GillesVandewiele:
> Hummm, indeed, that's an issue, if it leads to more constraints on the scikit-learn version.
> 
> This is a general issue, at the moment, we do not explicitly state which version of scikit-learn is required, but we should probably do, since it is a critical dependency for tslearn.

A solution would be to have a fallback in place (but it's really not pretty):
```
try:
  from sklearn.neighbors._base import KNeighborsMixin
except:
  from sklearn.neighbors.base import KNeighborsMixin
```

I don't really understand the reasoning behind making these Mixins private from sklearn's side...

GillesVandewiele:
It seems like the new `check_is_fitted` fails sometimes. At first sight, it looks like it will look for attributes ending in an underscore:
```
attrs = [v for v in vars(estimator) if v.endswith("_") and not v.startswith("__")]
```

But the attributes do not start with an `_` in neighbors.py. They end with it. (so `_x` instead of `x_`).

rtavenar:
Hum, OK, I do not know what to do at the moment with that. What about splitting this in two: 

* your edit on the gallery example could be included in #204 
* this PR could be left open to investigate the scikit-learn updates in a longer term

WDYT?

GillesVandewiele:
I think I can just rename the variables, I'll try it in a bit. I'm not sure if I can commit to someone else's PR in this repository.

rtavenar:
OK and anyway this example was already great so the need for update was not urgent.

I think we should stick to sklearn naming conventions for the variables https://scikit-learn.org/stable/developers/develop.html#parameters-and-init

GillesVandewiele:
Agreed, so that means the private variables should be `x_`, right? Or am I reading it wrong?

I will soon create a PR for the 2 shapelet examples. If I have some more time today/tomorrow, I could look into either updating the README, refactoring some of the test_estimator code (patch the sklearn data generators to generate timeseries data) or I could try fixing some of the issues of the ShapeletModels. Any preference?

GillesVandewiele:
Travis now fails for Python 2.7

@rtavenar would you agree with the hacky solution I proposed:
```
try:
  from sklearn.neighbors._base import KNeighborsMixin
except:
  from sklearn.neighbors.base import KNeighborsMixin
```

Or would you have a more elegant solution?

GillesVandewiele:
I updated the Shapelet Location example as well. There are still warnings when running them though.

`FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated;`
--> I think this has to do with the version of tensorflow < 2 and the version of numpy > 1.17

`FutureWarning: The sklearn.cluster.k_means_ module is  deprecated in version 0.22`
--> This is the same error as I had with the KNN example, the name will change to `k_means_`. We will probably have to solve all of these in a separate PR, but should determine the best solution first (see my hacky solution for example)

`Passing attributes to check_is_fitted is deprecated and will be removed in 0.23.`
--> Again similar to KNN example, but I propose to do one big sklearn PR where the deprecatedWarnings are solved.

rtavenar:
100% agree with your last point. Maybe then we should have one issue / pr to discuss how to deal with changes in sklearn and another pr for your changes in the examples no ?

Also it seems #204 already fixes the iid stuff in the knn gallery example.

GillesVandewiele:
Do you want me to revert the changes in neighbors.py and move them to an other PR or something? Or do we leave it as is? I still have to update the second shapelet example as well.

rtavenar:
Yep, maybe revert the changes in neighbors.py and open a New pr that discusses those, and once again I think we should have a global strategy regarding sklearn changes

rtavenar:
@GillesVandewiele : I just noticed your PR pointed to `dev`.

The idea would be to merge the updated gallery examples to master before `tslearn` v0.4.

So maybe the best would be to save you changes somewhere and start a new PR for doc examples using a branch that is generated from `master`, to make the merge easier, WDYT?

PR #208, opened 2020-03-28
Improve testing

felixdivo:
This is more of a tiny infrastructure thing. It changes:
- `.travis.yml` to be simpler
- `test_utils.py` now creates test files in the temporary folder
- added the test artifact `tmp-tslearn-test.txt` to the ignore list of Git (it is still created by the doctests of e.g. `save_time_series_txt()`, despite the change mentioned above)
- `setup.cfg` to silence a warning from `pytest` (see [the log](https://dev.azure.com/romaintavenard/tslearn/_build/results?buildId=46&view=logs&j=9d81d885-e67f-519e-4101-a242db001def&t=021405e3-cc99-59cf-75d6-cb1513c81a12&l=43) and the [docs of pytest](https://docs.pytest.org/en/latest/reference.html#confval-junit_family))


felixdivo:
This is WIP as I first want to see if the tests pass with the new config

felixdivo:
Seems like `NamedTemporaryFile` has a strange implementation Windows (that's why it failed on Azure). Guess I'll revert that change and add the file to the `.gitignore`.

.gitignore
```diff
@@ -102,6 +102,10 @@ ENV/
 .idea/
 .DS_Store
 .pytest_cache
+.mypy_cache
+
+# Test leftovers
+tmp-tslearn-test.txt
```
rtavenar, line 108: No longer needed since you specify a new file name in the test, right?

felixdivo: Ah, I updated the PR description accordingly. It is still required since executing the doctests still produces these files since for example `save_time_series_txt()` still uses them. For the propose of readability, I wouldn't touch that snippet of documentation.

rtavenar: ok

PR #209, opened 2020-03-28
Improve the documentation and examples on barycenters

felixdivo:
These are the changes:

#### General:
- better overview in the `tslearn.barycenter` module
- that module now links to the `plot_barycenter.py` example

#### On `plot_barycenter.py`:
- textual overview over the four methods added
- some inline comments + code is shorter
- all methods use the same number of iterations and tolerance for early stopping
- the plot is visually a little bit nicer
- get rid of the convergence warning of `dtw_barycenter_averaging_subgradient()` [still TODO]

### DIFF
--- a/tslearn/barycenters.py
+++ b/tslearn/barycenters.py
@@ -7,8 +7,9 @@ A barycenter (or *Fréchet mean*) is a time series which minimizes some
 distance metric to the time series of a given data set, much like a
 centroid minimizes the (euclidean) distances to a set of points.
 
-Only the methods :func:`dtw_barycenter_averaging` and :func:`softdtw_barycenter`
-can operate on variable-length time-series (see :ref:`here<variable-length-barycenter>`).
+Only the methods :func:`dtw_barycenter_averaging` and
+:func:`softdtw_barycenter` can operate on variable-length time-series
+(see :ref:`here<variable-length-barycenter>`).
 
 See the :ref:`barycenter examples<sphx_glr_auto_examples_plot_barycenters.py>`
 for an overview over the four available methods.
--- a/tslearn/barycenters.py
+++ b/tslearn/barycenters.py
@@ -3,16 +3,17 @@
 The :mod:`tslearn.barycenters` module gathers algorithms for time series
 barycenter computation.
 
-A barycenter (or *Fréchet mean*) is a time series which minimizes some
-distance metric to the time series of a given data set, much like a
-centroid minimizes the (euclidean) distances to a set of points.
+A barycenter (or *Fréchet mean*) is a time series :math:`b` which minimizes
+the sum of distances to the time series of a given data set :math:`x`:
+
+.. math:: \\min \\sum_i d^2( b, x_i )
 
 Only the methods :func:`dtw_barycenter_averaging` and
 :func:`softdtw_barycenter` can operate on variable-length time-series
 (see :ref:`here<variable-length-barycenter>`).
 
 See the :ref:`barycenter examples<sphx_glr_auto_examples_plot_barycenters.py>`
-for an overview over the four available methods.
+for an overview.
 """
 
 # Code for soft DTW is by Mathieu Blondel under Simplified BSD license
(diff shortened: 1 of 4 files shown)


felixdivo:
I can't seem to get rid of the apparent convergence problem:

```
tslearn/tslearn/barycenters.py:806: ConvergenceWarning:
DBA loss is increasing while it should not be. Stopping optimization.
```

felixdivo:
Just as a side not, Travis seems to have problems:
https://travis-ci.org/github/rtavenar/tslearn/jobs/668165235#L646

rtavenar:
> Just as a side note, Travis seems to have problems:
> https://travis-ci.org/github/rtavenar/tslearn/jobs/668165235#L646

Yep, there are lots of timeouts during dataset downloads these days...

tslearn/barycenters.py
```diff
@@ -1,6 +1,18 @@
+# -*- coding: utf-8 -*-
 """
 The :mod:`tslearn.barycenters` module gathers algorithms for time series
 barycenter computation.
+
+A barycenter (or *Fréchet mean*) is a time series which minimizes some
```
rtavenar: I would replace that sentence with:

```
A barycenter (or *Fréchet mean*) is a time series which minimizes
the sum of distances to the time series of a given data set:

.. math::

    b = \min \sum_i d( b, x_i )
```

tslearn/barycenters.py
```diff
@@ -1,6 +1,18 @@
...
+distance metric to the time series of a given data set, much like a
+centroid minimizes the (euclidean) distances to a set of points.
+
+Only the methods :func:`dtw_barycenter_averaging` and
+:func:`softdtw_barycenter` can operate on variable-length time-series
+(see :ref:`here<variable-length-barycenter>`).
+
+See the :ref:`barycenter examples<sphx_glr_auto_examples_plot_barycenters.py>`
+for an overview over the four available methods.
```
rtavenar: Maybe not write "the four available methods" here, since we might forget to change that if, at some point, we add another method.

tslearn/docs/examples/plot_barycenters.py
```diff
@@ -4,50 +4,84 @@
...
+
+*tslearn* provides three methods for calculating barycenters for a given
+set of time series:
+
+* *Euclidean barycenter* is simply the arithmetic mean for
+  each individual point in time, minimizing the summed euclidean distance
+  for each of them. As can be seen below, it is very different from the
+  DTW-based methods and may often be inappropriate. However, it is the
+  fastest of the methods shown.
+* *DTW Barycenter Averaging (DBA)* is an iteratively refined barycenter,
```
rtavenar: Maybe here and for the next: add references for DBA and softDTW.

tslearn/barycenters.py
felixdivo: Sure. Hesitated for a moment while writing that too. ^^

tslearn/docs/examples/plot_barycenters.py
felixdivo: Done. I also added the reference for Soft-DTW to `tslearn.barycenters.softdtw_barycenter()`. It was missing there.

tslearn/barycenters.py
```diff
@@ -1,6 +1,19 @@
+# -*- coding: utf-8 -*-
 """
 The :mod:`tslearn.barycenters` module gathers algorithms for time series
 barycenter computation.
+
+A barycenter (or *Fréchet mean*) is a time series :math:`b` which minimizes
+the sum of distances to the time series of a given data set :math:`x`:
+
+.. math:: \\min \\sum_i d^2( b, x_i )
```
rtavenar: I'd rather write d(b, x_i)^2 

;)

rtavenar:
I have a last comment and we could probably merge after that.

tslearn/barycenters.py
rtavenar: And maybe write sum of squared distances in the text also.

PR #210, opened 2020-03-29
[MRG] Updated three examples

GillesVandewiele:
# Updated three examples

## plot_knnts_sklearn.py

* Removed iid from GridSearchCV (deprecated)
* Use classes 1, 2, 3 instead of 0, 1, 2 to avoid ties
* Some extra printing

## plot_shapelets.py

* Added random_state to ShapeletModel for determinism
* Added a few more comments
* Smaller batch size to ensure classification accuracy of 100%

## plot_shapelet_locations.py

* Added random_state to ShapeletModel for determinism
* Added a lot more comments
* Plot the shapelets on the best test timeseries (minimal distances) as opposed to just the first timeseries
* Remove AdaGrad optimizer for convergence issues

## Warnings

The examples still raise warnings. These mostly are sklearn DeprecationWarnings coming from imports in tslearn modules. We should fix these separately in a different PR.


GillesVandewiele:
About the PEP-8 warning. Apparently you cannot use the following variable names in Python: `E741: ‘l’, ‘O’, or ‘I’` due to them looking like `0` or `1`. We should probably rename these.

tslearn/docs/examples/plot_shapelet_locations.py
```diff
@@ -14,43 +14,58 @@
 # License: BSD 3 clause
 
 import numpy
```
rtavenar: It would probably be a good idea to have:

* a more explicit title (to distinguish between both gallery examples related to shapelets)
* a word about what shapelet locations are

no?

GillesVandewiele: You are right. I haven't actually made any updates in the headers of these three examples, but I will update them!

rtavenar:
I have added a comment, and I agree we should follow PEP8 recommendations. Then we can probably merge, thanks for your help!

GillesVandewiele:
> I have added a comment, and I agree we should follow PEP8 recommendations. Then we can probably merge, thanks for your help!

This would require me to rename some variables in `shapelets.py` (the `l` of the `grabocka_` function most notably). Do you want me to do that in this PR?

rtavenar:
> > I have added a comment, and I agree we should follow PEP8 recommendations. Then we can probably merge, thanks for your help!
> 
> This would require me to rename some variables in `shapelets.py` (the `l` of the `grabocka_` function most notably). Do you want me to do that in this PR?

No, I meant in the new code introduced in this PR

GillesVandewiele:
Well it points to this:
```
shapelet_sizes = grabocka_params_to_shapelet_size_dict(n_ts=n_ts,
                                                       ts_sz=ts_sz,
                                                       n_classes=n_classes,
                                                       l=0.125,
                                                       r=1)
```

I only changed the value of `l`, not the variable name... Perhaps I could remove that and use the default-value?

rtavenar:
Ah OK. So lets ignore it for the moment

GillesVandewiele:
Ok! I will make a commit with better headers for these three examples in a bit. I'm currently working on a (hopefully) better README (PR incoming soon)

tslearn/docs/examples/plot_shapelet_locations.py
GillesVandewiele: Is it more clear like this?

GillesVandewiele:
I updated the header texts, if those are good to you, then I think we can merge this into the master.

tslearn/docs/examples/plot_shapelets.py
```diff
@@ -62,8 +74,7 @@
 plt.tight_layout()
 plt.show()
 
-# The loss history is accessible via the `model` attribute that is a keras
-# model
+# The loss history is accessible via the `model` that is a keras model
```
rtavenar: `model_` no?

rtavenar:
I have a last comment and then I'll merge.

tslearn/docs/examples/plot_shapelets.py
GillesVandewiele: Fixed. This was something weird with my git (I never touched this part)

Issue #194, opened 2020-03-05
Refactor common tests in tslearn/tslearn/tests/test_estimators.py 


rth:
Currently `tslearn/tests/test_estimators.py` redefine a custom `check_estimator` and then monkeypatch some of the tests in `sklearn.utils.estimator_checks` to work with time series data.

In the latest version scikit-learn introduced `parametrize_with_checks` which should allow to simplify this quite a bit https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator (e.g. and avoid re-defining a check_estimator).

[Estimator tags](https://scikit-learn.org/stable/developers/develop.html#estimator-tags)  to control how common tests from scikit-learn are run. For cases where input is not 2D array, and it could be worth considering if solving https://github.com/scikit-learn/scikit-learn/issues/14057 would allow to avoid that monkey-patching. Generally don't hesitate to comment in https://github.com/scikit-learn/scikit-learn/issues/6715 and related issues in scikit-learn if estimator_checks are not behaving as expected or are difficult to use.

cc @thomasjpfan  in case you are interested in a practical use case of scikit-learn common tests applied when the X_dtype is not a 2D array.

GillesVandewiele:
I could definitely have a look at this, as soon as I managed to finalize PR #152. Not sure whether to prioritize this (as it currently works), or the requested changes for the ShapeletModels...

PR #211, opened 2020-03-29
[WIP] Fix sklearn import deprecation warnings

GillesVandewiele:
This PR fixes the deprecation warnings that are raised when importing certain (now private) modules from sklearn.

# Private API
Many things will move to a private API in the new sklearn version. Their module name will change and have a leading underscore. e.g. `sklearn.neighbors.base` becomes `sklearn.neighbors._base`. Unfortunately, these new module names will cause a crash in environments with older sklearn versions.

The proposed fix is the following for all deprecation warninings:
```python3
try:
    from sklearn.neighbors._base import KNeighborsMixin
except ImportError:
    from sklearn.neighbors.base import KNeighborsMixin
```


GillesVandewiele:
There seems to be smth wrong with my repository, as many of these changes have already been made in a previous PR (the sax-mindist one).

GillesVandewiele:
The change in `check_is_fitted` does not seem to be backward compatible with Python 2.7 (which still expects 2 arguments there).

rtavenar:
Just one note about this PR: I am doing my best on the PRs to be merged in 0.3.1, but my agenda is very busy these days, and I am very likely to postpone my reaction to these to later times (eg. when my kids are back to school :p ). Please do not take that lack of reaction personally :)

johannfaouzi:
[check_is_fitted](https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html#sklearn-utils-validation-check-is-fitted) still allows for the `attributes` parameter, it's just that the default behavior (None) is to check for attributes with a trailing underscore. I would keep private attributes private (starting with an underscore) because we don't want the users to access them if they are meant to be private.

The `attributes` parameter is mandatory for `scikit-learn < 0.22`, and version 0.20.X was the last version of scikit-learn to support Python 2.7.

rtavenar:
Maybe #194 is also related and could be considered in the same PR?

GillesVandewiele:
> Maybe #194 is also related and could be considered in the same PR?

#194 is definitely related as well, in addition, we should patch the `sklearn.dataset.generator` functions such as `make_blobs` to make timeseries datasets.

So that's a bunch of sklearn refactors that could all be handled in this PR!

tslearn/neighbors.py
```diff
@@ -3,11 +3,17 @@
...
 from sklearn.neighbors import (KNeighborsClassifier, NearestNeighbors,
-                               KNeighborsRegressor)
-from sklearn.neighbors.base import KNeighborsMixin
+                               KNeighborsRegressor, VALID_METRICS,
+                               NearestNeighbors)
+try:
+    from sklearn.neighbors._base import KNeighborsMixin
+except ImportError:
+    warnings.warn('Detected an older version of sklearn, which will'
+                  'no longer be supported in a future version.')
```
johannfaouzi: I would remove this warning, as it does not provide a lot of useful information for users:
* We use this trick to handle changes in the API of a dependency, not tslearn itself.
* Version 0.22.X of scikit-learn is very recent, and we have no idea when it will be the minimum supported version for tslearn.

rtavenar:
@GillesVandewiele and @johannfaouzi : you seem to disagree about `check_is_fitted`'s second argument. @GillesVandewiele , where have you seen that it's being deprecated? If it's just `None` by default, then I would suggest we keep our calls as they are for the moment.

For the private API, I do not see a better hack than what you suggest @GillesVandewiele , so I would be OK with it, unless someone has a better idea.

Then, there is #194 which is harder I guess (I am involved in a discussion https://github.com/scikit-learn/scikit-learn/issues/14057 about it btw, and maybe longer term.

rtavenar:
@GillesVandewiele : do you get a warning if you keep the `check_is_fitted` calls with an `attributes` parameter?

johannfaouzi:
> @GillesVandewiele : do you get a warning if you keep the `check_is_fitted` calls with an `attributes` parameter?

No (it used to be the case on the `0.22.dev0` but they changed it shortly before the release of `0.22.0`) and it is backward compatible (i.e. if you specify the `attributes` parameter, it will only look at the given attributes).

GillesVandewiele:
Indeed, I seemed to be wrong! I will revert those changes later.

rtavenar:
@GillesVandewiele maybe, in the end, the easiest thing to do would be to start a new PR that just uses your try/except trick to import the kNN mixin from `sklearn`, no?

rtavenar:
> @GillesVandewiele maybe, in the end, the easiest thing to do would be to start a new PR that just uses your try/except trick to import the kNN mixin from `sklearn`, no?

@GillesVandewiele do you think you should do more than just fix the import for v0.4?

rtavenar:
By the way, here is the list of deprecation warnings from sklearn that  the last Azure Pipelines run gets:

```
##[warning]The sklearn.cluster.k_means_ module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.cluster. Anything that cannot be imported from sklearn.cluster is now part of the private API.
##[warning]The sklearn.utils.testing module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.utils. Anything that cannot be imported from sklearn.utils is now part of the private API.
##[warning]The sklearn.neighbors.base module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.neighbors. Anything that cannot be imported from sklearn.neighbors is now part of the private API.
##[warning]Function pairwise_estimator_convert_X is deprecated; pairwise_estimator_convert_X is deprecated in version 0.22 and will be removed in version 0.24.
##[warning]Function choose_check_classifiers_labels is deprecated; choose_check_classifiers_labels is deprecated in version 0.22 and will be removed in version 0.24.
```

GillesVandewiele:
Hi Romain, 

Tomorrow I shall fix all import deprecation warnings and then we could merge this for v.0.4. We could discuss other potential deprecation issues in another PR afterwards then (just to finish this one up quickly).

GillesVandewiele:
I tried fixing all the deprecation warnings related to imports and private function usage. It was not a lot of work, so sorry that it took so long!

rtavenar:
Great ! Could you just add as comments which is the most recent import and the version for which the old one stops working ?

GillesVandewiele:
Like this?

rtavenar:
Great !

Issue #162, opened 2019-10-21
Haversine distance for the metrics modules


Danisodu:
It would be nice to have the option to choose other distances, for instance haversine, , to be used in the metrics module.

GillesVandewiele:
This would indeed be a nice suggestion, but for the time being, you can use `https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html` in combination with an sklearn Pipeline with KNN (or some clustering algorithm).

e.g.:

```
pipe = Pipeline([
	(
		'norm', 
		TimeSeriesScalerMeanVariance()
	),
	(
		'transform', 
		SymbolicAggregateApproximation(n_segments=16, alphabet_size_avg=10)
	),
	(
		'distance', 
		FunctionTransformer(calc_distances, validate=False, pass_y=False)
	)
])

all_X = numpy.vstack((X_train, X_test))
distances = pipe.transform(all_X)

# We only need the distances to the timeseries from the training set.
# Both for the test and the training set.
X_train_dist = distances[:len(X_train), :len(X_train)]
X_test_dist = distances[len(X_train):, :len(X_train)]

knn = KNeighborsClassifier(n_neighbors=1, metric='precomputed')
knn.fit(X_train_dist, y_train)
predictions = knn.predict(X_test_dist)
print('Accuracy score on test set with KNN on SAX using MINDIST')
print(confusion_matrix(y_test, predictions))
print(accuracy_score(y_test, predictions))
```

Danisodu:
I wanted this to explore the metrics module, not clustering

GillesVandewiele:
Well what do you want to do with the metrics module? Do you want a pairwise distance matrix? If so, `distances` from my code snippet will be just that.

GillesVandewiele:
Ok, so the goal would be to allow to define the metric you are using in order to calculate the pairwise distance matrix in which DTW looks for the optimal path.

In other words, instead of using always using Euclidean distances as in the matrix below, we could allow a configurable metric.

![image](https://user-images.githubusercontent.com/5750603/78170088-b6b9ed80-7452-11ea-9259-d807be3f567e.png)

Danisodu:
[...]

PR #219, opened 2020-04-23
[MRG] New feature : Custom metrics for dtw_path (metrics.dtw_path_from_metric)

rfayat:
Following the discussion about adding the possibility to use different distance metrics for computing the dynamical time wrapping #162 "Haversine distance for the metrics modules", I added **dtw_path_from_metric** to tslearn.metrics.

It extends @rtavenar's suggestion to create a new function (rather than modifying dtw_path) taking as input a precomputed distance matrix between each pair of points of two time series to wrap by also allowing to give as input a string referring to a function compatible with scipy's [cdist](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.cdist.html) or directly by passing a callable that will be used for the computation of pairwise distances (similar to the syntax of sklearn [pairwise_distances](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)).

Similarly to dtw_path, the function returns both the path and a similarity metric which is the sum of the input distance metric for the two wrapped time series.

For instance, to compute the path in the same way as dtw_path, the squared euclidean distance can be used equivalently in three ways:
- By precomputing the distance matrix :
```python
from scipy.spatial.distance import cdist
precomputed_distance = cdist(s1, s2, metric="sqeuclidean")
path, similarity = dtw_path_from_metric(dist_matrix, metric="precomputed")
```
- By specifying a metric compatible with cdist:
```python
path, similarity = dtw_path_from_metric(s1, s2, metric="sqeuclidean")
```
- By creating a custom function computing the squared euclidean distance:
```python
import numpy as np
def sqeuclidean(x, y):
    return(np.sum((x-y)**2))
path, similarity = dtw_path_from_metric(s1, s2, metric=sqeuclidean)
```
In this case, the output similarity would be the square of the one returned by dtw_path (sum of squared euclidean distances for each point vs euclidean distance).

NDLR: Thank you for this great toolbox ! This is my first pull request so I am 100% open to feedback / suggestions.

Cheers,

Romain


tslearn/metrics.py
```diff
@@ -21,6 +22,17 @@
 GLOBAL_CONSTRAINT_CODE = {None: 0, "": 0, "itakura": 1, "sakoe_chiba": 2}
 VARIABLE_LENGTH_METRICS = ["dtw", "gak", "softdtw"]
 
+_VALID_METRICS = ['euclidean', 'braycurtis', 'canberra', 'chebyshev',
```
rtavenar, line 25: Why not import those from scikit-learn? They will probably be more reactive than we can be if new metrics are added in scipy, I guess.

rfayat: Agreed, I'll even replace scipy's cdist by scikit's pairwise_distances and let it take care directly of what metrics are allowed or not.

rtavenar:
Hi @rfayat ,

Thanks, this is now very close to being ready for merge. I will leave final comments in a minute.

tslearn/docs/dtw.rst
```diff
@@ -85,6 +85,30 @@ simplicity):
        return sqrt(C[n, m])
 
 
+Using a different ground metric
```
rtavenar, line 88: I took the freedom to change this page so that the paragraph is included earlier, which looks more natural, I think.

tslearn/docs/examples/plot_dtw_custom_metric.py
```diff
@@ -0,0 +1,152 @@
...
+
+# Definitions for the axes
+left, bottom = [.01, .51], .1
+w_ts, h_ts = .1, .2
+left_h = [l + w_ts + .01 for l in left]
+width, height = .325, .65
+bottom_h = bottom + height + .02
+
+# Create the axes and plot the distance matrices
+plt.figure(1, figsize=(12, 6))
```
rtavenar, line 101: This example is great. I have maybe one minor remark: it would be better to generate two distinct figures and pick the second one as the thumbnail in the gallery, as done there: https://github.com/rtavenar/tslearn/blob/master/tslearn/docs/examples/metrics/plot_lb_keogh.py

tslearn/metrics.py
```diff
@@ -207,6 +209,188 @@ def dtw_path(s1, s2, global_constraint=None, sakoe_chiba_radius=None,
...
+
+    Valid values for metric are the same as for scikit-learn
+    `pairwise_distances`_ function I.E. a string (E.G. "euclidean",
+    "sqeuclidean", "hamming") or a function that is used to compute the
+    pairwise distances. See `scikit`_ and `scipy`_ documentations for more
+    information about the available metrics.
+
+    Parameters
+    ----------
+    s1 : array, shape = (sz1, sz2) if metric=="precomputed", (sz1,) otherwise
```
rtavenar, line 266: I'd rather set the `(sz1, d)` case as default than the precomputed case. (also add the feature dimension d for s2)

tslearn/metrics.py
```diff
@@ -207,6 +209,188 @@ def dtw_path(s1, s2, global_constraint=None, sakoe_chiba_radius=None,
...
+    the user and return both the path and the similarity.
+
+    Similarity is computed as the cumulative cost along the wrapped time
+    series.
+
+    It is not required that both time series share the same size, but they must
+    be the same dimension. DTW was originally presented in [1]_.
+
+    Valid values for metric are the same as for scikit-learn
+    `pairwise_distances`_ function I.E. a string (E.G. "euclidean",
```
rtavenar, line 259: I would not use capital letters for i.e. nor e.g.

rtavenar:
And it would be good to fix PEP8 issues too.

tslearn/docs/dtw.rst
rfayat: Looks better indeed !

tslearn/metrics.py
rfayat: Works for me, I'll also change the default metric to "euclidean" instead of "precomputed" to be consistent with the default input.

rfayat:
Hi @rtavenar,

Thank you for your feedback, I've just added the requested changes. I still could'nt solve all PEP8 issue (the two remaining are due to urls in metrics.dtw_path_from_metric's doc making the corresponding lines longer than 79 caracters), hope this won't be an issue.

Cheers

rtavenar:
Thanks a lot @rfayat !

This is really good work.
Since your PR points to `master`, I'll probably wait a bit before merging (I'd like to merge some of the currently open PRs to `dev`, then merge `dev` to `master`, branches are a bit messy at the moment), but it will be integrated in tslearn v0.4 for sure.

Issue #215, opened 2020-04-02
TimeSeriesKMeans with different length time series


MAGomes95:
Hello,

Without any delays, I will begin explaining the bug on TimeSeriesKMeans function.

At the moment, I am handling a dataset with several time series having different lengths.
When I try to cluster this time series by means of the dtw or soft-dtw metrics, and despite n_cluster parameter, the model predicts at max two clusters, in other situations, assigns all the time series to the same cluster.

Following I give an example of both situations

![bug_tslearn](https://user-images.githubusercontent.com/30867275/78295662-657a2e80-7524-11ea-826e-82616e8a6f6e.jpeg)

Thank you,

M.

LiJinyang12:
Hi all, I ran into the same bug these days. When I tried to use TimeSeriesKMeans to cluster time series of variable length, besides the inproper clustering result, I also noticed that, if I applay fit() then predict(), a ValueError was raised: **ValueError: Dimensions (except first) must match!** . But if I applied fit_predict(), no error or warning happened. However, the clustering result did not match with n_cluster parameter. In addition, the cluster_centers_ are all of the length of the shorted serie in my dataset.

rtavenar:
@LiJinyang12 your bug about dimensions not matching is probably rather related to #220

rtavenar:
@MAGomes95 

A fix has been implemented and merged to `dev` branch. It will be available in tslearn v0.4.
Your example is tested and works OK (number of predicted clusters is equal to `n_clusters`). If you want to use it before v0.4 is out, you should first install the `dev` tslearn:

```bash
python -m pip install https://github.com/rtavenar/tslearn/archive/dev.zip
```

ParthibanGowthamanON:
When we change the metric to euclidean in above example its throwing error, Is there any fix to use euclidean for different length series?

```
import os
os.chdir('C:/Users/Prathiban.gowthaman/Documents/GitRepos/CI_Multiple/disaggregation/data/processed')
from tslearn.clustering import TimeSeriesKMeans
from tslearn.utils import to_time_series_dataset
data=[[1,2,3],[7,8,9,11],[0.1,0.2,2],[1,1,1,9],[10,20,30,1000]]
data = to_time_series_dataset(data)
TimeSeriesKMeans(n_clusters=2,metric='euclidean').fit_predict(data)
```

[...]

PR #222, opened 2020-04-26
[WIP] Better kmeans initialization

rtavenar:
This PR is a follow-up to Issue #215 .

The point is that it makes little sense to compute initial centroids using kmeans++ based on euclidean distances if the clustering is intended to be run wrt DTW or softDTW. So this PR adds a custom `_k_init` (adaptation from the one in scikit-learn) to the clustering module that draws samples wrt the target metric.

### DIFF
--- a/tslearn/clustering.py
+++ b/tslearn/clustering.py
@@ -21,7 +21,8 @@ from tslearn.barycenters import euclidean_barycenter, \
     dtw_barycenter_averaging, softdtw_barycenter
 from sklearn.utils import check_array
 from sklearn.utils.validation import check_is_fitted
-from tslearn.preprocessing import TimeSeriesScalerMeanVariance
+from tslearn.preprocessing import TimeSeriesScalerMeanVariance, \
+    TimeSeriesResampler
 from tslearn.utils import (to_time_series_dataset, to_time_series,
                            ts_size, check_dims)
 from tslearn.cycc import cdist_normalized_cc, y_shifted_sbd_vec
@@ -152,17 +153,11 @@ def _check_full_length(centroids):
     """Check that provided centroids are full-length (ie. not padded with
     nans).
 
-    If some centroids are found to be padded with nans, the last value is
-    repeated until the end.
+    If some centroids are found to be padded with nans, TimeSeriesResampler is
+    used to resample the centroids.
     """
-    centroids_ = numpy.empty(centroids.shape)
-    n, max_sz = centroids.shape[:2]
-    for i in range(n):
-        sz = ts_size(centroids[i])
-        centroids_[i, :sz] = centroids[i, :sz]
-        if sz < max_sz:
-            centroids_[i, sz:] = centroids[i, sz-1]
-    return centroids_
+    resampler = TimeSeriesResampler(sz=centroids.shape[1])
+    return resampler.fit_transform(centroids)
 
 
 def _compute_inertia(distances, assignments, squared=True):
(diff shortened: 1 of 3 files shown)


rtavenar:
In fact, this PR now covers the more general question of kmeans initialization. I changed check_full_length also so that it resamples the barycenters if needed instead of simply repeating the last value.

rtavenar:
@johannfaouzi @GillesVandewiele would any of you be available for a quick PR review?

The idea of this PR is to improve centroid initialization for kmeans. I have worked on two things:

1. kmeans++: this algo initializes centroids by drawing them in such a way that the resulting set is diverse (wrt a given distance measure), so I adapted it to use the metric at use in the kmeans instead of stiocking to Euclidean distance for initialization
2. the last step of initialization was to force initial centroids to be of full length: it was done by padding the last value again and again at the end of shorter barycenters, and I decided to switch to linear temporal resampling which seems like a more reasonable choice imho

tslearn/clustering.py
```diff
@@ -30,6 +33,99 @@
...
 
 
+def _k_init_metric(X, n_clusters, cdist_metric, random_state,
+                   n_local_trials=None):
+    """Init n_clusters seeds according to k-means++ with a custom distance
+    metric.
+
+    Parameters
+    ----------
+    X : array or sparse matrix, shape (n_samples, n_features)
```
johannfaouzi, line 43: The shape of a dataset is different in tslearn and I don't think that sparse matrices are supported

tslearn/clustering.py
```diff
@@ -30,6 +33,99 @@
...
+        # This is what Arthur/Vassilvitskii tried, but did not report
+        # specific results for other than mentioning in the conclusion
+        # that it helped.
+        n_local_trials = 2 + int(numpy.log(n_clusters))
+
+    # Pick first center randomly
+    center_id = random_state.randint(n_samples)
+    if sp.issparse(X):
+        centers[0] = X[center_id].toarray()
+    else:
```
johannfaouzi, line 88: Not needed I think

tslearn/clustering.py
```diff
@@ -30,6 +33,99 @@
...
+        # Decide which candidate is the best
+        best_candidate = numpy.argmin(candidates_pot)
+        current_pot = candidates_pot[best_candidate]
+        closest_dist_sq = distance_to_candidates[best_candidate]
+        best_candidate = candidate_ids[best_candidate]
+
+        # Permanently add best center candidate found in local tries
+        if sp.issparse(X):
+            centers[c] = X[best_candidate].toarray()
+        else:
```
johannfaouzi, line 123: Likewise

tslearn/clustering.py
```diff
@@ -30,6 +33,99 @@
...
+        Function to be called for cross-distance computations
+
+    random_state : RandomState instance
+        Generator used to initialize the centers.
+
+    n_local_trials : integer, optional
+        The number of seeding trials for each center (except the first),
+        of which the one reducing inertia the most is greedily chosen.
+        Set to None to make the number of trials depend logarithmically
+        on the number of seeds (2+log(k)); this is the default.
```
johannfaouzi, line 60: Number of clusters?

rtavenar: Nope, it's something in the algo, at each step, you pick a certain number of candidates at random (with probability proportional to their squared distance to the closest center in the current set) and you keep the one that minimizes the sum of distances to other training samples.

rtavenar:
> Also I may be wrong but I did not see a new test testing for the number of clusters. Do the examples provided in the issue work as expected now?

Good idea, I will check. Thanks for your feedback, I have integrated your comments.

rtavenar:
Thanks @johannfaouzi for your feedback! I'm going to merge now.

ssimontacchi:
Hi,

Just wondering, what is the d parameter in the kmeans init when you pass an array in the form (n_clusters, ts_size, d)?

thanks!

Issue #220, opened 2020-04-24
ValueError: Dimensions (except first) must match! with TimeSeriesKMeans


vvaghu:
**Describe the bug**
I am getting ValueError: Dimensions (except first) must match! during clustering with TimeSeriesKMeans( 'dtw' and 'softdtw') on time series dataset with dimensions (9295, 7, 1) where 716 of the time series has NaN value at the end. The exception is generated from the predict() method. 
The error disappears when I remove  the time series with NaN at the end.

The printout of the X_train dimensions before fit() and after fit is the same:
Before fit() (9295, 7, 1)
After fit() (9295, 7, 1)

Here is error output:

#-----------------------------------------------------------------------------------
--> 200     _y_pred = _km.predict(_X_train)
    201 
    202     _labels = np.unique(_y_pred)

C:\ProgramData\Anaconda2\envs\py37tf\lib\site-packages\tslearn\clustering.py in predict(self, X)
    867         X = check_array(X, allow_nd=True, force_all_finite='allow-nan')
    868         check_is_fitted(self, 'cluster_centers_')
--> 869         X = check_dims(X, self.cluster_centers_)
    870         X_ = to_time_series_dataset(X)
    871         return self._assign(X_, update_class_attributes=False)

C:\ProgramData\Anaconda2\envs\py37tf\lib\site-packages\tslearn\utils.py in check_dims(X, X_fit, extend)
     67         raise ValueError('Dimensions (except first) must match!'
     68                          ' ({} and {} are passed shapes)'.format(X_fit.shape,
---> 69                                                                  X.shape))
     70 
     71     return X

ValueError: Dimensions (except first) must match! ((8, 6, 1) and (9295, 7, 1) are passed shapes)
#------------------------------------------------------------------------------------

**Environment (please complete the following information):**
 - OS: Windows 10, conda 4.8.2, Python 3.7.6
 - tslearn version 0.3.1

GillesVandewiele:
Hi @vvaghu,

Is it possible that you fitted the model with timeseries having 6 measurements, while making predictions on data with 7 measurements?

vvaghu:
Hi,
No, I am fitting and predicting on the same time series data set. But I see now that 716 time series have NaN values at the start (i.e. the first value) and not at the end. It seems that library does not expect the first value to be NaN.

rtavenar:
Hi @vvaghu 

[...]

PR #223, opened 2020-04-27
[MRG] Better shape checks at test time

rtavenar:
This PR ensures that if an estimator can deal with variable-length inputs, then the only constraint at test time is that the number of features is equal to that provided at fit time.

This is a bugfix for #220

### DIFF
--- a/tslearn/clustering.py
+++ b/tslearn/clustering.py
@@ -4,7 +4,6 @@ algorithms.
 """
 
 from __future__ import print_function
-import scipy.sparse as sp
 from sklearn.base import BaseEstimator, ClusterMixin
 from sklearn.cluster.k_means_ import _k_init
 from sklearn.metrics.cluster import \
@@ -583,7 +582,7 @@ class GlobalAlignmentKernelKMeans(BaseEstimator, BaseModelPackage,
         """
         X = check_array(X, allow_nd=True, force_all_finite=False)
         check_is_fitted(self, '_X_fit')
-        X = check_dims(X, self._X_fit.shape)
+        X = check_dims(X, self._X_fit.shape, check_n_features_only=True)
         K = self._get_kernel(X, self._X_fit)
         n_samples = X.shape[0]
         dist = numpy.zeros((n_samples, self.n_clusters))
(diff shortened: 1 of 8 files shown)


GillesVandewiele:
Great work @rtavenar I couldn't review yesterday, but I took a quick glance and it looks good! Especially that you created a new custom test for it is really nice.

I could give a more extensive review later today if need be!

rtavenar:
That would be great @GillesVandewiele

rtavenar:
Link to Travis CI job: https://travis-ci.org/github/rtavenar/tslearn/builds/680423839

tslearn/clustering.py
```diff
@@ -981,9 +980,12 @@ def predict(self, X):
         """
         X = check_array(X, allow_nd=True, force_all_finite='allow-nan')
         check_is_fitted(self, 'cluster_centers_')
-        X = check_dims(X, self.cluster_centers_.shape)
-        X_ = to_time_series_dataset(X)
-        return self._assign(X_, update_class_attributes=False)
+        if self.metric == "euclidean":
```
GillesVandewiele, line 983: An idea would be to support variable lengths for euclidean distances as well, but this could be a future PR. It could be calculated similarly as shapelets by e.g. shifting the shorter timeseries over the longer one and taking the minimal distance.

tslearn/shapelets.py
```diff
@@ -560,9 +560,10 @@ def locate(self, X):
                [0],
                [0]])
         """
-        X = check_dims(X, self._X_fit_dims)
+        check_is_fitted(self, '_X_fit_dims')
```
GillesVandewiele, line 563: We should probably do this refactor on all places where `check_is_fitted` is used.

tslearn/svm.py
```diff
@@ -40,13 +40,18 @@ def _preprocess_sklearn(self, X, y=None, fit_time=False):
...
             X, y = check_X_y(X, y, allow_nd=True,
                              force_all_finite=force_all_finite)
-        X = check_dims(X, X_fit_dims=None)
         X = to_time_series_dataset(X)
 
         if fit_time:
             self._X_fit = X
             self.gamma_ = gamma_soft_dtw(X)
             self.classes_ = numpy.unique(y)
+        else:
```
GillesVandewiele, line 49: To reduce the LoC, you could set `check_n_features_only` to `self.kernel in VARIABLE_LENGTH_METRICS` to avoid an if/else. Of course, this reduces the readability a bit.

tslearn/clustering.py
rtavenar: Hum, that would feel awkward to support variable length for Euclidean metric, I do not think I would be in favor of that.

tslearn/shapelets.py
rtavenar: OK, I'll check.

tslearn/svm.py
rtavenar: Makes sense. Will do.

PR #227, opened 2020-05-03
[MRG] Save shapelet models to disk

rtavenar:
This PR is an attempt to provide functionalities to save and load Shapelet models to / from disk.
At the moment, only the JSON format is supported.


rtavenar:
At the moment, I do not re-build the locator and transformer models. I was planning on having a dedicated `_build_derivated_models` method that would be dedicated to that and could be called from both `fit` and `_organize_model`.

rtavenar:
I need to fix the binary classification case, will do that later.

GillesVandewiele:
Hi Romain, I could take a look at this later. Now that the ShapeletModels are serializable, it would perhaps be a good time to remove the `SerializableShapeletModel` ?

rtavenar:
Hi Gilles, yes, I would be happy to have you opinion about this PR once it's in a more advanced state.
The reason for `SerializableShapeletModel`, if I remember correctly is rather that it can be cloned (and hence used in GridSearchCV, etc.), while `ShapeletModel` cannot.

rtavenar:
OK, so at the moment one can save/load from/to JSON and pickle formats. hdf5 would probably require some extra work which I will probably not be able to do before release of v0.4. So I would suggest to have only two formats at the moment and see later if someone is motivated to code the hdf5 one.

@GillesVandewiele : I'm now open to your comments.

rtavenar:
Maybe I should also test non-conventional fit/dump patterns like:

fit / dump / load / re-fit 

to ensure that I did not miss something while reloading.

rtavenar:
I tried to remove timeouts on Azure since, for Shapelet models, the most time-consuming part seems to be the graph building one, which is repeated over and over since many calls to fit are required.

GillesVandewiele:
I'll have to take a look during the weekend @rtavenar as it looks like I'll need quite a bit of time to read through the new code thoroughly, if that's ok?

rtavenar:
By the way, part of the new code is not directly related to shapelets, there are also some fixes for other small problems, you can focus your review on ShapeletModel

rtavenar:
> I would also like to look into whether we can remove the SerializableShapeletModel as it is nearly twice the same code.

I think that in order to test that, you will have to try to `clone` a `ShapeletModel` and see if something breaks (I think it was related to the optimizer not being serializable by default, except when a string is passed).

GillesVandewiele:
Yes I remember that there was an issue with that during the sklearn testing refactor in 2019. I do remember as well that I managed to implement a fix by then, but that it required an extra dependency and that we decided not to implement it back then, perhaps I could look into another solution.

tslearn/bases.py
```diff
@@ -93,6 +96,8 @@ def _listify(model_params):
 
             if isinstance(param, np.ndarray):
                 model_params[k] = param.tolist()  # for json support
+            elif isinstance(param, list) and isinstance(param[0], np.ndarray):
+                model_params[k] = [p.tolist() for p in param]  # json support
```
GillesVandewiele, line 100: Why is this part exactly needed? You can call `tolist()` on nD arrays as well:
```python3
>>> x = np.array([[[1]],[[2]],[[3]]])
>>> json.dumps(x.tolist())
```

rtavenar: Yes but keras model weights are lists of arrays, not arrays.

tslearn/hdftools.py
```diff
@@ -5,7 +5,7 @@
 from warnings import warn
 
 
-def save_dict(d, filename, group, raise_type_fail=True):
+def save_dict(d, filename, group, raise_type_fail=True, append=False):
```
GillesVandewiele, line 8: Will this append work? Will it always deserialize or can you only append specific things? Also, a small update in the docs could be useful here (the `FileExistsError` will only be raised if append=False and the path does already exist.

tslearn/neighbors.py
```diff
@@ -177,19 +177,21 @@ class KNeighborsTimeSeries(KNeighborsTimeSeriesMixin, NearestNeighbors,
     ----------
     n_neighbors : int (default: 5)
         Number of nearest neighbors to be considered for the decision.
-    metric : {'dtw', 'softdtw', 'euclidean', 'sqeuclidean', 'cityblock',
-              'sax'} (default: 'dtw')
+
+    metric : {'dtw', 'softdtw', 'euclidean', 'sqeuclidean', 'cityblock', 'sax'} (default: 'dtw')
```
GillesVandewiele, line 181: Any reason for this? You want to keep the possible values of the parameter on the same line as the name of the parameter? It does violate PEP-8.

tslearn/neighbors.py
```diff
@@ -177,19 +177,21 @@ class KNeighborsTimeSeries(KNeighborsTimeSeriesMixin, NearestNeighbors,
...
         Metric to be used at the core of the nearest neighbor procedure.
         DTW and SAX are described in more detail in :mod:`tslearn.metrics`.
         When SAX is provided as a metric, the data is expected to be
         normalized such that each time series has zero mean and unit
         variance. Other metrics are described in `scipy.spatial.distance doc
         <https://docs.scipy.org/doc/scipy/reference/spatial.distance.html>`_.
+
     metric_params : dict or None (default: None)
         Dictionnary of metric parameters.
```
GillesVandewiele, line 190: Nitpicking, not your PR/change but I spot a small typo here (dictionary is 1 n)

tslearn/hdftools.py
rtavenar: Not sure, I implemented it while trying to add `to_hdf5` for SahepeltModel, but I finally did not succeed, so I can remove it.

tslearn/neighbors.py
rtavenar: But then the docs do not look OK. I do not know how to break a long first line in sphinx/numpydoc specification of attributes.

tslearn/shapelets.py
```diff
@@ -407,39 +410,35 @@ def fit(self, X, y):
...
         self.locator_model_ = None
-        self.categorical_y_ = False
-        self.label_binarizer_ = None
         self.d_ = d
 
-        if y.ndim == 1 or y.shape[1] == 1:
-            self.label_binarizer_ = LabelBinarizer().fit(y)
-            y_ = self.label_binarizer_.transform(y)
-            self.classes_ = self.label_binarizer_.classes_
+        self.classes_ = unique_labels(y)
```
GillesVandewiele, line 415: Perhaps we could create a private function for this to reduce the complexity of the `fit` method (`_process_labels` or something?)

rtavenar: yep, why not

tslearn/neighbors.py
GillesVandewiele: Could you try adding a backslash before the newline? 

https://stackoverflow.com/questions/14414878/how-do-i-break-a-link-in-a-rst-docstring-to-satisfy-pep8

GillesVandewiele:
Left a few comments.

Some other comments:
* Could you try removing these 2 blocks of code in sklearn_patches. We had issues with the test, but perhaps you fixed them by refactoring the label/y processing code:
```python3
    if name in ['ShapeletModel', 'SerializableShapeletModel']:
        raise SkipTest('Skipping check_classifiers_classes for shapelets'
                       ' due to convergence issues...')
```
* I cannot really remember why we had SerializableShapeletModel. Cloning and serialization now both work. Moreover, they are tested in the same way by test_estimators. If there was something that didn't work for the ShapeletModel and we managed to fix it for the  SerializableShapeletModel, then we should probably write a test on that (e.g. have that as a specific doctest for the SerializableShapeletModel or something like that).

tslearn/neighbors.py
rtavenar: Worked! Thx!

rtavenar:
OK, I have taken all your comments into account except the last one about serialization. If I remember correctly, serialization used to fail for ShapeletModel if one used an optimizer provided as a keras optimizer instance. I guess we should have a look [there](https://keras.io/api/optimizers/#getweights-method) to fix it.

rtavenar:
> * Could you try removing these 2 blocks of code in sklearn_patches. We had issues with the test, but perhaps you fixed them by refactoring the label/y processing code:
> 
> ```python
>     if name in ['ShapeletModel', 'SerializableShapeletModel']:
>         raise SkipTest('Skipping check_classifiers_classes for shapelets'
>                        ' due to convergence issues...')
> ```

Did not work.

GillesVandewiele:
Ok, I guess we still have the same convergence issues... We could leave the removal of SerializableShapeletModel for another PR!

rtavenar:
> Ok, I guess we still have the same convergence issues... We could leave the removal of SerializableShapeletModel for another PR!

Yes, but maybe only for the `SerializableShapeletModel` class, that we should remove at some point. Also, I have tried a quick fix for `get_params` / `set_params` but it didn't work, and I tried to clone when the optimizer is an keras.optimizer instance, and it still fails, yet the error message says that deepcopy is only available in eager mode, so maybe moving to TF2 (which we will have to do at some point) would help, so maybe we could link these two improvements together.

Issue #228, opened 2020-05-05
Unable to access documentation page (readthedocs)


hnishi:
**Describe the bug**
I cannot see the documentation page
https://tslearn.readthedocs.io/en/latest/index.html

the page says
"This page does not exist yet."

**To Reproduce**
click the link
https://tslearn.readthedocs.io/en/latest/index.html

**Expected behavior**
I can read the documentation

**Environment (please complete the following information):**
 - OS: Windows 10 Pro 
 - Chrome: バージョン: 81.0.4044.129（Official Build） （64 ビット）

**Additional context**

GillesVandewiele:
Hi!

Thanks for reporting this. Where did you find this link exactly? It is indeed a wrong link. [This link](https://tslearn.readthedocs.io/) should work!

rtavenar:
Yes, I decided to deactivate the 'latest' version of the docs in order to prevent misunderstandings, and now the 'stable' version is the one that should be used. As @GillesVandewiele said, it would be nice if you could tell us where you found this link so that we can update it.

johannfaouzi:
On RTD you can change the default version that is displayed (latest, stable, a specific branch). I would have changed the default version to stable but kept the latest version so that we can identify potential typos / bad rendering. Otherwise we have to make sure that there is no undesired output before a new release by looking at the documentation generated from a PR (if this feature has been enabled).

rtavenar:
Why not but if we do so, we should first make sure that no publicly available link point to it.

johannfaouzi:
I don't see the issue, if users want to use a brand new feature available on master but that will not be available for a while until the next release, they should be able to see the documentation.

It's available for [scikit-learn](https://scikit-learn.org/dev/), [numpy](https://numpy.org/devdocs/), etc.

rtavenar:
Well, these projects are somehow more responsive than we are. I'd like to avoid the pain of dealing with Issues from people referring to the wrong docs version and not understanding why such version is not available in their installed tslearn. But maybe with a clear warning banner at the top could do it (I think RTD offers that, not sure how it looks though) and I'm just too pessimistic.

hnishi:
Thank you!

now I can see the documentation on [this link](https://tslearn.readthedocs.io/en/stable/index.html).

I found the wrong link in the [README.md](https://github.com/tslearn-team/tslearn/blob/master/README.md#documentation) of master branch.

I created a pull request https://github.com/tslearn-team/tslearn/pull/229

[...]

PR #229, opened 2020-05-05
fix documentation URL from latest to stable

hnishi:
Issue 
https://github.com/tslearn-team/tslearn/issues/228


rtavenar:
Thanks for reporting **and** fixing!

PR #232, opened 2020-05-15
Better tag inheritance to match with latest scikit-learn versions

### LINKED ISSUE #231: When using sklearn v0.23, some estimator checks fail
**Describe the bug**
It seems the way estimator tags are obtained has changed in `sklearn` since version 0.23 that was just released: some tests that used to pass are now failing because of missing tags.

**To Reproduce**
See example of failing test [here](https://dev.azure.com/romaintavenard/tslearn/_build/results?buildId=203&view=logs&j=2e703997-b85c-5955-fe15-cdf93bef6757&t=b1b94f87-1d38-5554-e373-8f4cd3c9d658&l=141).

**Expected behavior**
Tests should pass as they used to, since default tag value should be set.

**Environment (please complete the following information):**
 - scikit-learn v0.23

rtavenar:
The goal is to fix #231

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ install:
 - pip install codeclimate-test-reporter codecov
 - rm -f tslearn/*.c
 - python setup.py build_ext --inplace
+- if [[ "$TRAVIS_PYTHON_VERSION" == "3.5" ]]; then export TEST_IGNORE="--ignore tslearn/tests/test_estimators.py"; fi
 
 script:
 # run keras backend init to initialize backend config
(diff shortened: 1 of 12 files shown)


rtavenar:
OK, it seems that most tests pass. The only failing tests are for Python3.5 since I refactored the checks for them to match sklearn checks in v0.23 but for this Python version, there is no sklearn v0.23 package available. I will probably just patch the pipeline configuration to skip estimator checks for this case.

rth:
Great!

> The only failing tests are for Python3.5 since I refactored the checks for them to match sklearn checks in v0.23 but for this Python version, there is no sklearn v0.23 package available.

Yeah, skipping them on Python 3.5 sounds good.

rtavenar:
Thanks a lot @rth ! Now I think we can use the estimator checks again :)

Issue #238, opened 2020-05-20
Docs: some references are ill formatted


rtavenar:
**Is your feature request related to a problem? Please describe.**
Some of the references are ill-formatted in the docs (because alignments are not set as expected by numpydoc I think), cf screenshot:
<img width="786" alt="numpydoc" src="https://user-images.githubusercontent.com/18221203/82421573-7e618380-9a81-11ea-814d-a6194f1a6190.png">

().

**Describe the solution you'd like**

EDIT: We would expect the ref numbering to be `[1]`.
This seems to happen only where references are used in class docstrings (not in method docstrings for example)

This should be fixed everywhere needed. Help much appreciated!

GillesVandewiele:
Hi @rtavenar,

I will do this quickly! :)

rtavenar:
Hum, it seems this bug was due to `numpydoc` and by upgrading to version 1.0.0.dev0 does the trick (cf. `latest` docs). @GillesVandewiele I am really sorry that I wasted your time :/ I think I can close your PR, let me know if this is not the case.

GillesVandewiele:
That's no problem at all @rtavenar! You can indeed close the PR.

PR #239, opened 2020-05-20
Fixing indentation of refs in docs

GillesVandewiele:
This PR handles #238 and fixes the indentation of the references in our docs. 

As the branch structure changed quite a bit, I did not know to which branch I had to merge my PR, tell me if I need to change this :).

### DIFF
--- a/tslearn/barycenters.py
+++ b/tslearn/barycenters.py
@@ -214,8 +214,8 @@ def dtw_barycenter_averaging_petitjean(X, barycenter_size=None,
     References
     ----------
     .. [1] F. Petitjean, A. Ketterlin & P. Gancarski. A global averaging method
-       for dynamic time warping, with applications to clustering. Pattern
-       Recognition, Elsevier, 2011, Vol. 44, Num. 3, pp. 678-693
+           for dynamic time warping, with applications to clustering. Pattern
+           Recognition, Elsevier, 2011, Vol. 44, Num. 3, pp. 678-693
     """
     X_ = to_time_series_dataset(X)
     if barycenter_size is None:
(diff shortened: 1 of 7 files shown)


rtavenar:
> As the branch structure changed quite a bit, I did not know to which branch I had to merge my PR, tell me if I need to change this :).

Now it should be simple: merges to master are the correct way to proceed :)

rtavenar:
This looks great! I'll have a look at the resulting built docs before merging though :)

GillesVandewiele:
Yes please do, I actually didn't check it locally... I will take a look as well once it is done building :)

GillesVandewiele:
There's still a few ones mall-formatted. E.g. https://tslearn--239.org.readthedocs.build/en/239/gen_modules/piecewise/tslearn.piecewise.PiecewiseAggregateApproximation.html#tslearn.piecewise.PiecewiseAggregateApproximation

Let me see what's going on there...

rtavenar:
`KShape` is still buggy.
`UCR_UEA_datasets` and `CachedDatasets` too.
`OneD_SAX`, `SAX` and `PAA` too.
`ShapeletModel` and `SerializedShapeletModel` too.

I haven't checked why.

GillesVandewiele:
Is there some documentation available on this citation format? I am currently checking https://build-me-the-docs-please.readthedocs.io/en/latest/Using_Sphinx/UsingBibTeXCitationsInSphinx.html but this looks quite different to our format...

rtavenar:
I couldn't find any good one except this: https://numpydoc.readthedocs.io/en/latest/format.html (where they do not have the indenting I thought was the solution...).

I can try out things locally and let you know.

rtavenar:
OK, so my initial guess was definitely wrong: the bug does not come from indentation. It happens when the same reference is used in several places in the same resulting doc page. I am investigating how things should be done in this case.

rtavenar:
OK so my previous guess was wrong too :)

I think it happens each time we try to have a numbered reference in a class docstring. I don't know why (couldn't find anything related on the web) and I don't know if we should rather remove all numbered references in class docstrings (as done for `GloballignmentKernelKMeans` for example) or just leave it as is with weird reference numbering.

GillesVandewiele:
Perhaps we could change from a number system to this system where the author names are used?

rtavenar:
> Perhaps we could change from a number system to this system where the author names are used?

I tried, but it is ineffective: whatever the key we use, some prefix is added.

rtavenar:
Closing since this was in fact fixed by upgrading numpydoc to 1.0

PR #242, opened 2020-05-22
Use scipy to load arff data 

rtavenar:
Loading multivariate UCR UEA datasets would be a great thing to have in tslearn, so I'm re-opening this PR that was closed automatically when we removed branch dev from the main repo.


rtavenar:
Follow-up in #248

PR #245, opened 2020-05-24
[MRG] Non-callable attributes that end in `_` are automatically registered for serialization

### LINKED ISSUE #230: Record more information for all models when dumping to disk
**Is your feature request related to a problem? Please describe.**
I just noticed while working on a personal project that for kmeans, only the cluster centers were dumped to disk when using `to_*` functions. 

In my case, I would have liked to have access to predicted labels, to avoid recomputing them afterwards. I just changed it for kmeans in a commit, but it would be nice to check if the same thing should be done for other estimators. Also, maybe if someone wants to work on a PR to fix this, it could be nice to have a place somewhere were we list all provided attributes and then have `_get_model_params` and `_is_fitted` rely on this one.

If you want to work on this, please let us know by answering to this thread.

rtavenar:
Fixes #230 

@kushalkolar could you tell me if this corresponds to what you had in mind?

### DIFF
--- a/tslearn/bases.py
+++ b/tslearn/bases.py
@@ -50,11 +50,13 @@ class BaseModelPackage:
         params = {}
         for attr in dir(self):
             # Do not save properties
-            if hasattr(type(self), attr) and isinstance(getattr(type(self), attr), property):
+            if (hasattr(type(self), attr) and
+                    isinstance(getattr(type(self), attr), property)):
                 continue
-            if not attr.startswith("__") and attr.endswith("_") and not callable(getattr(self, attr)):
+            if (not attr.startswith("__") and
+                    attr.endswith("_") and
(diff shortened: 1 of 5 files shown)


tslearn/bases.py
```diff
@@ -45,10 +45,21 @@ def _is_fitted(self):
...
+            # Do not save properties
+            if (hasattr(type(self), attr) and
+                    isinstance(getattr(type(self), attr), property)):
+                continue
+            if (not attr.startswith("__") and
+                    attr.endswith("_") and
+                    not callable(getattr(self, attr))):
+                params[attr] = getattr(self, attr)
+        return params
+
```
kushalkolar, line 61: yup, this is what I had in mind

kushalkolar:
@rtavenar looks good!

PR #246, opened 2020-05-25
[MRG] Better credentials handling for codecov reports

rtavenar:
Since at the moment PR are marked as failing because Azure wont allow access to credentials from the PR.


rtavenar:
OK, there is a bug on codecov side discussed [there](https://community.codecov.io/t/codecov-uploads-from-azure-pipelines-are-failing-with-build-numbers-do-not-match/1100/22). So I guess I should stick to that implem that seems basic.

PR #252, opened 2020-05-30
[MRG] SAX scaling operates at the dataset level

rtavenar:
@GillesVandewiele this is related to the point you made in #126


GillesVandewiele:
Looks good @rtavenar, one thing we could perhaps consider is using sklearn's MinMaxScaler? But that would require to reshape the 3D timeseries matrix to a 2D (one timeseries per dimension) and back to a 3D after a transform, so I guess this implementation is better suited.

One other thing that we should perhaps test if this integrates well with the KNeighbors algorithms (I guess it should automatically, but not entirely sure...)

rtavenar:
You're right, I do not have dedicated test at the moment. Will see what I can do when I have time.

Issue #188, opened 2020-02-25
ValueError: Input contains NaN, infinity or a value too large for dtype('float64')


skasai5296:
**Describe the bug**
ValueError when having a long time series input

**To Reproduce**
```python
from tslearn.generators import random_walk_blobs
X, y = random_walk_blobs(n_ts_per_blob=10, sz=406, d=2, n_blobs=2)
y = y.astype(np.float) + np.random.randn(20) * .1
reg = TimeSeriesSVR(kernel="gak", gamma="auto")
reg.fit(X, y)
```

**Expected behavior**
The model should not return an error

**Environment (please complete the following information):**
 - OS: macOS Mojave 10.14.6
 - tslearn version 0.3.0

**Additional context**
I'm not sure why this is happening, but the max size of temporal size is 405.

johannfaouzi:
The issue comes from [tslearn.metrics.cdist_gak](https://tslearn.readthedocs.io/en/latest/gen_modules/metrics/tslearn.metrics.cdist_gak.html#tslearn.metrics.cdist_gak): The returned array contains NaN values for some reason. This is not expected though.

```python
import numpy as np
from tslearn.generators import random_walk_blobs
from tslearn.metrics import cdist_gak

X, y = random_walk_blobs(n_ts_per_blob=10, sz=406, d=2, n_blobs=2, random_state=42)
X_new = cdist_gak(X, sigma=1000)
np.isnan(X_new).any()
```
I can try to find why this is happening.

johannfaouzi:
`cdist_gak` [returns the product of three matrices](https://github.com/rtavenar/tslearn/blob/6c716eb06159063792cd09cf5fbfd7bcc57ae8f6/tslearn/metrics.py#L1262):
```python
return (diagonal_left.dot(matrix)).dot(diagonal_right)
```
where `matrix` contains the pairwise values of [unnormalized_gak](https://github.com/rtavenar/tslearn/blob/6c716eb06159063792cd09cf5fbfd7bcc57ae8f6/tslearn/metrics.py#L1084).

[...]

PR #256, opened 2020-06-02
[WIP] Add User guide

### LINKED ISSUE #247: Improved docs
**Is your feature request related to a problem? Please describe.**
At the moment, the "Getting started" section of the docs mixes two kinds of information: most of the entries are related to basic usage of `tslearn`, which is OK, but we also have a section about DTW that is rather a tutorial about what DTW is. And we should probably have more pages like that that would present the different algorithms we have in `tslearn` with more background than what the docstrings offer.

**Describe the solution you'd like**
1. Have more tutorial pages about the methods implemented in `tslearn` (eg. _Learning Shapelets_, clustering for time series, early classification once it is merged to master, etc.
2. Have a dedicated doc section for these tutorials (not sure "tutorial" is the right term here).

EDIT: I think what I have in mind is called "User Guide" in scikit-learn.

rtavenar:
Fixes #247 

Might take some time before being ready for merge...


GillesVandewiele:
This is a great idea Romain. If you want, we can definitely divide the work (i.e. I write a tutorial on some other technique as well, although you already wrote the page for the technique I'm most familiar with ;)).

I'll quickly go through the written page and give a few comments!

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
+.. _shapelets:
+
+Shapelets
+=========
+
+Shapelets are defined in [1]_ as "subsequences that are in some sense
+maximally representative of a class".
+Unformally, if we assume a binary classification setting, a shapelet is said
```
GillesVandewiele, line 8: Informally

GillesVandewiele: I would also remove "said":

"a shapelet is said discriminant" -> "a shapelet is discriminant"

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
...
+discriminant if it is **present** in most series of one class and absent from
+series of the other class.
+To assess the level of presence, one uses shapelet matches:
+
+.. math::
+
+    d(\mathbf{x}, \mathbf{s}) =
+        \min_t \| \mathbf{x}_{t\rightarrow t+L} - \mathbf{s} \|_2
```
GillesVandewiele, line 16: is this rightarrow notation a standard? I never seen it written like that (I use `:` but I'm probably biased by only writing Python code)

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
...
+
+where :math:`L` is the length (number of timestamps) of shapelet
+:math:`\mathbf{s}`. If the above-defined distance is small enough, then
+shapelet :math:`\textbf{s}` is supposed to be present in time series
+:math:`\mathbf{x}`.
+
+**TODO: illustration here**
+
+In a classification setting, the goal is then to find the most discriminant
+shapelets given some time series data.
```
GillesVandewiele, line 26: some labeled time series data (to highlight that it is supervised)

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
...
+gradient-descent.
+
+Learning Time-series Shapelets
+------------------------------
+
+``tslearn`` provides an implementation of "Learning Time-series Shapelets",
+introduced in [2]_, that is an instance of the latter category.
+In :ref:`Learning Shapelets <class-tslearn.shapelets.ShapeletModel>`,
+shapelets are learned such
+that time series represented in their shapelet-transform space are linearly
```
GillesVandewiele, line 37: time series represented in their shapelet-transform space (i.e. their distances to each of the shapelets) are linearly

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
...
+
+    [d(\mathbf{x}, \mathbf{s}_1), \cdots, d(\mathbf{x}, \mathbf{s}_k)]
+
+In ``tslearn``, computing shapelet-transforms from a fitted model is
+
+.. code-block:: python
+
+    from tslearn.shapelets import ShapeletModel
+
+    model = ShapeletModel(n_shapelets_per_size={3: 2})
```
GillesVandewiele, line 52: I would make a note on this dictionary. You can either specify them yourself or you can use the heuristic function that we provide.

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
...
+where :math:`L` is the length (number of timestamps) of shapelet
+:math:`\mathbf{s}`. If the above-defined distance is small enough, then
+shapelet :math:`\textbf{s}` is supposed to be present in time series
+:math:`\mathbf{x}`.
+
+**TODO: illustration here**
+
+In a classification setting, the goal is then to find the most discriminant
+shapelets given some time series data.
+Shapelets can be sampled from the training set [1]_ or learned using
```
GillesVandewiele, line 27: They were not really "sampled" in the original study, right? They did a brute-force search over all possible candidates and calculated the information gain. They also did it recursively to build a decision tree.

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,82 @@
...
+    from tslearn.shapelets import ShapeletModel
+
+    model = ShapeletModel(n_shapelets_per_size={3: 2})
+    model.fit(X, y)
+    shapelet_transform = model.transform(X)
+
+
+**TODO: example of separability here in a 2d space**
+
+Once fit, the model can shamelessly exhibit its shapelets:
```
GillesVandewiele, line 59: you can easily access' the learned shapelets:

rtavenar:
> This is a great idea Romain. If you want, we can definitely divide the work (i.e. I write a tutorial on some other technique as well, although you already wrote the page for the technique I'm most familiar with ;)).
> 
> I'll quickly go through the written page and give a few comments!

You're right, I should have asked. I guess you should have push access to the branch, so maybe I can leave it to you to finalize this one and I can move to another one (_e.g._ a time series clustering page that explains why it's useful to use dedicated methods for time series clustering and what these methods do).

By the way, while writing the shapelet page, something sounded very strange to me: I used `ShapeletModel` for the class name while it's only one of many shapelet models. Maybe switching to `LearningShapelets` would be better (and keeping an alias for backward compatibility), WDYT?

GillesVandewiele:
Ok that's a good idea! I'll finish up the shapelet page!

I agree, it is a very generic name. It's not much of a problem atm, as we only have one implementation in here, but it's definitely  a change that is good for the long-term (we might have other implementations in the future)!

tslearn/docs/shapelets.rst
GillesVandewiele: I have currently left this as is. I guess it's just a matter of consistency, so if you have used this rightarrow notation everywhere else, then this is definitely the best option :)

GillesVandewiele:
I'll take the matrix profiling page for my account as well Romain (but I'll do it later). I know a bit about that technique and it's an ideal opportunity to learn some more about it!

Perhaps we could create a check-list with all the algorithms of which we want a user guide page (and the responsible person). That's always nice to track progress..

rtavenar:
Tentative TODO list:

* [ ] shapelets (@GillesVandewiele )
* [ ] clustering (@rtavenar )
* [ ] matrix profile (@GillesVandewiele )
* [ ] early classification (@rtavenar )

If anyone wants to help, feel free to add items to the list or add your name as an extra author for an item.

GillesVandewiele:
Is there any guide on how the these `..ref` actually work? I find a lot of different information on that. How would I, for example, refer to a function? This doesn't work:
```
:ref:`heuristic <class-tslearn.shapelets.grabocka_params_to_shapelet_size_dict>`. 
```

Probably due to the `class-` prefix.

rtavenar:
> Is there any guide on how the these `..ref` actually work? I find a lot of different information on that. How would I, for example, refer to a function? This doesn't work:
> 
> ```
> :ref:`heuristic <class-tslearn.shapelets.grabocka_params_to_shapelet_size_dict>`. 
> ```
> 
> Probably due to the `class-` prefix.

Yep, for this one the id is: `fun-tslearn.shapelets.grabocka_params_to_shapelet_size_dict`. Basically, the idea is written (with an underscore prepended) at the beginning of the automatically generated doc files. For functions, these files follow the following template:

```
.. _fun-{{ fullname }}:

{{ fullname | underline }}

.. currentmodule:: {{ module }}

.. autofunction:: {{ objname }}

.. include:: ../backreferences/{{fullname}}.examples

.. raw:: html

     <div style='clear:both'></div>
```

hence the id is `_fun-` followed by the function name (with module info included)

GillesVandewiele:
Ok I think that wraps up a first draft for the shapelets page, @rtavenar. Am I missing something important that definitely should be mentioned?

rtavenar:
> Ok I think that wraps up a first draft for the shapelets page, @rtavenar. Am I missing something important that definitely should be mentioned?

Great! I'll provide a review asap (not today, definitely).

rtavenar:
I have a beta version of the clustering page (not sure I will have time to write about KShape in the near future, so I'd rather leave it for future works), if you want to review it. I will review your shapelet user guide before the end of the week. And I'll write the early classification one later when the module is merged to master.

rtavenar:
@GillesVandewiele do you prefer that I edit the file directly or should I provide comments for you to edit?

tslearn/docs/shapelets.rst
rtavenar: Yep, it's as you wish. I don't think it's used anywhere else.

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,121 @@
...
+
+where :math:`L` is the length (number of timestamps) of shapelet
+:math:`\mathbf{s}` and :math:`\mathbf{x}_{t\rightarrow t+L}` is the subsequence
+extracted from time series :math:`\mathbf{x}` that starts at time index
+:math:`t` and stops at :math:`t+L`.
+If the above-defined distance is small enough, then
+shapelet :math:`\textbf{s}` is supposed to be present in time series
+:math:`\mathbf{x}`.
+
+.. figure:: _static/img/shap_distance.png
```
rtavenar, line 26: Shouldn't we generate our own image here? It would feel more homogeneous to have only code-generated images in our docs, to highlight the fact that Python+tslearn is a good match for time series data analysis, no? We could still have some statement like "inspired from [1]".

GillesVandewiele:
> @GillesVandewiele do you prefer that I edit the file directly or should I provide comments for you to edit?

Whatever you prefer! :)

tslearn/docs/shapelets.rst
GillesVandewiele: The same will then apply for my other schematic figure I guess?

rtavenar: Yep, maybe, though this other one might require a bit more work. Maybe we could turn that into a gallery example by the way. I kind of like the viz there (taken from the page accompanying the paper): 



WDYT?

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,121 @@
...
+    from tslearn.shapelets import LearningShapelets
+
+    model = LearningShapelets(n_shapelets_per_size={3: 2})
+    model.fit(X_train, y_train)
+    train_distances = model.transform(X_train)
+    test_distances = model.transform(X_test)
+    shapelets = model.shapelets_as_time_series_
+
+
+This first line creates the
```
rtavenar, line 70: I would remove this first sentence.

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,121 @@
...
+
+This first line creates the 
+:ref:`LearningShapelets <class-tslearn.shapelets.LearningShapelets>`.
+A :ref:`LearningShapelets <class-tslearn.shapelets.LearningShapelets>` has
+several
+hyper-parameters, such as the maximum number of iterations and the batch size.
+One important hyper-parameters is the ``n_shapelets_per_size``
+which is a dictionary where the keys correspond to the desired lengths of the 
+shapelets and the values to the desired number of shapelets per length. When 
+set to `None`, this dictionary will be determined by a
```
rtavenar, line 78: ``None``

tslearn/docs/shapelets.rst
```diff
@@ -0,0 +1,121 @@
...
+learned shapelets by using the ``shapelets_as_time_series_`` attribute.
+
+.. figure:: _static/img/learning_shapelets.png
+    :width: 80%
+    :align: center
+
+    A schematic overview of the "Learning Shapelets" approach.
+
+
+**It is important to note that due to the fact that a technique based on
```
rtavenar, line 92: I'd rather not use bold, even if I agree that this is an important point.

rtavenar:
I think the shapelet page is already in a good shape, we could work on these illustrations that is the most salient point IMO. Let me know if you want help on something.

tslearn/docs/clustering.rst
```diff
@@ -0,0 +1,116 @@
+.. _clustering:
+
+Time Series Clustering
+======================
+
+Dedicated methods exist for time series clustering.
```
GillesVandewiele, line 6: I would say something like:

"Clustering is the task of grouping together similar objects. As neighbouring measurements in time series are correlated, dedicated methods are required. The Figure below illustrates this, by showing how the $k$-means algorithm does not produce meaningful results, due to the fact that it is not invariant to time shifts."

tslearn/docs/clustering.rst
```diff
@@ -0,0 +1,116 @@
...
+
+This Figure is the result of a :math:`k`-means clustering that uses Euclidean
+distance as a base metric.
+One issue with this metric is that it is not invariant to time shifts, while
+the dataset at stake clearly holds such invariants.
+
+:math:`k`-means and Dynamic Time Warping
+----------------------------------------
+
+Hence, it would be profitable to use :math:`k`-means clustering with a metric
```
GillesVandewiele, line 27: To overcome the previously illustrated issue, distance metrics dedicated to time series, such as DTW, are required. As can be seen in the Figure below, the usage of such metrics produce more meaningful results.

tslearn/docs/clustering.rst
```diff
@@ -0,0 +1,116 @@
...
+.. figure:: _static/img/kmeans_dtw.svg
+    :width: 100%
+    :align: center
+
+    :math:`k`-means clustering with Dynamic Time Warping. Each subfigure represents series from a given cluster and their centroid (in red).
+
+First, clusters gather time series of similar shapes, which is due to the
+ability of Dynamic Time Warping (DTW) to deal with time shifts, as explained
+above.
+Second, cluster centers (aka centroids) are computed with respect to DTW, hence
```
GillesVandewiele, line 42: Perhaps we could call these barycenters? As that is the terminology in tslearn

tslearn/docs/clustering.rst
```diff
@@ -0,0 +1,116 @@
...
+
+Another option to deal with such time shifts is to rely on the kernel trick.
+Indeed, [1]_ introduces a positive semidefinite kernel for time series,
+inspired from DTW.
+Then, the kernel :math:`k`-means algorithm [2]_, that is equivalent to a
+:math:`k`-means
+that would operate in the Reproducing Kernel Hilbert Space associated to the
+chosen kernel, can be used:
+
+.. figure:: _static/img/kernel_kmeans.svg
```
GillesVandewiele, line 76: I find this figure to have little added value in comparison to the previous one. It seems like the only difference is that the centroids/barycenters are not displayed in this figure?

rtavenar: I included it for the following reasons, but maybe it's not necessary, I don't know:
- it makes things homogeneous (one method -> one illustration)
- it highlights the fact that centroids cannot be obtained with this method
- it allows to discuss the fact that GAK and softdtw are not invariants to time shifts, which leads to results that are qualitatively different from those obtained with dtw-kmeans.

rtavenar: Should I rephrase to emphasize on this points or remove the fig?

GillesVandewiele: Ow yes. I only noticed now that the results are actually different (and not as good) as the DTW K-means! Perhaps highlight the fact that cluster 2 and 3 seem to be mixed up by the algorithm?

GillesVandewiele: I follow your arguments! I think it is indeed better for consistency to keep the figure, but do highlight those points better indeed!

rtavenar:
OK, I think I took your remarks into account for the clustering page.

GillesVandewiele:
Ok great Romain! I will be able to create custom figures for the shapelets page & start the MP page as well over the weekend. I will take another look at the clustering page as well then. I think we should be able to merge soon!

rtavenar:
I just added a text about early classification. The corresponding method is almost ready for merge to master (just lacks a gallery example, which could be used to illustrate the user guide too).

rtavenar:
@GillesVandewiele I also moved files to a dedicated folder, so I suggest you pull before you start working on MP or any other user guide section.

tslearn/docs/examples/classification/plot_shapelet_locations.py
```diff
@@ -61,18 +65,28 @@
...
+ax[0].axvline(pos, color='k', linestyle='--', alpha=0.25)
+ax[0].set_title("The aligned extracted shapelet")
+
+# We calculate the distances from the shapelet to the timeseries ourselves.
+distances = []
+time_series = X_train[test_ts_id].ravel()
+for i in range(len(time_series) - len(shap)):
+    distances.append(numpy.linalg.norm(time_series[i:i+len(shap)] - shap))
+ax[1].plot(distances)
+ax[1].axvline(numpy.argmin(distances), color='k', linestyle='--', alpha=0.25)
```
rtavenar, line 88: Maybe you could use pos here, no ?

rtavenar:
Nice improvement for the shapelet localization example !

tslearn/docs/examples/classification/plot_shapelet_locations.py
GillesVandewiele: Yes, I did do this deliberately to "confirm" that the returned locations are really the argmins, but I can definitely change this back to `pos`! I am currently working on a "distance" example as well with the distance-transformed timeseries & decision boundary.

Am I correct that I will get the corresponding weights for the linear decision boundaries in `shp_clf.model_.get_weights()[-1]`? This gives me 4 parameters for 2 shapelets (bias term and intercept?).

GillesVandewiele: Scratch that last bit. It depends on the number of classes of course. Would you prefer a binary example with 1 decision boundary or multiclass with K decision boundaries?

rtavenar: For `pos`, I understand your point, choose the option you prefer.

In order to visualize decision boundaries, I'd rather use something like:
https://nbviewer.jupyter.org/github/rtavenar/rtavenar.github.io/blob/master/teaching/ml/notebooks/01_SVM.ipynb

Of course, there is no need for a margin here.

GillesVandewiele: I just made something like this:

Screenshot from 2020-06-07 14-56-08

Is that OK? It's building...

rtavenar: I have pushed suggested changes. It's building. I think it's a great illustration both for the user guide and for the gallery!

tslearn/docs/examples/classification/plot_shapelet_distances.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
+"""
+Learning Shapelets: decision boundaries in 2D distance space
+============================================================
+
+This example illustrates the use of the "Learning Shapelets" method in order
+to learn a collection of shapelets that linearly separates the timeseries.
+In this example, we will extract two shapelets which are then used to
+transform our input time series in a two-dimensional space. Moreover, we
```
rtavenar, line 9: may be add something like "in a two-dimensional space called the shapelet-transform space in the related literature."

tslearn/docs/examples/classification/plot_shapelet_locations.py
GillesVandewiele: I made some extra additions on top of your changes. I think it would be nice to also display the timeseries as well (to better show how those markers in the distance space correspond to those time series), but perhaps it is a bit too cluttered now with all timeseries?

rtavenar:
I have tried something to have one subplot per class. If you like it, use it, otherwise, feel free to revert it.

rtavenar:
By the way, 0.9 seems unusually large for a learning rate dont you think ? Having a more standard value might be a good idea for a gallery example.

tslearn/docs/examples/classification/plot_shapelet_distances.py
```diff
@@ -0,0 +1,112 @@
...
+                            optimizer=SGD(lr=0.9),
+                            max_iter=1000,
+                            verbose=0,
+                            scale=False,
+                            random_state=42)
+shp_clf.fit(X_train, y_train)
+
+# We will plot our distances in a 2D space
+distances = shp_clf.transform(X_train).reshape((-1, 2))
+weights, biases = shp_clf.model_.layers[-1].get_weights()
```
rtavenar, line 57: Is there a reason for not using get weights from the LearningShapelet instance directly here ?

GillesVandewiele:
> By the way, 0.9 seems unusually large for a learning rate dont you think ? Having a more standard value might be a good idea for a gallery example.

I seem to have problems with convergence if I use lower learning rates (I really need a low logloss in order to have a good plot)... I like splitting up the timeseries per class, I'm just going add a small title (class K) per subplot and change it in the user guide.

tslearn/docs/examples/classification/plot_shapelet_distances.py
GillesVandewiele: I actually have to remove that code since I do not use that no longer (I just use meshgrid and use the predict function to draw the decision boundaries)

GillesVandewiele: Ok no sorry I do use them to make my predictions, let me change that.

GillesVandewiele: Fixed! Replaced by `shp_clf.get_weights('classification')`

GillesVandewiele:
I switched to an Adam optimizer which converged with reasonable learning rates (0.01)

GillesVandewiele:
It strangely enough failed for two builds (I guess those are those sporadic convergence issues). I will try to write the matrix profile page in the next coming days!

rtavenar:
> It strangely enough failed for two builds (I guess those are those sporadic convergence issues). I will try to write the matrix profile page in the next coming days!

It's definitely not related to this PR. 
I will investigate, but it seems it is just that predict_proba(...).argmax() and predict() do not agree. I assume that this can happen in cases in which the class indexes are re-indexed (eg classes [0, 1, 2] are mapped to [0, 2, 1] and the mapping is performed in predict only).

GillesVandewiele:
It indeed seems to be related to that, if you would flip the 0 and 1 labels in the predictions the test would have passed.

```
AssertionError:  Arrays are not equal  Mismatched elements: 10 / 15 (66.7%) Max absolute difference: 1 Max relative difference: 1.  x: array([0, 0, 2, 1, 2, 1, 1, 2, 2, 0, 2, 1, 0, 0, 1])  y: array([1, 1, 2, 0, 2, 0, 0, 2, 2, 1, 2, 0, 1, 1, 0])
```

rtavenar:
> It indeed seems to be related to that, if you would flip the 0 and 1 labels in the predictions the test would have passed.
> 
> ```
> AssertionError:  Arrays are not equal  Mismatched elements: 10 / 15 (66.7%) Max absolute difference: 1 Max relative difference: 1.  x: array([0, 0, 2, 1, 2, 1, 1, 2, 2, 0, 2, 1, 0, 0, 1])  y: array([1, 1, 2, 0, 2, 0, 0, 2, 2, 1, 2, 0, 1, 1, 0])
> ```

Then, the strange thing is that we do not do anything related to classes from our side, we just call sklearn's estimators, so I quite don't understand how these can pass on their side and fail on ours. I'll check if SVC goes through this check on sklearn side.

GillesVandewiele:
My apologies for this commit mess up (I was too quick up with my arrow buttons). I created an example for the matrix profile and first draft of user-guide page. It could definitely use some feedback I think. The example output would look as follows:

Screenshot from 2020-06-08 19-51-24

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -0,0 +1,37 @@
...
+
+.. figure:: ../_static/img/distance_matrix_profile.svg
+    :width: 80%
+    :align: center
+
+    For each segment, the distances to all subsequences of the time series are calculated and the minimal distance that is not zero is then used as its corresponding matrix profile value.
+
+
+Possible Applications
+=====================
```
rtavenar, line 16: Use `---...` instead

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -0,0 +1,37 @@
+.. _matrix-profile:
+
+Matrix Profile
+==============
+
+The Matrix Profile, :math:`MP`, is a new time series that can be calculated based on an input time series :math:`T` and a subsequence length :math:`m`. :math:`MP_i` corresponds to the minimal distance from the query subsequence :math:`T_{i\rightarrow i+m}` to :math:`T` [1]_. The region around query subsequence is not included when calculating the minimal distances, as these trivial matches would result in distances of :math:`0`. The Matrix Profile is closely related to the distance calculation used to transform time series into their shapelet-transform space, as illustrated below.
```
rtavenar, line 6: In our implementation, we do not discard from "the region around query subsequence", but just the identity match (match at time `i` for subsequence extracted from time `i`). Do you have a reference that would suggest to remove other neighboring matches too?

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -0,0 +1,37 @@
...
+
+.. figure:: ../_static/img/distance_matrix_profile.svg
+    :width: 80%
+    :align: center
```
rtavenar, line 10: Maybe it would be nice that the user guide takes more time to detail the extraction scheme. 

I used [this tutorial](https://stumpy.readthedocs.io/en/latest/Tutorial_The_Matrix_Profile.html) as a guide for my implementation. It's probably too detailed, but it really helps understanding what the MP transform is.

tslearn/docs/user_guide/matrix-profile.rst
GillesVandewiele: Yeah I am probably wrong there! It can return 0 if the segment can be matched in another location than the original one. I'll change that!

tslearn/docs/user_guide/matrix-profile.rst
GillesVandewiele: What do you exactly mean with the extraction scheme? How segments are extracted and how these are again normalized?

rtavenar: Something like they have in their stumpy tutorial, but will fewer details: a step-by-step example that explains the computations involved so that someone that is not familiar with MP can quickly get the underlying idea of how it's computed.

GillesVandewiele: Ok! I tried to achieve that with the current figure already, but I can definitely extend it. I'll take a look at the tutorial.

GillesVandewiele: I just added another small paragraph to refer to StumPy (if users are only requiring a matrix profile implementation, they better go there). Perhaps it's kind of silly to re-make their tutorial as well? It is a pretty impressive one with the animations etc!

rtavenar:
I really like your illustration. Maybe we could split it in two parts (the top image to explain the sliding distances, the bottom one to introduce the matrix profile from the sliding distances). Do you mind if I try something?

GillesVandewiele:
Please go ahead! :)

tslearn/docs/user_guide/matrix-profile.rst
seanlaw: Thank you all for the positive notes! `tslearn` is great for the community!

@rtavenar Just a word of warning that [the tutorial](https://stumpy.readthedocs.io/en/latest/Tutorial_The_Matrix_Profile.html) was meant to be rather naive and that, for the sake of clarity, it purposely omits z-normalization of the subsequences (it only computes straight Euclidean distances which is wrong) and so STUMPY's matrix profile calculations actually do a little bit more and leverages algorithmic optimizations that make the computations O(n*n) complexity (rather than O(m*n*n)).

rtavenar: Yep, we do the normalization, but we don't use the trick.

tslearn/docs/user_guide/matrix-profile.rst
seanlaw: Regarding what is referred to in the literature as the "exclusion zone" for the " self-match or trivial match", the authors recommend setting this to `excl_zone = int(np.ceil(m / 4))` where `m` is the length of the subsequence window. So, say you have a distance profile, `D`, and the index of the `ith` subsequence is `idx`, then you would apply the exclusion zone as:

```
zone_start = max(0, idx - excl_zone)  # This prevents negative indexing
zone_stop = min(D.shape[-1], idx + excl_zone)   # This prevents indexing beyond the length of `D`
D[..., zone_start : zone_stop + 1] = np.inf  # Notice that we need to add `1` since stop indexing is "exclusive" and we need to ensure that the discard region is symmetrical 
```

tslearn/docs/user_guide/matrix-profile.rst
seanlaw: Great. So, depending on the length of your time series (and your implementation), you'll start noticing a "slow down" around length 1,000-5,000

seanlaw:
Here are some comments/suggestions for you to take/leave:

1. I think it's actually really important that the time series, distance profile, and matrix profile all share the same x-axis (i.e., the plots should be stacked and the values should be aligned along the x-axis and not that they should be on the same plot). This is because the location actually matters and they are complementary. Right now, some of the plots are shorter and it makes interpreting things a bit more challenging. Perhaps, you can put the "segment" as an inset plot within the time series? In case it may be useful, I was thinking about something like [this plot](https://stumpy.readthedocs.io/en/latest/Tutorial_STUMPY_Basics.html#Find-a-Motif-Using-STUMP) (you'll need to scroll down to the next nearest figure). Alternatively, here's an [interactive demo](http://stumpy-live-demo.herokuapp.com/demo) of how I think about it (note that it may take 30 seconds to load since it is hosted on a Heroku free tier and takes time to boot up).
2. Where it says "minimal distance", I would change that to either "minimum distance" or "distance to its nearest (subsequence) neighbor"? 
3. In the references, it may be beneficial to point the user directly to the [STUMPY matrix profile tutorial](https://stumpy.readthedocs.io/en/latest/Tutorial_The_Matrix_Profile.html) rather than the STUMPY Github repo
4. I mentioned this above but you probably want to make sure that you are implementing the `i +/- m/4` exclusion zone as this is the standard and is also already implemented in STUMPY

rtavenar:
I will work on 4. and let Gilles write about it in the user guide

Sent from a mobile device. Please excuse my brevity.

________________________________

GillesVandewiele:
Thanks for these great comments Sean! I will integrate these.

GillesVandewiele:
I updated the example, the plot now looks as follows (note that the distances in the distance and matrix profile do not yet correspond to each other because the trivial zone is not yet implemented):

Screenshot from 2020-06-10 09-10-29

seanlaw:
> trivial zone is not yet implemented

You mean not yet implement for "distance profile" and hence the minimum value is zero?

rtavenar:
> > trivial zone is not yet implemented
> 
> You mean not yet implement for "distance profile" and hence the minimum value is zero?

Well, in fact, a trivial exclusion zone of width 1 is implemented. PR #261 implements a larger one.

GillesVandewiele:
I wasn't clear indeed. I meant that the value I found manually, taking into account the trivial zone, in the distance profile (4.288) is not equal to the value returned by the matrix profile (2.023) as it did not yet include the trivial zone at the time I made the plot.

rtavenar:
@GillesVandewiele now that the master branch is merged into this one, I guess you can get rid of these tricks and the standard MP implementation should be OK. Thanks for pointing this out.

GillesVandewiele:
Yes :ok_hand: the plot is now correct with this merge (i.e. the manually calculated distance corresponds to the one returned by matrix profile):

Screenshot from 2020-06-11 10-09-12

Are there any other suggestions for this plot/example? I will update the user guide page as well.

seanlaw:
Looks good to me!

rtavenar:
@GillesVandewiele stumpy integration for MPs is not far from being merged to master, maybe you could update the implementation section accordingly (ie: we either provide a naive numpy-based implementation or rely on stumpy in which case stumpy should be installed and this is controlled through parameter "implementation" of the `MatrixProfile` class). And the paragraph should maybe be placed before the examples, WDYT?

Also, maybe the illustration from your user guide should be updated to match the one from the gallery?

I will finalize the early classif user guide once early classif is merged to master, and then we should be very close to v0.4! :)

rtavenar:
Todo for myself : add link to user guide in early classif module

GillesVandewiele:
Hi Romain, I'll try doing this today. Indeed, the figure in the user-guide should be replaced. It would be a nice-to-have if we could link figures to the output of an example in the gallery automatically, do you think that is possible?

rtavenar:
> Hi Romain, I'll try doing this today. Indeed, the figure in the user-guide should be replaced. It would be a nice-to-have if we could link figures to the output of an example in the gallery automatically, do you think that is possible?

Great!
I don't know how to link to gallery examples, but that would be a good idea, indeed, I'll have a look.
I've added a section about kernel methods, which should fix #259 and document #188

rtavenar:
From this: https://sphinx-gallery.github.io/stable/advanced.html#know-your-gallery-files

It seems the path to output images is expected to be quite easy to guess, so we could rely on that, I guess (in our case, the output format is svg).

GillesVandewiele:
Does that require a plt.savefig() to be present in the examples then?

rtavenar:
I don't think so.
Romain Tavenard
Maître de conférences / Assistant professor
Univ. Rennes - LETG
Tél. / Phone : +33 2 99 14 18 04
http://rtavenar.github.io/research/ <http://rtavenar.github.io/research/>

GillesVandewiele:
I linked the figures of user guide and example gallery where possible. I guess now all that remains is to have perhaps a final check on the Matrix Profile page if everything I wrote there is OK? After that, LGTM :)

tslearn/docs/user_guide/dtw.rst
```diff
@@ -51,7 +51,7 @@ The following image exhibits the DTW path (in white) for a given pair of time
 series, on top of the cross-similarity matrix that stores :math:`d(x_i, y_j)`
 values.
 
-.. figure:: ../_static/img/dtw_thumb.png
+.. figure:: ../_images/sphx_glr_plot_dtw_thumb.svg
```
rtavenar, line 54: Link fails

GillesVandewiele: Yeah, I already found it weird that this was the only one with just '../' as prefix while the others have '../../'! Let's hope the fix works :)

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -3,19 +3,28 @@
...
     :align: center
 
     For each segment, the distances to all subsequences of the time series are calculated and the minimal distance that not corresponds to the original location of the segment (where the distance is zero) is returned.
 
 
+Implementation
+---------------
+
+The Matrix Profile implementation provided in ``tslearn`` uses numpy or wraps around STUMPY [2]_. Three different versions are available: 
+* ``numpy``: a slow implementation that supports multi-dimensional data
```
rtavenar, line 19: These bullets are ill-formatted

GillesVandewiele: Nicely spotted :) I am sometimes a little bit clueless on this RST syntax... I guess a blank newline is needed to indicate a new paragraph?

rtavenar:
In fact the plan is to not support multidimensional mp for any implementation in the coming version. Id rather not discuss the multidimensional case in the user guide. Wdyt ?

GillesVandewiele:
> In fact the plan is to not support multidimensional mp for any implementation in the coming version. Id rather not discuss the multidimensional case in the user guide. Wdyt ?

I think it is important to note to the user that IF they have multi-dimensional data, they MUST use the `numpy` variant? Or did I interpret that wrong?

GillesVandewiele:
I now notice that I will have to restore the dtw_thumb picture as it is used on the homepage of the docs.

rtavenar:
> I think it is important to note to the user that IF they have multi-dimensional data, they MUST use the `numpy` variant? Or did I interpret that wrong?

The thing is: I had initially implemented a naive multivariate version, but it seems that this does not correspond to what multivariate MPs are as defined in the literature. So my suggestion is that we should not support the multivariate case with any implementation for 0.4 and we'll have it in a later version.

Regarding the user guide, I suspect that if we write that we do not have multivariate implemented, we will somehow forget to update the text later when we implement it. This is why I'd rather not discuss the multivariate case in the user guide.

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -0,0 +1,48 @@
+.. _matrix-profile:
+
+Matrix Profile
+==============
+
+The Matrix Profile, :math:`MP`, is a new time series that can be calculated based on an input time series :math:`T` and a subsequence length :math:`m`. :math:`MP_i` corresponds to the minimal distance from the query subsequence :math:`T_{i\rightarrow i+m}` to :math:`T` [1]_.  As the distance from the query subsequence to itself will be equal to zero, :math:`T_{i-\frac{m}{4}\rightarrow i+\frac{m}{4}}` is considered as an exclusion zone. In order to construct the Matrix Profile, a distance profile which is similar to the distance calculation used to transform time series into their shapelet-transform space, is calculated for each subsequence, as illustrated below:
```
rtavenar, line 6: > corresponds to the minimal distance from the query subsequence :math:`T_{i\rightarrow i+m}` to :math:`T`

Maybe change to

> corresponds to the minimal distance from the query subsequence :math:`T_{i\rightarrow i+m}` to any subsequence in :math:`T`

to make things clearer?

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -0,0 +1,48 @@
...
+
+.. figure:: ../../_images/sphx_glr_plot_distance_and_matrix_profile_001.svg
+    :width: 80%
+    :align: center
+
+    For each segment, the distances to all subsequences of the time series are calculated and the minimal distance that not corresponds to the original location of the segment (where the distance is zero) is returned.
```
rtavenar, line 12: not corresponds -> does not correspond

tslearn/docs/user_guide/matrix-profile.rst
rtavenar: Also, in the last sentence of the line, maybe you could add a link to the corresponding section in the shapelet user guide?

rtavenar:
The few comment I left are my final ones, after that I guess we could merge.

tslearn/docs/user_guide/matrix-profile.rst
```diff
@@ -0,0 +1,48 @@
+.. _matrix-profile:
+
+Matrix Profile
+==============
+
+The Matrix Profile, :math:`MP`, is a new time series that can be calculated based on an input time series :math:`T` and a subsequence length :math:`m`. :math:`MP_i` corresponds to the minimal distance from the query subsequence :math:`T_{i\rightarrow i+m}` to any subsequence in :math:`T` [1]_.  As the distance from the query subsequence to itself will be equal to zero, :math:`T_{i-\frac{m}{4}\rightarrow i+\frac{m}{4}}` is considered as an exclusion zone. In order to construct the Matrix Profile, a distance profile which is :ref:`similar to the distance calculation used to transform time series into their shapelet-transform space <shapelets.rst>`, is calculated for each subsequence, as illustrated below:
```
rtavenar, line 6:
```suggestion
The Matrix Profile, :math:`MP`, is a new time series that can be calculated based on an input time series :math:`T` and a subsequence length :math:`m`. :math:`MP_i` corresponds to the minimal distance from the query subsequence :math:`T_{i\rightarrow i+m}` to any subsequence in :math:`T` [1]_.  As the distance from the query subsequence to itself will be equal to zero, :math:`T_{i-\frac{m}{4}\rightarrow i+\frac{m}{4}}` is considered as an exclusion zone. In order to construct the Matrix Profile, a distance profile which is :ref:`similar to the distance calculation used to transform time series into their shapelet-transform space <shapelets>`, is calculated for each subsequence, as illustrated below:
```

GillesVandewiele:
After some struggle with RST syntax (I didn't notice that you suggested the correct solution from the start), I managed to integrate all your comments :)

PR #261, opened 2020-06-10
[MRG] MatrixProfile: Implement an ignore zone

rtavenar:
This is related to point 4. in the following comment: https://github.com/tslearn-team/tslearn/pull/256#issuecomment-641476389

@seanlaw: could you tell me if it's what you had in mind?


rtavenar:
Oh, and also maybe one more important stuff: at the moment, the analyses of time series in the provided set are independent. Is there a standard for that @seanlaw? Should we rather compute distances wrt subsequences of all series from the set? 

Should not be too much work, but I am not sure what is considered standard. And it should be documented, definitely.

Issue #260, opened 2020-06-09
Scalable matrix profile


rtavenar:
**Is your feature request related to a problem? Please describe.**
`tslearn` has a matrix profile module that relies on a naive implementation. Based on a discussion with @seanlaw in #126 we could maybe consider having STUMPY as an optional dependency for this matrixprofile module in order to benefit from their scalable implementations.

**Describe the solution you'd like**
That would require to improve on the existing `MatrixProfile` class by allowing to pick an implementation (using parameters passed at `__init__` time) and the `_transform(...)` method should call the correct function

One additional thing to check is how stumpy deals with:
* [x] variable-length time series
* [ ] multidimensional time series

I will probably not have time to work on it. If anyone is interested to give a hand on this, feel free to tell.

seanlaw:
@rtavenar What is it meant by "variable-length time series"?

Regarding `multi-dimensional time series", STUMPY has support for this in the `stumpy.mstump` function (and with Dask (distributed server) support via `stumpy.mstumped` - note that functions ending with "ed" refer to Dask distributed support) but, with an increase in the number of dimensions, note that the computational cost will be higher.

rtavenar:
> @rtavenar Can you clarify what is it meant by "variable-length time series"? Are you referring to streaming data?

What I mean is that in `tslearn`, we do not consider single time series but sets of time series. The question is whether STUMPY implementation would make it easy to deal with sets of time series in which all time series do not have the same number of timestamps.

seanlaw:
> What I mean is that in tslearn, we do not consider single time series but sets of time series. The question is whether STUMPY implementation would make it easy to deal with sets of time series in which all time series do not have the same number of timestamps.

I'm not sure if this helps but, technically, STUMPY does not actually take "time" into account (i.e., there is no concept of time). In the dominant case, it just takes a NumPy array (or a vector in the 1-D case) of floating numbers, `T`, along with a window size, `m` and computes the matrix profile and matrix profile indices. So, you can analyze a 1D NumPy array of any length. For a set of time series, one would either compute the matrix profile for each time series independently OR one would concatenate the set into one long time series and then separate each time series with a single `np.nan` value. This latter "trick" allows you to discover motifs across multiple time series while making sure that time series boundaries (i.e., the end of `T1` and the beginning of `T2`) are ignored or not considered as a "relevant".

[...]

PR #262, opened 2020-06-10
[MRG] Wrap stumpy matrix profile computation

tcrasset:
Starting implementation of #260.
As we're still having a discussion around the topic of which multi-dimensional approach we shall take, I restricted the usage of the stumpy library to only the one-dimensional case, and we can build upon this base when the decision is made.

I welcome any feedback!


tslearn/matrix_profile.py
```diff
@@ -71,15 +71,21 @@ class MatrixProfile(TransformerMixin,
     Parameters
     ----------
     subsequence_length : int (default: 1)
-        Length of the subseries (also called window size) to be used for 
+        Length of the subseries (also called window size) to be used for
         subseries distance computations.
-    
+
+    algorithm : str (default: None)
```
rtavenar, line 77: I wouldn't call that "algorithm" since it's rather an implementation switch than an algorithm one at the moment.

tslearn/matrix_profile.py
```diff
@@ -71,15 +71,21 @@ class MatrixProfile(TransformerMixin,
...
+        Algorithm to use to compute the matrix profile.
+        Defaults to None to use the tslearn implementation.
```
rtavenar, line 79: Maybe use "numpy" for the base, pure numpy, implementation?

tslearn/matrix_profile.py
```diff
@@ -91,15 +97,27 @@ class MatrixProfile(TransformerMixin,
...
     """
 
-    def __init__(self, subsequence_length=1, scale=True):
+    def __init__(self, subsequence_length=1, algorithm=None, scale=True):
         self.subsequence_length = subsequence_length
         self.scale = scale
+        self.check_if_algorithm_exists(algorithm)
+        self.algorithm = algorithm
+
+    def check_if_algorithm_exists(self, algorithm):
```
rtavenar, line 112: This method should not be public, so please rename it so that it starts with `_` (hence, for example, it will not appear in the documentation and will not be considered part of the public API).

tslearn/matrix_profile.py
```diff
@@ -91,15 +97,27 @@ class MatrixProfile(TransformerMixin,
...
+       Matrix Profile I: All Pairs Similarity Joins for Time Series: A
        Unifying View that Includes Motifs, Discords and Shapelets.
        ICDM 2016.
     """
 
-    def __init__(self, subsequence_length=1, scale=True):
+    def __init__(self, subsequence_length=1, algorithm=None, scale=True):
         self.subsequence_length = subsequence_length
         self.scale = scale
+        self.check_if_algorithm_exists(algorithm)
```
rtavenar, line 109: I believe this check should rather occur at transform time. Otherwise, if someone inits a MP with a correct algorithm name and then change it before `fit_transform`, we'll have an unexpected behavior.

tslearn/matrix_profile.py
```diff
@@ -130,21 +148,36 @@ def _transform(self, X, y=None):
...
+                band = (np.tri(n_segments, n_segments,
+                            band_width, dtype=np.bool) &
+                        ~np.tri(n_segments, n_segments,
+                                -(band_width + 1), dtype=np.bool))
+                dists[band] = np.inf
+                X_transformed[i_ts] = dists.min(axis=1, keepdims=True)
+
+        elif self.algorithm == "stump":
+            import stumpy
+            if d > 1:
```
rtavenar, line 171: We'll have to check that in more details, maybe in the frame of this PR, WDYT?

tslearn/tests/test_matrixprofile.py
```diff
@@ -17,4 +17,19 @@ def test_consistent_woth_stumpy():
     X_tr = mp.fit_transform(X)
     X_tr_stumpy = stumpy.stump(X_stumpy, m=10)[:, 0].astype(np.float)
 
-    np.testing.assert_allclose(X_tr.ravel(), X_tr_stumpy)
\ No newline at end of file
+    np.testing.assert_allclose(X_tr.ravel(), X_tr_stumpy)
+
+
+def test_stumpy_wrapper():
```
rtavenar, line 23: Maybe it could just be an extra check in the previous test no?

rtavenar:
Hi @tcrasset 

Thanks for working on this! I've left few early comments, and I guess the next step could be to list which `stumpy` implementations we should use.

tslearn/matrix_profile.py
seanlaw: Would something like `kind` or `method` a la [`np.sort`](https://numpy.org/doc/1.18/reference/generated/numpy.sort.html) be appropriate?

rtavenar: Yep, or "implementation"

tslearn/matrix_profile.py
```diff
@@ -130,21 +148,36 @@ def _transform(self, X, y=None):
...
+                raise NotImplementedError("Currently only the single"
```
seanlaw, line 172: This error makes it sound like STUMPY can only support 1D matrix profiles and this is not the case (mstump/mstumped can handle the multi-dimensional case). It's that the multi-dimensional matrix profile has a specific definition (based on the original author's paper) that is somewhat different from the 1D case. 

However, it certainly feels like you are after the 1D uses

tslearn/matrix_profile.py
```diff
@@ -91,15 +97,27 @@ class MatrixProfile(TransformerMixin,
...
+        self.algorithm = algorithm
+
+    def check_if_algorithm_exists(self, algorithm):
+        available_algorithms = ["stump"]
+        if (
+            algorithm is not None
+            and algorithm not in available_algorithms
+        ):
+            raise ValueError(
+                "This matrix profile algorithm is not implemented."
```
seanlaw, line 119: Rather than "not implemented", is it better to say something like `f'This "{algorithm}" matrix profile algorithm is not recognized'`?

tslearn/matrix_profile.py
```diff
@@ -130,21 +148,36 @@ def _transform(self, X, y=None):
...
+                segments_2d = segments.reshape((-1, self.subsequence_length * d))
+                dists = squareform(pdist(segments_2d, "euclidean"))
+                band = (np.tri(n_segments, n_segments,
+                            band_width, dtype=np.bool) &
+                        ~np.tri(n_segments, n_segments,
+                                -(band_width + 1), dtype=np.bool))
+                dists[band] = np.inf
+                X_transformed[i_ts] = dists.min(axis=1, keepdims=True)
+
+        elif self.algorithm == "stump":
```
seanlaw, line 169: Would it be better to flip the if/else condition so that the pure-numpy version is the final `else` default? 

```
if self.algorithm == "stump":
    ...
else:  # the self.algorithm is None case
    ...
```

This way, you'll always have a guaranteed fall back and try hard to not error out?

tslearn/matrix_profile.py
seanlaw: @rtavenar What details are you specifically looking for?

tslearn/matrix_profile.py
```diff
@@ -91,15 +97,27 @@ class MatrixProfile(TransformerMixin,
...
 
-    def __init__(self, subsequence_length=1, scale=True):
+    def __init__(self, subsequence_length=1, algorithm=None, scale=True):
         self.subsequence_length = subsequence_length
         self.scale = scale
+        self.check_if_algorithm_exists(algorithm)
+        self.algorithm = algorithm
+
+    def check_if_algorithm_exists(self, algorithm):
+        available_algorithms = ["stump"]
```
seanlaw, line 113: @tcrasset There was a question in [this comment](https://github.com/tslearn-team/tslearn/pull/126#issuecomment-641316807) regarding GPU matrix profiles. STUMPY has a similar function to `stump` called `gpu_stump` which essentially has the same function signature. I recommend adding this as one of the implementations. If no valid NVIDIA GPUs are installed (i.e., no CUDA driver) then STUMPY will check this and error out appropriately before proceeding.

Between `stump` and `gpu_stump`, I think these are probably sufficient

tslearn/matrix_profile.py
```diff
@@ -130,21 +148,36 @@ def _transform(self, X, y=None):
...
+                X_transformed[i_ts] = dists.min(axis=1, keepdims=True)
+
+        elif self.algorithm == "stump":
+            import stumpy
+            if d > 1:
+                raise NotImplementedError("Currently only the single"
+                                          "dimension matrix profile is"
+                                          "supported with stumpy.")
+            for i_ts in range(n_ts):
+                result = stumpy.stump(
```
seanlaw, line 176: @tcrasset I'm really happy to see that wrapping STUMPY doesn't seem to be too complicated or am I missing something?

tslearn/matrix_profile.py
```diff
@@ -71,15 +71,21 @@ class MatrixProfile(TransformerMixin,
...
-        Length of the subseries (also called window size) to be used for 
+        Length of the subseries (also called window size) to be used for
         subseries distance computations.
-    
+
+    algorithm : str (default: None)
+        Algorithm to use to compute the matrix profile.
+        Defaults to None to use the tslearn implementation.
+        Any other algorithm must be one of ['stump'], from the
+        stumpy python library.
```
seanlaw, line 81: Is it possible to add a link to [stumpy docs](https://stumpy.readthedocs.io/en/latest) here (or to the [STUMPY API](https://stumpy.readthedocs.io/en/latest/api.html) page) so that users can read more about whether or not they want to use/install it?

tslearn/matrix_profile.py
seanlaw: Would it make sense to add a note as to when a user might want to use STUMPY instead of the default?  The primary take home message is speed/performance and memory optimization in STUMPY. 

Essentially, we want to avoid the situation(s) where the user thinks that this transformer is too slow or too memory intensive and decides not to use it. Instead, it may be helpful to detect this (say, by checking the size of the time series or the memory consumption) and warn the user or provide clear options. Just a thought.

rtavenar: Yes, I think the doc should be clear that `stumpy`-based implementations are expected to be faster and less memory consuming and we could have a link to stumpy docs.

tslearn/matrix_profile.py
rtavenar: I haven't fully understood your explanation about MPs for multidimensional time series (because I haven't taken the time): I think we should definitely implement the standard published way to do so, which is what is done in stumpy, but then we will probably have to modify our numpy-based code.

tslearn/tests/test_matrixprofile.py
tcrasset: Computing three times the matrix profile -- the original, the pure numpy version and the stumpy wrap -- and comparing the wrap with the original one, like this ? 
```
    mp = MatrixProfile(subsequence_length=10)
    mp_stumpy = MatrixProfile(subsequence_length=10, algorithm="stump")

    X_tr = mp.fit_transform(X)
    X_tr_stumpy_wrap = mp_stumpy.fit_transform(X)
    X_tr_stumpy = stumpy.stump(X_stumpy, m=10)[:, 0].astype(np.float)

    np.testing.assert_allclose(X_tr.ravel(), X_tr_stumpy)
    np.testing.assert_allclose(X_tr, X_tr_stumpy_wrap)
```

rtavenar: Yep!

tslearn/matrix_profile.py
tcrasset: Why not pass the implementation type at transform time? Is it to stay consistent with the scikit-learn method signature?

I'll move the check at transform time and make it private!

tslearn/matrix_profile.py
tcrasset: No, the STUMPY library is very easy to use, so it just meant passing in the correctly shaped array to one of the functions. I am impressed by the ease of use of your library!

tslearn/matrix_profile.py
seanlaw: Okay, I will take your lead. Please let me know how I can assist

tslearn/matrix_profile.py
seanlaw: That's great to hear! I tried very hard to keep (nearly) everything as close to the original papers as possible and, more importantly, as stateless functions and that design choice has worked out well for us. I really appreciate the feedback!

tcrasset:
Hello!
Thanks for your feedback and kind words.
I took note of your comments and changed things accordingly.

However, regarding the "numpy" implementation argument in `__init__()`, I hesitated between putting it in place of None and thus having `elif self.implementation == "numpy"` or just letting it be None with `else` as a fail safe.

tslearn/matrix_profile.py
rtavenar: Yes, this is the reason: we want to stick to sklearn API as much as possible.

rtavenar:
> Hello!
> Thanks for your feedback and kind words.
> I took note of your comments and changed things accordingly.
> 
> However, regarding the "numpy" implementation argument in `__init__()`, I hesitated between putting it in place of None and thus having `elif self.implementation == "numpy"` or just letting it be None with `else` as a fail safe.

I think I'd prefer that we throw an error if the parameter is set to an invalid value rather than default to something else.
For example, if a user writes `implementation="STUMPY"`, then defaulting to "numpy" is probably a bad idea. So I'd suggest you raise an error in the `else` clause.

tslearn/matrix_profile.py
```diff
@@ -71,15 +71,24 @@ class MatrixProfile(TransformerMixin,
     Parameters
     ----------
     subsequence_length : int (default: 1)
-        Length of the subseries (also called window size) to be used for 
+        Length of the subseries (also called window size) to be used for
         subseries distance computations.
-    
+
+    implementation : str (default: "numpy")
+        Matrix profile implementation from the stumpy library to use for
```
rtavenar, line 78: This text is not clear that one has 3 options: numpy (which does not require stumpy to be installed), stumpy and gpu_stumpy (which can work if you have some specific configuration: this should be detailed)

tslearn/matrix_profile.py
```diff
@@ -91,15 +100,22 @@ class MatrixProfile(TransformerMixin,
...
     """
 
-    def __init__(self, subsequence_length=1, scale=True):
+    def __init__(
+        self, subsequence_length=1, implementation="numpy", scale=True
+    ):
         self.subsequence_length = subsequence_length
         self.scale = scale
+        self.implementation = implementation
+        self._available_implementations = ["numpy", "stump", "gpu_stump"]
```
rtavenar, line 118: you should not set such parameter in `__init__` (sklearn conventions). And will you use it in several places or just one? Another option would be to set it as a module-level variable

tslearn/matrix_profile.py
```diff
@@ -130,21 +146,56 @@ def _transform(self, X, y=None):
...
+                band = (np.tri(n_segments, n_segments,
+                            band_width, dtype=np.bool) &
+                        ~np.tri(n_segments, n_segments,
+                                -(band_width + 1), dtype=np.bool))
+                dists[band] = np.inf
+                X_transformed[i_ts] = dists.min(axis=1, keepdims=True)
+
+        else:
+            raise ValueError(
+                'This "{}" matrix profile algorithm is not'
```
rtavenar, line 193: You could replace algorithm by implementation I would say

tslearn/matrix_profile.py
```diff
@@ -130,21 +146,56 @@ def _transform(self, X, y=None):
...
+                                          "multi-dimensional matrix profiles "
+                                          "from the stumpy library.")
+            for i_ts in range(n_ts):
+                result = stumpy.gpu_stump(
+                    T_A=X[i_ts, :, 0].ravel(),
+                    m=self.subsequence_length)
+                X_transformed[i_ts, :, 0] = result[:, 0].astype(np.float)
+
+        elif self.implementation == "numpy":
+            scaler = TimeSeriesScalerMeanVariance()
```
rtavenar, line 175: Maybe we should check the dimension here too so that in all cases we rely on 1d data and we leave multivariate case as future works maybe (or do you plan to deal with it in this PR?)

tslearn/matrix_profile.py
tcrasset: Yes, forgot that one!

tslearn/matrix_profile.py
tcrasset: Just in one place, thought it would be easier to add things to add afterwards. Will move it to the appropriate place.

tslearn/matrix_profile.py
```diff
@@ -130,21 +146,56 @@ def _transform(self, X, y=None):
...
-            dists = squareform(pdist(segments_2d, "euclidean"))
-            band = (np.tri(n_segments, n_segments,
-                           band_width, dtype=np.bool) &
-                    ~np.tri(n_segments, n_segments,
-                            -(band_width + 1), dtype=np.bool))
-            dists[band] = np.inf
-            X_transformed[i_ts] = dists.min(axis=1, keepdims=True)
+
+        if self.implementation == "stump":
+            import stumpy
```
rtavenar, line 151: Instead of importing in the function, I'd rather have a mechanism like the following at the top of the module:

https://github.com/tslearn-team/tslearn/blob/0d4c4ce5391f00aa68769e0ad81bccd0b1c32354/tslearn/bases.py#L9-L23

tslearn/matrix_profile.py
rtavenar: By the way, why do you use "stump" and "gpu_stump" instead of "stumpy" that more clearly relates to the so-called library?

tcrasset: I just took the names of the implementations from the STUMPY library itself. Do you think it would be less confusing to use "stumpy" and "gpu_stumpy", even though STUMPY methods are `stumpy.stump()` and `stumpy.gpu_stump()` ?

tslearn/matrix_profile.py
tcrasset: No, I don't plan on implementing the multivariate case in this PR.

tslearn/matrix_profile.py
seanlaw: @rtavenar In fact, the `stumpy` package has functions called `stump` and `gpu_stump` (note the missing "y"). So, it is `stumpy.stump` and `stumpy.gpu_stump` - basically, the naming was derived from the fact that the original published algorithms were called "stamp" and "stomp" and so, since we were using `NumPy` and `Numba`, we decided to call our implementation "stump" accordingly

tslearn/matrix_profile.py
seanlaw: It probably makes sense to wait until users start asking for the multi-dimensional feature before adding it. I support your decision!

rtavenar:
This looks almost ready for merge, great!

I have two comments left:

1. I'd rather have the dimensionality check independent of the implementation (since the numpy implem does not seem to implement the multivariate MP from the literature): we will have the multivariate case in a future release but at least we won't have backward compatibility issues
2. I think we should not raise a warning regarding stumpy not being installed if the user asks for numpy implem: this was the idea of the code snippet I shared above: try to import and have a variable store information about whether it failed or not so that at transform-time, you can raise an error if the user asks for stumpy and does not have it installed

Then, one last thing will be to update the changelog to state that for MP, we either have a naive numpy implementation or the option to rely on stumpy.

seanlaw:
In case it may be useful, for our own testing in STUMPY, we use super naive pure numpy implementations to test against. These may be useful to you


https://github.com/TDAmeritrade/stumpy/blob/9eb224480fe29a4359a11a9a18286cc96f26fdc3/tests/naive.py#L146

tcrasset:
There you go @rtavenar, I put the warnings in their right place.
I let you handle the conflicts in `CHANGELOG.md` so that you'll be able to decide how you want to formulate it.
Thank you for guiding me through my first pull request! Many more to come ;)

rtavenar:
LGTM! Just waiting for the checks to pass.

Congrats @tcrasset on your first merged contribution here!

For the merge issue, there has been lots of stuff merged to master recently, so it was expected to happen at some point, and pretty easy for me to fix so no worries :)

seanlaw:
@tcrasset You've done a great job and I'm blown away that this is only your first PR! You're certainly welcome to contribute to STUMPY 😄 

@rtavenar Once this is merged, would it be okay if I announced this integration/PR on Twitter through the "@stumpy_dev" handle? You tell me what/when is appropriate.

rtavenar:
@seanlaw of course you can! FYI, this is the last PR to be merged before the release of v0.4, which should be ready today or tomorrow.

seanlaw:
> @seanlaw of course you can! FYI, this is the last PR to be merged before the release of v0.4, which should be ready today or tomorrow.

Perfect! I'll wait for the release. Is there a `tslearn` Twitter handle?

rtavenar:
Nope, it's usually one of the core contributors who advertise on twitter once everything is settled (all wheels are built, etc.).

rtavenar:
https://twitter.com/Gillesvdwiele/status/1272640103614119937?s=20 😜

Issue #270, opened 2020-06-23
Relative imports and directory-based packages


rtavenar:
At the moment, we use absolute imports in all our codebase and each `tslearn` subpackage is just a single file. I think we should fix these two shortcomings at some point.

jbbqqf:
Hi! Triaging older issues this week — I think both items on this checklist have shipped and the issue can be closed.

- **Directory-based subpackages**: `tslearn.clustering`, `tslearn.metrics`, `tslearn.barycenters`, `tslearn.preprocessing` (and others) are now multi-file packages, not single modules. For example, `tslearn/clustering/` contains `kshape.py`, `kmeans.py`, `dbscan.py`, `utils.py` ([listing](https://github.com/tslearn-team/tslearn/tree/main/tslearn/clustering)). Same shape for [`tslearn/metrics/`](https://github.com/tslearn-team/tslearn/tree/main/tslearn/metrics) and [`tslearn/barycenters/`](https://github.com/tslearn-team/tslearn/tree/main/tslearn/barycenters).
- **Relative imports**: the new package `__init__.py` files use relative imports — e.g. [`tslearn/clustering/__init__.py` L8](https://github.com/tslearn-team/tslearn/blob/main/tslearn/clustering/__init__.py#L8) (`from .kshape import KShape`) and [`tslearn/barycenters/__init__.py`](https://github.com/tslearn-team/tslearn/blob/main/tslearn/barycenters/__init__.py).

Would you mind closing this out? If you had a specific subpackage in mind that's still single-file and should be split (`tslearn.shapelets`, `tslearn.neighbors`, …), point me at it and I'll open a follow-up.

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the references above were verified manually against the current `main` branch.)

PR #273, opened 2020-06-30
[WIP] Organize the subpackages into subfolders

rtavenar:
This is related to #270 

At the moment, `metrics` is covered, I'll deal with `clustering` and `utils` afterwards.


GillesVandewiele:
Hi Romain, 

This is a great idea that structures the code a lot better in my opinion! One thing we should definitely check is whether the sklearn-estimators are still automatically found by the tests: https://github.com/tslearn-team/tslearn/blob/e78ba4b0594acf3b3a901054e941d2dcea57975c/tslearn/tests/test_estimators.py#L62

I assume this will be OK, but better to check it.

felixdivo:
Should add "Closes #320" to description.

Issue #282, opened 2020-08-12
Couldn’t  loading `DuckDuckGeese` and `Handwritting`


khirotaka:
**Describe the bug**
`DuckDuckDuckGeese` and `Handwritting` cannot be loaded by `tslearn.datasets.UCR_UEA_datasets.load_dataset` .

**To Reproduce**

DuckDuckDuckGeese
```python
from tslearn.datasets import UCR_UEA_datasets
loader = UCR_UEA_datasets()
X_train, y_train, X_test, y_test = loader.load_dataset("DuckDuckGeese")

print(type(X_train))    # None
```

```sh
ls /home/user/.tslearn/datasets/UCR_UEA/DuckDuckGeese/DuckDuckGeese
DuckDuckGeese.arff		       DuckDuckGeeseDimension395_TRAIN.arff
DuckDuckGeese.jpg		       DuckDuckGeeseDimension396_TEST.arff
DuckDuckGeese.txt		       DuckDuckGeeseDimension396_TRAIN.arff
DuckDuckGeeseDimension1000_TEST.arff   DuckDuckGeeseDimension397_TEST.arff
....
```

Handwritting
```python
loader = UCR_UEA_datasets()
X_train, y_train, X_test, y_test = loader.load_dataset("Handwritting")
print(X_train)    # None
```

```sh
ls /root/.tslearn/datasets/UCR_UEA/Handwritting/
None
```

**Expected behavior**
I was expecting to get the training and evaluation data  corresponding to the specified dataset names.

**Environment (please complete the following information):**

I used Kaggle's Python Notebook environment as of 08/12/2020.

 - OS: Ubuntu 18.04.4
 - tslearn version '0.4.1'

**Additional context**
Apparently, I was able to download the `DuckDuckGeese` dataset itself.
`Handwritting`, on the other hand, appears to be failing to download. The directory is created, but its contents are empty.

johannfaouzi:
Thanks for the report. The `Handwritting` dataset can be manually downloaded from this [link](http://www.timeseriesclassification.com/Downloads/Handwriting.zip). We'll investigate where this bug comes from.

felixdivo:
I guess that you mean the `Handwriting` dataset. Apparently, there seems to be no `.txt` (only `.ts` and `.arff`) files in the ZIP file (any more), so it's hard to fix that on the tslean side without implementing support for the other data formats. The same problem appears with `DuckDuckGeese`. Other datasets like `TwoPatterns` still work fine.

[...]

PR #297, opened 2020-10-09
Improve warnings in UCR_UEA_datasets.load_dataset()

felixdivo:
While debugging #282, I improved the warnings that are raised by `tslearn.UCR_UEA_datasets.load_dataset()`. It now more clearly explains why things fail in the case of #282.

### DIFF
--- a/tslearn/datasets.py
+++ b/tslearn/datasets.py
@@ -132,7 +132,7 @@ class UCR_UEA_datasets(object):
                                    "CinCECGtorso", "CinCECGTorso")
             in_file_string_replace(self._baseline_scores_filename,
                                    "StarlightCurves", "StarLightCurves")
-        except:
+        except Exception:
             self._baseline_scores_filename = None
 
         self._ignore_list = ["Data Descriptions"]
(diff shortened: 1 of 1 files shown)


felixdivo:
I've just seen that these changes are a bit dated an will rebase them next week.

johannfaouzi:
In my project (https://github.com/johannfaouzi/pyts), I have a function to load datasets in `.arff` format. It uses [scipy.io.arff.loadarff](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.arff.loadarff.html) and requires `scipy >= 1.3.0` to work with relational ARFF. Maybe worth adding to tslearn? 

[Relevant file](https://github.com/johannfaouzi/pyts/blob/master/pyts/datasets/uea.py), lines 269-273 and 286-314.

Basically, it loads the `.arff` files using `scipy.io.arff.loadarff` and then it transforms this object into a 3D numpy array, but the order of the dimension is different in my project. It could be easily adapted I think.

felixdivo:
@johannfaouzi Thanks for the idea. I just realized that [on the current master branch of tslearn](https://github.com/tslearn-team/tslearn/blob/master/tslearn/datasets.py#L331). So #282 should actually work by now so I need to do some further digging. I might have a look at your implementation and will still port the warning logging over as I think that would be helpful if things go wrong in the future.

PR #298, opened 2020-10-09
Add MyPy cache directory to .gitignore list

felixdivo:
It's also a standard directory.


felixdivo:
Not required after a rebase.

PR #299, opened 2020-10-11
Improve warnings and documentation in UCR_UEA_datasets 

felixdivo:
This PR improves the documentation (inline and user facing) around `UCR_UEA_datasets.load_dataset()` and prints more specific error messages. Also properly closes all file descriptors after use & actually respect the `use_cached=False` case in `UCR_UEA_datasets.load_dataset()`.

It also uses the `warnings` module for ... warnings. The rationale for that was: (1) more helpful error messages due to the included stack trace excerpt and (2) the possibility to selectively disable such warnings in user or library code. AFAIK, that's what libraries like numpy do as well, instead of raw `sys.stderr.write()`. As I've just seen, this also seems to be the only place where that was used instead of the `warnings` module so maybe it was more of a residue.

Developed as part of debugging #282 and replaces #297

### DIFF
--- a/tslearn/datasets.py
+++ b/tslearn/datasets.py
@@ -399,9 +399,9 @@ class UCR_UEA_datasets:
         for dataset_name in self.list_datasets():
             try:
                 self.load_dataset(dataset_name)
-            except Exception:
-                warnings.warn("Could not cache dataset %s properly."
-                              % dataset_name,
+            except Exception as exception:
+                warnings.warn("Could not cache dataset \"%s\" properly: %s"
+                              % (dataset_name, str(exception)),
                               category=RuntimeWarning, stacklevel=2)
 
 
--- a/tslearn/datasets.py
+++ b/tslearn/datasets.py
@@ -400,7 +400,7 @@ class UCR_UEA_datasets:
             try:
                 self.load_dataset(dataset_name)
             except Exception as exception:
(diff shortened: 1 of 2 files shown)


rtavenar:
Hi @felixdivo 

It seems that with your version, `load_dataset()` fails both in the docs and in the tests. Could you please check that and see what needs to be changed?

felixdivo:
Sure! I somehow forgot to check the CI results on this one. Will ping you once it's working.

felixdivo:
@rtavenar Finished.

tslearn/datasets.py
```diff
@@ -208,12 +212,17 @@ def list_univariate_datasets(self):
         >>> l = UCR_UEA_datasets().list_univariate_datasets()
         >>> len(l)
         85
+
+        Returns
+        -------
+        list of str:
+            A list of the names of all univariate dataset namas.
```
rtavenar, line 219: Typo

rtavenar:
Ive just noticed a typo. Could you fix it and then i'll merge ?

tslearn/datasets.py
felixdivo:
```suggestion
            A list of the names of all univariate datasets.
```

felixdivo:
Jup, done.

rtavenar:
Thanks @felixdivo !

PR #313, opened 2020-12-10
[MRG] More metrics (new metric lcss)

Danisodu:
Hey! In the context of adding other similarity measures, I implemented the Longest Common Subsequence. Additionally, the function to find its path, just like in DTW, was also implemented. I added both the tests and an example.


tslearn/metrics.py
```diff
@@ -2031,6 +2039,241 @@ def cdist_soft_dtw_normalized(dataset1, dataset2=None, gamma=1.):
...
+    s1 : array, shape = (sz1,)
+        First time series.
+
+    s2 : array, shape = (sz2,)
+        Second time series
+
+    eps : float (default: 1.)
+        Maximum matching distance threshold.
+
+    delta : int (default: inf)
```
rtavenar, line 2098: This delta parameter looks a lot like the Sakoe-Chiba band width. Maybe using the same notations in both cases would help the end-user see that

rtavenar:
Thanks @Danisodu for that!

LCSS is definitely a standard similarity measure for time series, so it's very useful to have it in tslearn.
I'm not sure how important it can be to implement path constraints for it. I have commented your delta parameter that is the equivalent of the Sakoe Chiba band one.

tslearn/metrics.py
```diff
@@ -37,6 +37,14 @@ def _local_squared_dist(x, y):
     return dist
 
 
+@njit()
+def _local_eucl_dist(x, y):
```
rtavenar, line 41: this is manhattan distance, not euclidean distance, the name should be changed.

tslearn/metrics.py
```diff
@@ -2031,6 +2039,241 @@ def cdist_soft_dtw_normalized(dataset1, dataset2=None, gamma=1.):
...
+        Longest Common Subsequence similarity score between both time series.
+
+    """
+    l1 = s1.shape[0]
+    l2 = s2.shape[0]
+    acc_cost_mat = numpy.full((l1 + 1, l2 + 1), 0)
+
+    for i in range(1, l1 + 1):
+        for j in range(1, l2 + 1):
+            if _local_eucl_dist(s1[i - 1], s2[j - 1]) <= eps\
```
rtavenar, line 2073: we should use the same default inner distance as for Dynamic Time Warping (and maybe allow one to change that upon request)

tslearn/metrics.py
Danisodu: > I'm not sure how important it can be to implement path constraints for it. 

So you do think I should remove that from LCSS?

tslearn/metrics.py
```diff
@@ -2031,6 +2031,440 @@ def cdist_soft_dtw_normalized(dataset1, dataset2=None, gamma=1.):
...
+    Returns
+    -------
+    lcss_score : float
+        Longest Common Subsquence score between both time series.
+    """
+    l1 = s1.shape[0]
+    l2 = s2.shape[0]
+    acc_cost_mat = njit_lcss_accumulated_matrix(s1, s2, eps, mask)
+
+    return round(float(acc_cost_mat[-1][-1]) / min([l1, l2]), 2)
```
rtavenar, line 2102: No need to round here: let the end user chose if (s)he wants to round.

tslearn/metrics.py
```diff
@@ -2031,6 +2031,440 @@ def cdist_soft_dtw_normalized(dataset1, dataset2=None, gamma=1.):
...
+        s1, s2,
+        sakoe_chiba_radius=sakoe_chiba_radius)
+
+    l1 = s1.shape[0]
+    l2 = s2.shape[0]
+
+    acc_cost_mat = njit_lcss_accumulated_matrix(s1, s2, eps, mask)
+    path = _return_lcss_path(s1, s2, eps, mask, acc_cost_mat, l1, l2)
+
+    return path, round(float(acc_cost_mat[-1][-1]) / min([l1, l2]), 2)
```
rtavenar, line 2293: Remove round

tslearn/metrics.py
```diff
@@ -2031,6 +2031,440 @@ def cdist_soft_dtw_normalized(dataset1, dataset2=None, gamma=1.):
...
+        sz1 = s1.shape[0]
+        sz2 = s2.shape[0]
+        mask = compute_mask(
+            s1, s2, sakoe_chiba_radius)
+        dist_mat = pairwise_distances(s1, s2, metric=metric, **kwds)
+
+    acc_cost_mat = njit_lcss_accumulated_matrix_from_dist_matrix(dist_mat, eps, mask)
+    path = _return_lcss_path_from_dist_matrix(dist_mat, eps, acc_cost_mat, mask, sz1, sz2)
+
+    return path, round(float(acc_cost_mat[-1][-1]) / min([sz1, sz2]), 2)
```
rtavenar, line 2465: No round here.

rtavenar:
This looks better and better @Danisodu 

I would suggest you had the Itakura parallelogram constraint to LCSS in the same vein you added SakoeChiba band.

Danisodu:
Will do. FYI I will also add cdist() for LCSS.

Danisodu:
The errors I'm getting are regarding tensorflow, which is outside of these PR changes. If I'm not mistaken this is referenced in #320, but it still hasn't been closed. I will wait for the merge.

felixdivo:
> If I'm not mistaken this is referenced in #320, but it still hasn't been closed.

Yes, these are exactly those errors coming from tensorflow.

rtavenar:
Hi @Danisodu 

The conflict is due to the fact that I recently merged a major refactoring to master. Let me know if you need help to deal with this conflict.

Sorry for the inconvenience.

rtavenar:
Hi @Danisodu,

Since I generated the mess, I decided that I should clean it myself. Now the PR should be almost ready for merge to master.

Last two things I would ask would be to:
- add an example in the gallery (by adding a `plot_lcss.py` file in docs/examples/metrics that illustrates the behavior of LCSS)
- add an entry in CHANGELOG.md that documents the feature you implemented

We plan to release a new tslearn version soon: do you have any idea of when you would be able to implement these last things, so that we know if we should include LCSS in the release or not)?

Thanks again for your work!

Danisodu:
Hi @rtavenar,

Thank you for your help.
I think by the end of the week I'll be able to do changes. Is that fine for you?

rtavenar:
That would be awesome!

By the way, I just fixed a bug I introduced in the merge.

Danisodu:
Hi @rtavenar,

Were there any changes made to computing a matrix's mask? With the updates made from the master branch, it seems that it's getting some errors.

rtavenar:
Hi @Danisodu 

Using a Sakoe-Chiba band with radius 0 is probably what brings trouble. You should try with a larger value (eg. 1).

Also, It's unclear to me what your definition of a LCSS path is. Shouldn't it be connected?

Danisodu:
Hi @rtavenar,

No, it shouldn't be connected. The underlying idea of LCSS is to ignore certain regions where the distance between two time-series is too big, so the final path will not include those positions.

rtavenar:
> Hi @rtavenar,
> 
> No, it shouldn't be connected. The underlying idea of LCSS is to ignore certain regions where the distance between two time-series is too big, so the final path will not include those positions.

Oh, right. Wikipedia just told me that what I had in mind was called the "longest common *substring*" problem.

docs/examples/metrics/plot_lcss.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
+"""
+LCSS computation
```
rtavenar, line 3:
```suggestion
Longest Common Subsequence
```

Since there is this possible misunderstanding with longest common substring, I would prefer that the title be explicit about which algorithm we have in store.

docs/examples/metrics/plot_lcss_custom_metric.py
```diff
@@ -0,0 +1,157 @@
...
+dataset_1 = random_walks(n_ts=n_ts, sz=sz, d=1)
+scaler = TimeSeriesScalerMeanVariance(mu=0., std=pi)  # Rescale the time series
+dataset_scaled_1 = scaler.fit_transform(dataset_1)
+
+# LCSS using a function as the metric argument
+path_1, sim_1 = metrics.lcss_path_from_metric(
+    dataset_scaled_1[0], dataset_scaled_1[1], metric=arc_length
+)
+
+# Example 2 : Hamming distance between 2 multi-dimensional boolean time series
```
rtavenar, line 78: The output of this second is not very satisfactory: the diagonal match is fully preserved. Maybe you could change the value of epsilon or completely discard this one, I don't know.

rtavenar: second *example*

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    l2 = s2.shape[0]
+    acc_cost_mat = njit_lcss_accumulated_matrix(s1, s2, eps, mask)
+
+    return float(acc_cost_mat[-1][-1]) / min([l1, l2])
+
+
+def lcss(s1, s2, eps=1., global_constraint=None, sakoe_chiba_radius=None,
+        itakura_max_slope=None):
+    r"""Compute the Longest Common Subsequence (LCSS) similarity measure
+    between (possibly multidimensional) time series and return both the
```
rtavenar, line 1444: only the similarity here

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    value from the length of the longest common subsequence, the
+    percentage of that value regarding the length of the shortest time
+    series is returned.
+
+    According to this definition, the values returned by LCSS range from
+    0 to 1, the highest value taken when two time series fully match,
+    and vice-versa. It is not required that both time series share the
+    same size, but they must be the same dimension. LCSS was originally
+    presented in [1]_ and is discussed in more details in our
+    :ref:`dedicated user-guide page <lcss>`.
```
rtavenar, line 1461: there is no such page

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    1.0
+
+    See Also
+    --------
+    lcss : Get only the similarity score for LCSS
+    lcss_path_from_metric: Compute LCSS using a user-defined distance metric
+
+    References
+    ----------
+     .. [1] M. Vlachos, D. Gunopoulos, and G. Kollios. 2002. "Discovering
```
rtavenar, line 1654: There is an extra indent (1 character) here that might be the reason why the ref fails on the docs.

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    >>> lcss([1, 2, 3], [-2., 5., 7.], eps=3)
+    0.6666666666666666
+
+    See Also
+    --------
+    lcss_path: Get both the matching path and the similarity score for LCSS
+
+    References
+    ----------
+     .. [1] M. Vlachos, D. Gunopoulos, and G. Kollios. 2002. "Discovering
```
rtavenar, line 1519: Extra indent here: cf comment below

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    To retrieve a meaningful similarity value from the length of the
+    longest common subsequence, the percentage of that value regarding
+    the length of the shortest time series is returned.
+
+    According to this definition, the values returned by LCSS range from
+    0 to 1, the highest value taken when two time series fully match,
+    and vice-versa. It is not required that both time series share the
+    same size, but they must be the same dimension. LCSS was originally
+    presented in [1]_ and is discussed in more details in our
+    :ref:`dedicated user-guide page <lcss>`.
```
rtavenar, line 1592: There is no such page

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    percentage of that value regarding the length of the shortest time
+    series is returned.
+
+    According to this definition, the values returned by LCSS range from
+    0 to 1, the highest value taken when two time series fully match,
+    and vice-versa. It is not required that both time series share the
+    same size, but they must be the same dimension. LCSS was originally
+    presented in [1]_ and is discussed in more details in our
+    :ref:`dedicated user-guide page <lcss>`.
+
```
rtavenar, line 1462:
```suggestion

    Note
    ----
    Contrary to Dynamic Time Warping and variants, an LCSS path does not need to be contiguous.

```

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    longest common subsequence, the percentage of that value regarding
+    the length of the shortest time series is returned.
+
+    According to this definition, the values returned by LCSS range from
+    0 to 1, the highest value taken when two time series fully match,
+    and vice-versa. It is not required that both time series share the
+    same size, but they must be the same dimension. LCSS was originally
+    presented in [1]_ and is discussed in more details in our
+    :ref:`dedicated user-guide page <lcss>`.
+
```
rtavenar, line 1593:
```suggestion

    Note
    ----
    Contrary to Dynamic Time Warping and variants, an LCSS path does not need to be contiguous.

```

tslearn/metrics/dtw_variants.py
```diff
@@ -1365,3 +1365,506 @@ def lb_envelope(ts, radius=1):
...
+    ...                      metric="precomputed")  # doctest: +ELLIPSIS
+    ([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5)], 1.0)
+
+    Notes
+    --------
+    By using a squared euclidean distance metric as shown above, the output
+    path and similarity is the same as the one obtained by using lcss_path 
+    (which uses the euclidean distance) simply because with the sum of squared
+    distances the matching threshold is still not reached.
+
```
rtavenar, line 1829:
```suggestion

    Also, contrary to Dynamic Time Warping and variants, an LCSS path does not need to be contiguous.
```

rtavenar:
Ok, I've just gone through a final review and, apart from these suggestions, this looks good to me.

tslearn/metrics/dtw_variants.py
Danisodu: What did you mean by this?

docs/examples/metrics/plot_lcss_custom_metric.py
Danisodu: If the diagonal match is preserved, then increasing the epsilon value will have no change. Should I delete it?

rtavenar: OK for me.

rtavenar:
Could you please take my last comments into account and then I'll merge?

tslearn/metrics/dtw_variants.py
Danisodu: This page is indeed missing. I will be adding it today.

rtavenar:
Thanks a lot, and congrats @Danisodu for this great work. I can now merge! 🎉

Danisodu:
Thank you @rtavenar for all your help! I am available to work on other metrics if needed. E.g. derivative dynamic time warping (ddtw)

rtavenar:
Thanks @Danisodu for your contribution, it's always a great pleasure to have new contributors on board!

Welcoming new time series metrics is definitely in our roadmap. 
I was also thinking that we should maybe work on a refactoring of the DTW Barycenter Averaging code so that it would be easier to plug any new alignment-based metric in the algorithm, which would allow, in turn, to use all these metrics in `TimeSeriesKMeans`, for example.

As a summary, feel free to give a hand on any of these topics, it will be highly appreciated!

Issue #314, opened 2020-12-11
TimeSeriesScalerMinMax gives NaN when all values are the same


GillesVandewiele:
**Describe the bug**
Scaling time series with `TimeSeriesScalerMinMax` results in NaNs, while this is not the case with `TimeSeriesScalerMeanVariance`.

**To Reproduce**
```python
>>> from tslearn.preprocessing import TimeSeriesScalerMeanVariance, TimeSeriesScalerMinMax
>>> scaler = TimeSeriesScalerMinMax()
>>> scaler.fit_transform([[0, 0, 0, 0]])
array([[[nan],
        [nan],
        [nan],
        [nan]]])
>>> scaler = TimeSeriesScalerMeanVariance()
>>> scaler.fit_transform([[0, 0, 0, 0]])
array([[[0.],
        [0.],
        [0.],
        [0.]]])
```

**Expected behavior**
I would expect a similar fix as done in `TimeSeriesScalerMeanVariance`. There, when std=0, the denominator is set to 1 instead (https://github.com/tslearn-team/tslearn/blob/775dadd/tslearn/preprocessing.py#L295). For the `TimeSeriesScalerMinMax`, we should set `range_t` to 1 where it is 0.

GillesVandewiele:
I'll work on this and open a PR in a bit.

rtavenar:
Fixed by #315 
Thanks @GillesVandewiele ;)

GillesVandewiele:
Most welcome, it was a low hanging fruit! Hopefully I can get the other ones in my TODO-list done soon.

Btw, this issue should close itself as soon as `subfolders` branch is merged into `master`.

PR #315, opened 2020-12-11
Avoid DivisionByZero in MinMaxScaler

GillesVandewiele:
This PR deals with issue #314. As a fix, I propose to set range_t to 1 where it is 0 so that no division by zero occurs.
```python
range_t[range_t == 0.] = 1.
```

I also added a test for this. We did not have a `test_preprocessing` file yet so I created it. The test is very simple:
```python
def test_single_value_ts_no_nan():
	X = to_time_series_dataset([[1, 1, 1, 1]])

	standard_scaler = TimeSeriesScalerMeanVariance()
	assert np.sum(np.isnan(standard_scaler.fit_transform(X))) == 0

	minmax_scaler = TimeSeriesScalerMinMax()
	assert np.sum(np.isnan(minmax_scaler.fit_transform(X))) == 0
```


GillesVandewiele:
It seems like I accidently pushed along two very old changes. Only #b491aa2 is relevant. I will try to undo these.

rtavenar:
Hi @GillesVandewiele 

Thanks for the quick fix! WOuld you mind using `subfolders` as base branch so that the changes don't get lost by consecutive merges?

GillesVandewiele:
Yes of course! Let me try that now

GillesVandewiele:
Is this OK @rtavenar? It seems to assign all these changes to this PR now. Want me to close this one and open a new one instead?

PR #317, opened 2020-12-13
Modernize utils package

felixdivo:
Modernizes the `tslearn.utils` module by removing code that was kept for Python versions that do not exist any more and using generator expressions and mainly simpler code in many places.

I also benchmarked the replacement for `time_series_to_str()`:
```python
# NEW:
>>> timeit(setup='import numpy as np;from io import StringIO;x = np.linspace(0,10,100); x=np.array((x,x,x))',stmt='i=StringIO();np.savetxt(i, x.T, fmt="%.18e", newline="|",encoding="bytes");s=i.getvalue()[:-1]', number=10000)
4.755633597000269

# OLD:
# already improved version with generators before complete replacement
>>> timeit(setup='import numpy as np;from tslearn.utils import time_series_to_str;x = np.linspace(0,10,100); x=np.array((x,x,x))',stmt='time_series_to_str(x)', number=10000)
4.712957062001806
```

And for `str_to_time_series()`:
```python
# NEW
timeit(setup='import numpy as np;from io import StringIO,BytesIO;from tslearn.utils import time_series_to_str,str_to_time_series,to_time_series;x = np.linspace(0,10,100); x=np.array((x,x,x));s=time_series_to_str(x)',stmt='dimensions = s.split("|");ts = [np.fromstring(dim_str, sep=" ") for dim_str in dimensions];to_time_series(np.transpose(ts))', number=10000)
2.6836724650056567

# OLD
timeit(setup='import numpy as np;from tslearn.utils import time_series_to_str,str_to_time_series;x = np.linspace(0,10,100); x=np.array((x,x,x));s=time_series_to_str(x)',stmt='str_to_time_series(s)', number=10000)
3.1848444619972724
```

I pointed this PR at the `subfolders` branch, is this correct?

I will get the CI to work and only then remove the `[DRAFT]` prefix from the title.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,4 @@ tmp-tslearn-test.txt
 docs/_build/
 docs/auto_examples/
 docs/examples/misc/ks_trained.hdf5
+docs/gen_modules/*/
--- a/tslearn/cycc.pyx
+++ b/tslearn/cycc.pyx
@@ -1,7 +1,6 @@
 STUFF_cycc = "cycc"
 
 import numpy
-from tslearn.utils import bit_length
 
 cimport numpy
 cimport cython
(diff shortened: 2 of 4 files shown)


rtavenar:
Looks good to me, thx @felixdivo !

felixdivo:
Nice! I'm also done here, just reverted the change in 85020a20f0dd5795d5749e9346cc7abe86d13287 since it seems like it is actually a valid way of raising exceptions in Python, although its kinda weird.

PR #318, opened 2020-12-14
Remove unnecessary helper method _arraylike_copy() from utils

felixdivo:
The helper method `_arraylike_copy(thing)` can be easily replaced by `numpy.array(thing, copy=True)`. It was only used in `to_time_series()`. The docstring now explicitly says that `to_time_series()` always returns a *copy*.


rtavenar:
Could you please deal with the conflict? Apart from that, it's a good idea, thx

felixdivo:
> Could you please deal with the conflict? Apart from that, it's a good idea, thx

Sure, Done.

PR #319, opened 2020-12-14
Update CI images and tools

felixdivo:
Also test on python version 3.9 and updates some VM images to their newest version. This way, we will find bugs in our repository CI arising from incompatibility with newer versions, before users will complain.

**EDIT:** This only updates the tools and not the python version to 3.9, which will be tracked in a separate issue (#322) as it is not ready to use for now.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,7 +24,7 @@ jobs:
     OMP_NUM_THREADS: '2'
     NUMBA_NUM_THREADS: '2'
     PIP_PREFER_BINARY: 'true'
-  
+
   steps:
   - task: UsePythonVersion@0
     inputs:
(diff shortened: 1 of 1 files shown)


felixdivo:
Has to wait for https://github.com/tensorflow/tensorflow/issues/44485

felixdivo:
@rtavenar Just realized that and reverted that part. :)

felixdivo:
Seems like there are actual errors caused by some of these changes: https://dev.azure.com/romaintavenard/tslearn/_build/results?buildId=670&view=logs&j=2e703997-b85c-5955-fe15-cdf93bef6757&t=b1b94f87-1d38-5554-e373-8f4cd3c9d658&l=365

Will dig a bit ...

felixdivo:
Seems like it's the new version of tensorflow: [2.4.0](https://github.com/tensorflow/tensorflow/releases/tag/v2.4.0). Previously, `tensorflow-2.3.1` was installed instead. These failures have nothing to do with the changes in this PR, as it also happened in #318.

rtavenar:
I have tried to fix the error related to tf2.4 in the subfolders branch. If tests pass there, you could just merge newest chages from subfolders in your branches and this should do the trick

felixdivo:
Fixed on that branch, tracked in #320. Will update this one.

PR #321, opened 2020-12-15
Parallelized time series k means barycenter update procedure

enricofallacara:
First of all I want to thank you for the amazing job that you have done creating this python package which is awesome and very useful. I have used it in my final thesis for a master degree in which I had to cluster some type of players based on their behaviors. The problem in using Time series k-means wiht my dataset derived from the fact that the actual implementation performs parallel computations only in distance matrices (DTW in my case),thus i re-wrote the code related to the MM algorithm and to the barycenter update proceudre in the following way: the barceynter update procedure creates a process for each cluster of the algorithm (equal to the value of k) and then each process creates a pool of threads (15 in my implementation), that are used to perform the barycenter update procedure in a parallel and efficient way.


enricofallacara:
I forgot to mention that also some redundant calls to the functions to_time_series_dataset() were removed because they caused memory allocation problems and they were basically not useful

kushalkolar:
Hi @enricofallacara , I'm not the maintainer for tslearn but I thought I'd share some comments since this would be useful for me.

1. tests are failing since you're importing `psutil` which is not in the `requirements.txt`, however you do not appear to use `psutil`.
2. `n_jobs` for joblib is hard-coded to 15, would probably be better have it as a keyword argument

enricofallacara:
> Hi @enricofallacara , I'm not the maintainer for tslearn but I thought I'd share some comments since this would be useful for me.
> 
> 1. tests are failing since you're importing `psutil` which is not in the `requirements.txt`, however you do not appear to use `psutil`.
> 2. `n_jobs` for joblib is hard-coded to 15, would probably be better have it as a keyword argument

You are completly right, sorry for the psutil mistake. I have used it for some stuffs and i forgot to remove it. Also, I know that I had hard-coded the number of threads, but I was in a hurry and I loaded my implementation. Tomorrow I will fix this stuffs, thanks!

rtavenar:
Hi @enricofallacara 

Your PR is now in conflict with the master branch since we have refactored the code. I can take care of the merge to resolve the conflicts if you want, after which I could do a review of your code.

enricofallacara:
Hello,

No problem with that. Sorry, I didn't have time to fix the code due to work
problems, I really appreciate your help.

Il giorno mar 5 gen 2021 alle ore 11:00 Romain Tavenard <
notifications@github.com> ha scritto:

tslearn/barycenters/dba.py
```diff
@@ -148,7 +149,8 @@ def dtw_barycenter_averaging_petitjean(X, barycenter_size=None,
        for dynamic time warping, with applications to clustering. Pattern
        Recognition, Elsevier, 2011, Vol. 44, Num. 3, pp. 678-693
     """
-    X_ = to_time_series_dataset(X)
```
rtavenar, line 151: I understand that you want to avoid a copy here, but making sure that the data is in the right format is important. Maybe we shoud tweak the `to_time_series_dataset` function to add an `avoid_copy_if_possible` parameter, WDYT?

tslearn/barycenters/dba.py
```diff
@@ -176,7 +178,7 @@ def dtw_barycenter_averaging_petitjean(X, barycenter_size=None,
     return barycenter
 
 
-def _mm_assignment(X, barycenter, weights, metric_params=None):
+def _mm_assignment(X, barycenter, weights, metric_params=None, n_treads=15):
```
rtavenar, line 181: We use `n_jobs` everywhere as a parameter name, and `None` as its default, we should stick to the same convention here, I think.

tslearn/barycenters/dba.py
```diff
@@ -210,15 +212,23 @@ def _mm_assignment(X, barycenter, weights, metric_params=None):
...
     cost = 0.
     list_p_k = []
-    for i in range(n):
-        path, dist_i = dtw_path(barycenter, X[i], **metric_params)
-        cost += dist_i ** 2 * weights[i]
-        list_p_k.append(path)
+    #with parallel_backend('threading'):
+    res = Parallel(backend='threading',prefer="threads",require='sharedmem',n_jobs=n_treads,verbose=10)(delayed(dtw_path)(barycenter, X[i], global_constraint="sakoe_chiba",sakoe_chiba_radius=1) for i in range(n))
+    paths, dists = zip(*res)
+    paths = list(paths)
```
rtavenar, line 218: Do you really need these 2 casts?

tslearn/barycenters/dba.py
```diff
@@ -210,15 +212,23 @@ def _mm_assignment(X, barycenter, weights, metric_params=None):
     n = X.shape[0]
     cost = 0.
     list_p_k = []
-    for i in range(n):
-        path, dist_i = dtw_path(barycenter, X[i], **metric_params)
-        cost += dist_i ** 2 * weights[i]
-        list_p_k.append(path)
+    #with parallel_backend('threading'):
+    res = Parallel(backend='threading',prefer="threads",require='sharedmem',n_jobs=n_treads,verbose=10)(delayed(dtw_path)(barycenter, X[i], global_constraint="sakoe_chiba",sakoe_chiba_radius=1) for i in range(n))
```
rtavenar, line 216: Should we define a `cdist_dtw_path` function instead?

tslearn/barycenters/dba.py
```diff
@@ -210,15 +212,23 @@ def _mm_assignment(X, barycenter, weights, metric_params=None):
...
+    #for i in range(n):
+    #    path, dist_i = dtw_path(barycenter, X[i], **metric_params)
+    #    cost += dist_i ** 2 * weights[i]
+    #    list_p_k.append(path)
+    #cost /= weights.sum()
     return list_p_k, cost
 
 
-def _subgradient_valence_warping(list_p_k, barycenter_size, weights):
+def _subgradient_valence_warping(list_p_k, barycenter_size, weights, n_treads):
```
rtavenar, line 231: We use `n_jobs` everywhere as a parameter name, and `None` as its default, we should stick to the same convention here, I think.

tslearn/barycenters/dba.py
```diff
@@ -253,19 +263,22 @@ def _subgradient_valence_warping(list_p_k, barycenter_size, weights):
...
-        sz_k = p_k[-1][1] + 1
-        w_k = numpy.zeros((barycenter_size, sz_k))
+    def create_w_k(p_k, barycenter_size):
+        w_k = numpy.zeros((barycenter_size, p_k[-1][1] + 1))
         for i, j in p_k:
             w_k[i, j] = 1.
-        list_w_k.append(w_k * weights[k])
-        list_v_k.append(w_k.sum(axis=1) * weights[k])
+        return w_k
+    w_k_ones = Parallel(backend='threading',prefer='threads',require='sharedmem', n_jobs=n_treads, verbose = True)(delayed(create_w_k)(p_k,barycenter_size) for p_k in list_p_k)
```
rtavenar, line 271: Is this part really crucial to parallelize? I would expect that it would be neglectible in terms of running times.

tslearn/barycenters/dba.py
```diff
@@ -253,19 +263,22 @@ def _subgradient_valence_warping(list_p_k, barycenter_size, weights):
...
+        return w_k * weight
+    list_w_k = Parallel(backend='threading',prefer='threads',require='sharedmem', n_jobs=n_treads, verbose = True)(delayed(mult_w_k_weights)(w_k,weights[k]) for k,w_k in enumerate(w_k_ones))
+    def mult_w_k_sum_wieghts(w_k, weight):
+        return w_k.sum(axis=1) * weight
+    list_v_k = Parallel(backend='threading',prefer='threads',require='sharedmem', n_jobs=n_treads, verbose = True)(delayed(mult_w_k_sum_wieghts)(w_k,weight) for (w_k, weight) in zip(w_k_ones, weights))
     return list_v_k, list_w_k
 
 
-def _mm_valence_warping(list_p_k, barycenter_size, weights):
+def _mm_valence_warping(list_p_k, barycenter_size, weights, n_treads):
```
rtavenar, line 281: We use `n_jobs` everywhere as a parameter name, and `None` as its default, we should stick to the same convention here, I think.

tslearn/barycenters/dba.py
```diff
@@ -304,13 +317,15 @@ def _mm_valence_warping(list_p_k, barycenter_size, weights):
         list_p_k=list_p_k,
         barycenter_size=barycenter_size,
         weights=weights)
-    diag_sum_v_k = numpy.zeros(list_v_k[0].shape)
-    for v_k in list_v_k:
-        diag_sum_v_k += v_k
+    #diag_sum_v_k = numpy.zeros(list_v_k[0].shape)
+    #for v_k in list_v_k:
+    #    diag_sum_v_k += v_k
+    diag_sum_v_k = Parallel(backend='threading',prefer='threads',require='sharedmem', n_jobs=n_treads, verbose = True)(delayed(numpy.sum)(x) for x in zip(*numpy.array(list_v_k)))
```
rtavenar, line 323: Once again, I'm not sure we need to parallelize this part.

tslearn/barycenters/dba.py
```diff
@@ -304,13 +317,15 @@ def _mm_valence_warping(list_p_k, barycenter_size, weights):
...
+    diag_sum_v_k = numpy.array(diag_sum_v_k)
     return diag_sum_v_k, list_w_k
 
 
-def _mm_update_barycenter(X, diag_sum_v_k, list_w_k):
+def _mm_update_barycenter(X, diag_sum_v_k, list_w_k, n_treads):
```
rtavenar, line 328: We use `n_jobs` everywhere as a parameter name, and `None` as its default, we should stick to the same convention here, I think.

tslearn/barycenters/dba.py
```diff
@@ -336,12 +351,17 @@ def _mm_update_barycenter(X, diag_sum_v_k, list_w_k):
...
-    sum_w_x = numpy.zeros((barycenter_size, d))
-    for k, (w_k, x_k) in enumerate(zip(list_w_k, X)):
-        sum_w_x += w_k.dot(x_k[:ts_size(x_k)])
-    barycenter = numpy.diag(1. / diag_sum_v_k).dot(sum_w_x)
+    #d = X.shape[2]
+    #barycenter_size = diag_sum_v_k.shape[0]
+    #sum_w_x = numpy.zeros((barycenter_size, d))
+    #for k, (w_k, x_k) in enumerate(zip(list_w_k, X)):
+    #    sum_w_x += w_k.dot(x_k[:ts_size(x_k)])
+    sum_w_x = Parallel(backend='threading',prefer='threads',require='sharedmem', n_jobs=n_treads, verbose = True)(delayed(numpy.dot)(w_k,x_k[:ts_size(x_k)]) for (w_k, x_k) in zip(list_w_k, X))
```
rtavenar, line 359: Once again, I'm not sure we need to parallelize this part.

tslearn/barycenters/dba.py
```diff
@@ -392,7 +412,7 @@ def _subgradient_update_barycenter(X, list_diag_v_k, list_w_k, weights_sum,
 
 
 def dtw_barycenter_averaging(X, barycenter_size=None, init_barycenter=None,
-                             max_iter=30, tol=1e-5, weights=None,
+                             max_iter=30, tol=1e-5, weights=None, n_treads=15,
```
rtavenar, line 415: We use `n_jobs` everywhere as a parameter name, and `None` as its default, we should stick to the same convention here, I think.

tslearn/barycenters/dba.py
```diff
@@ -514,6 +534,7 @@ def dtw_barycenter_averaging(X, barycenter_size=None, init_barycenter=None,
 def dtw_barycenter_averaging_one_init(X, barycenter_size=None,
                                       init_barycenter=None,
                                       max_iter=30, tol=1e-5, weights=None,
+									  n_treads=15,
```
rtavenar, line 537: We use `n_jobs` everywhere as a parameter name, and `None` as its default, we should stick to the same convention here, I think.

tslearn/barycenters/dba.py
```diff
@@ -576,23 +597,23 @@ def dtw_barycenter_averaging_one_init(X, barycenter_size=None,
        for Averaging in Dynamic Time Warping Spaces.
        Pattern Recognition, 74, 340-358.
     """
-    X_ = to_time_series_dataset(X)
```
rtavenar, line 579: See comment above about the array copy issue.

tslearn/clustering/kmeans.py
```diff
@@ -692,23 +693,37 @@ def _assign(self, X, update_class_attributes=True):
 
     def _update_centroids(self, X):
         metric_params = self._get_metric_params()
+        X_list = []
```
rtavenar, line 696: Here, I guess we could use a comprehension list for better readability.

tslearn/clustering/kmeans.py
```diff
@@ -692,23 +693,37 @@ def _assign(self, X, update_class_attributes=True):
...
-                self.cluster_centers_[k] = softdtw_barycenter(
-                    X=X[self.labels_ == k],
-                    max_iter=self.max_iter_barycenter,
-                    init=self.cluster_centers_[k],
-                    **metric_params)
-            else:
-                self.cluster_centers_[k] = euclidean_barycenter(
-                    X=X[self.labels_ == k])
+           X_list.append(X[self.labels_ == k])
+        cluster_centers = Parallel(backend='threading', prefer='threads', require='sharedmem', n_jobs=self.n_clusters, verbose=5)(delayed(dtw_barycenter_averaging)(
```
rtavenar, line 699: Take care that here, the function to be called depends on the value of `self.metric`

tslearn/metrics/utils.py
```diff
@@ -9,48 +9,39 @@ def _cdist_generic(dist_fun, dataset1, dataset2, n_jobs, verbose,
                    compute_diagonal=True, dtype=numpy.float, *args, **kwargs):
     """Compute cross-similarity matrix with joblib parallelization for a given
     similarity function.
-
```
rtavenar, line 12: Could you please revert all these line suppressions? They tend to pack the docstrings.

tslearn/metrics/utils.py
```diff
@@ -60,7 +51,7 @@ def _cdist_generic(dist_fun, dataset1, dataset2, n_jobs, verbose,
                                      k=0 if compute_diagonal else 1,
                                      m=len(dataset1))
         matrix[indices] = Parallel(n_jobs=n_jobs,
-                                   prefer="threads",
```
rtavenar, line 63: Should we always prefer this? Or should it be specified when calling the function?

tslearn/metrics/utils.py
```diff
@@ -74,12 +65,13 @@ def _cdist_generic(dist_fun, dataset1, dataset2, n_jobs, verbose,
         matrix[indices] = matrix.T[indices]
         return matrix
     else:
-        dataset2 = to_time_series_dataset(dataset2, dtype=dtype)
-        matrix = Parallel(n_jobs=n_jobs, prefer="threads", verbose=verbose)(
-            delayed(dist_fun)(
-                dataset1[i], dataset2[j],
-                *args, **kwargs
+        #dataset2 = to_time_series_dataset(dataset2, dtype=dtype)
+        with parallel_backend('loky'):
```
rtavenar, line 69: Is it a good idea to specify the `parallel_backend` here? Will loky always be a better alternative or could it depend on some specificities or one's data?

rtavenar:
I have left a few comments. Maybe the main point would be to assess the improvement (in terms of running times) that these changes offer on some benchmark.

rtavenar:
Also, tests seem to fail at the moment (let me know if it's due to a bug in my merge).

tslearn/barycenters/dba.py
PercyLau: I think we should design a high-efficiency cdist_dtw_path function instead. The main bottlenecks are _mm_assignment function and _subgradient_valence_warping from my view. However, as DBA is communication-intensive, we require a good tune on parallelizing DBA's implementation. Perhaps, a good target is to ensure that all the available computing units are at the desired utilization level.

tslearn/clustering/kmeans.py
PercyLau: Simply parallelize DBA over the clusters only speedup less than 0.5x times. More effort is required.

tslearn/barycenters/dba.py
PercyLau: This function cannot fully utilize CPU cores, even if we force to parallelize dba on different clusters and then becomes the bottleneck.

charavelg:
Some work has been done in #658, notably the optimization in the barycenters computation algorithm. The rework of of // computation per cluster is probably to much effort for very little gain. Closing this.

Issue #327, opened 2020-12-25
ModuleNotFoundError: No module named 'sklearn.cluster.k_means_'


philip-zelichenko:
**Describe the bug**
ModuleNotFoundError error when importing  'from tslearn.clustering import TimeSeriesKMeans'

**To Reproduce**
'from tslearn.clustering import TimeSeriesKMeans'

**Expected behavior**
no error

**Environment (please complete the following information):**
 - OS: Linux 2020 x86_64 x86_64 x86_64 GNU/Linux
 - tslearn version '0.4.1'

**Additional context**
tslearn is trying to import 'ModuleNotFoundError: No module named 'sklearn.cluster.k_means_'
although the module name is '_kmeans' under sklearn '0.24.0'

shima-create:
Has this bug been solved?
Is it possible to install another version of tslearn or scikit-learn to solve the issue?

chusloj:
@philip-zelichenko The `scikit-learn` docs show that the `_k_means` module has been changed to [`KMeans`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html). I'll open a PR soon to address this in `clustering.py`.

rtavenar:
Thanks @chusloj to work on this. While waiting for the fix, one can roll back to sklearn 0.23.* for the clustering module in tslearn to work properly.

rtavenar:
This is now fixed in master (and will be available in our next release).

PR #328, opened 2020-12-30
[MRG] Fix ModuleNotFoundError with Scikit-Learn 0.24 KMeans module

chusloj:
Addresses #327.

This PR fixes an issue where `ModuleNotFoundError` is thrown with `scikit.clustering._k_means` on `scikit-learn` 0.24. In version 0.24, the clustering `_k_means` module was [renamed](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html) to `scikit.clustering.KMeans`.

Additionally, the `n_jobs` argument to the `KMeans` module was deprecated, so it was removed from `TimeSeriesKMeans()` (the argument passed to `n_jobs` was `None` anyway).

**Environment:**
- `python` – 3.8.6
- `scikit-learn` - 0.24


rtavenar:
Hi @chusloj 

Thanks a lot for giving a hand on this!

I'm not sure what you are trying to do here. It seems to me (tell me if I'm wrong) that your option is to perform a standard kmeans (from sklearn) with 0 iterations so that only the initialization is performed. If that is correct, that sounds like a nice way to do things (and it allows not to have specific hacks to deal with the different names of the kmeans++ initialization routine for different versions of sklearn).

Yet I think that the `KMeans` object is returned by fit, and not the centers. Also, you should specify the number of iterations to be 0 explicitly so that only the initialization is performed.

Finally, I have just merged a major refactoring of the code (created subfolders for subpackages, see #273 ) so I guess it would be better for you to start from this new version.

chusloj:
@rtavenar I'll merge in changes from #273 and then reply to your other comments.

chusloj:
https://github.com/scikit-learn/scikit-learn/blob/42aff4e2edd8e8887478f6ff1628f27de97be6a3/sklearn/cluster/_kmeans.py#L1928-L1930

chusloj:
@rtavenar I misunderstood the code I implemented. I accidentally implemented an instance of the `KMeans` class rather than the initialization heuristic used previously in this code. In `scikit-learn` 0.24, the `_k_means` source file is renamed to `_kmeans`, and the `_k_init` method is renamed to [`kmeans_plusplus`](https://github.com/scikit-learn/scikit-learn/blob/6b4f82433dc2f219dbff7fe8fa42c10b72379be6/sklearn/cluster/_kmeans.py#L1929-L1931):

``` python
def kmeans_plusplus(X, n_clusters, *, x_squared_norms=None,
                    random_state=None, n_local_trials=None):
    """Init n_clusters seeds according to k-means++
```

All relevant tests now pass.

tslearn/clustering/kshape.py
```diff
@@ -1,11 +1,6 @@
 from sklearn.base import ClusterMixin
 
-try:
-    # Most recent
-    from sklearn.cluster._k_means import _k_init
-except ImportError:
-    # Deprecated from sklearn v0.24 onwards
-    from sklearn.cluster.k_means_ import _k_init
+from sklearn.cluster._kmeans import kmeans_plusplus
```
rtavenar, line 3: This one can be removed: not used in `kshape.py`

tslearn/clustering/kshape.py
```diff
@@ -294,4 +289,4 @@ def predict(self, X):
         X_ = check_dims(X, X_fit_dims=self.cluster_centers_.shape)
         X_ = TimeSeriesScalerMeanVariance(mu=0., std=1.).fit_transform(X_)
         dists = self._cross_dists(X_)
-        return dists.argmin(axis=1)
+        return dists.argmin(axis=1)
```
rtavenar, line 292: Could you revert that to keep a newline at the end of file?

tslearn/clustering/kmeans.py
```diff
@@ -840,4 +835,4 @@ def transform(self, X):
         return self._transform(X)
 
     def _more_tags(self):
-        return {'allow_nan': True, 'allow_variable_length': True}
+        return {'allow_nan': True, 'allow_variable_length': True}
```
rtavenar, line 838: Could you revert that to keep a newline at the end of file?

rtavenar:
Thanks for the update!

It would be better if things could work both with sklearn 0.24 and older versions.
An ideal setup could be to print a warning if an old version is used and state that it could be deprecated in a future release of `tslearn`.

For your information, the correct imports seem to be (to be checked):

- `from sklearn.cluster._kmeans import kmeans_plusplus` in sklearn v0.24 onwards
- `from sklearn.cluster._kmeans import _k_init` in sklearn v0.22 and v0.23
- `from sklearn.cluster.k_means_ import _k_init` in sklearn v0.21 and before

Also, note that `kmeans_plusplus` returns both the centroids and the index of the centroids in the input dataset, contrary to `_k_init`.

Finally, once all is sorted, the requirements in :
- `requirements.txt`
- `requirements_nocast.txt`
- `docs/requirements_rtd.txt`

 should be updated by removing the constraint `<0.24` on sklearn version, and the changes should be (briefly) documented in `CHANGELOG.md`.

rtavenar:
Hi @chusloj 

Do you think you will have time to implement these changes in the coming days? Otherwise, I could do it myself, let me know. We'd like to release a new version of tslearn soon and this PR should definitely be included in that release.

chusloj:
@rtavenar I'll implement these changes Today.

tslearn/clustering/kmeans.py
```diff
@@ -610,12 +605,12 @@ def _fit_one_init(self, X, x_squared_norms, rs):
             self.cluster_centers_ = self.init.copy()
         elif self.init == "k-means++":
             if self.metric == "euclidean":
-                self.cluster_centers_ = _k_init(
+                self.cluster_centers_ = _kmeans_plusplus(
                     X.reshape((n_ts, -1)),
                     self.n_clusters,
                     x_squared_norms,
```
ad299, line 611: To make it works, you also need to identify last 2 parameters :
x_squared_norms = x_squared_norms[0] (because of the dimension needed here)
random_state = rs

_kmeans_pluplus take only 2 positional arguments

chusloj: @ad299 I will properly specify the last 2 input arguments to the `_kmeans_plusplus` function. However, I'm not sure that only the first row of `x_squared_norms` needs to be specified. The `_k_init` function from `scikit-learn` 0.23 and this `_kmeans_plusplus` function from 0.24 take the `x_squared_norms` input argument with identical dimensions.

chusloj: @ad299 Could you explain why the `x_squared_norms` argument should only be passed the the first row of the`x_squared_norms` variable created in this file? I ask because `_kmeans_plusplus` from `scikit-learn` 0.24 and `_k_init` from 0.23 accept the `x_squared_norms` argument with the same dimensions.

chusloj:
@rtavenar Warnings will now be thrown with `scikit-learn` < 0.24 and all `requirements` files have been updates to release the version lock on `scikit-learn`.

rtavenar:
I just tested locally for earlier versions of sklearn and did a small patch that was necessary for tests in kmeans.py to pass. Now we still have an issue with sklearns.utils.testing being removed. I will have a look tomorrow at how this should be fixed.

tslearn/clustering/kmeans.py
ad299: When i test it without selecting the first row I have an error, this is why. But if both functions need the same dimensions dont consider my test, problem is probably on an other part  in my test !

rtavenar:
I've just had a look. It seems the remaining errors are due to changes in the signatures of sklearn check utilities.
This probably would require an important code refactor, which is probably not feasible if we want a release soon.
So I would suggest that while this is not fixed, we run complete tests using sklearn 0.23 and partial ones (all checks except `test_all_estimators`) on sklearn 0.24 (to check for estimator incompatibilities).

chusloj:
Tests with `scikit-learn` 0.24 (omitting `test_all_estimators`) pass locally for me with `Python==3.8`.

rtavenar:
Thanks @chusloj for your contribution! This is now merged to master.

PR #332, opened 2021-01-05
Update kmeans.py

ad299:
I don't understand this line, sklearn.cluster._k_means doesn't exist in sklearn 0.24, and _kmeans doesn't have _k_init...


rtavenar:
This is actually investigated in #328

rtavenar:
Closing since the merge of #328 into master fixes the issue.

PR #333, opened 2021-01-07
[MRG] add TimeSeriesScaleMeanMaxVariance()

tonylee2016:
When I was using the package for K-shape, one issue found is that the original `TimeSeriesScaleMeanVariance` may amplify small changes (noises) in the signal. Because that it normalizes each signal per its own variance. 

Here I am adding a new scaling class that preserves the amplitude relation across signals, this scaling method may be useful for some applications.

`TimeSeriesScaleMeanMaxVariance` is similar to the existing `TimeSeriesScaleMeanVariance` with one exception, the max std of all signals is used for normalization. As a result, the amplitude relation across signals is preserved.

### DIFF
--- a/tslearn/preprocessing/preprocessing.py
+++ b/tslearn/preprocessing/preprocessing.py
@@ -296,3 +296,91 @@ class TimeSeriesScalerMeanVariance(TransformerMixin, TimeSeriesBaseEstimator):
 
     def _more_tags(self):
         return {'allow_nan': True}
+
+
+class TimeSeriesScaleMeanMaxVariance(TransformerMixin, TimeSeriesBaseEstimator):
+    """Scaler for time series. Scales time series so that their mean (resp.
+    standard deviation) in the signal with the max amplitue  is
+    mu (resp. std). The scaling relationships between each signal are preserved
+    This is supplement to the TimeSeriesScalerMeanVariance method
+
+    Parameters
+    ----------
+    mu : float (default: 0.)
+        Mean of the output time series.
+    std : float (default: 1.)
+        Standard deviation of the output time series.
+
+    Notes
+    -----
+        This method requires a dataset of equal-sized time series.
+
+        NaNs within a time series are ignored when calculating mu and std.
+    """
+
+    def __init__(self, mu=0., std=1.):
+        self.mu = mu
+        self.std = std
+
+    def fit(self, X, y=None, **kwargs):
+        """A dummy method such that it complies to the sklearn requirements.
+        Since this method is completely stateless, it just returns itself.
+
+        Parameters
+        ----------
+        X
+            Ignored
+
+        Returns
+        -------
+        self
+        """
+        X = check_array(X, allow_nd=True, force_all_finite=False)
+        X = to_time_series_dataset(X)
+        self._X_fit_dims = X.shape
+        return self
+
+    def fit_transform(self, X, y=None, **kwargs):
+        """Fit to data, then transform it.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_ts, sz, d)
+            Time series dataset to be rescaled.
+
+        Returns
+        -------
+        numpy.ndarray
+            Resampled time series dataset.
+        """
+        return self.fit(X).transform(X)
+
+    def transform(self, X, y=None, **kwargs):
+        """Fit to data, then transform it.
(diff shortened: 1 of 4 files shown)


tonylee2016:
the method won't pass `check_methods_subset_invariance`, It is expected, I've excluded it from the test.

tslearn/preprocessing/preprocessing.py
```diff
@@ -296,3 +296,57 @@ def transform(self, X, y=None, **kwargs):
...
+        check_is_fitted(self, '_X_fit_dims')
+        X = check_array(X, allow_nd=True, force_all_finite=False)
+        X_ = to_time_series_dataset(X)
+        X_ = check_dims(X_, X_fit_dims=self._X_fit_dims, extend=False)
+        mean_t = numpy.nanmean(X_, axis=1, keepdims=True)
+        std_t = numpy.nanstd(X_, axis=1, keepdims=True)
+        # retain the scaling relation cross the signals,
+        # the max std_t is set to self.std
+        max_std = max(std_t)
+        if max_std ==0.: max_std = 1
```
GillesVandewiele, line 343: Split up over multiple lines.

tslearn/preprocessing/preprocessing.py
```diff
@@ -296,3 +296,57 @@ def transform(self, X, y=None, **kwargs):
...
+        numpy.ndarray
+            Rescaled time series dataset
+        """
+        check_is_fitted(self, '_X_fit_dims')
+        X = check_array(X, allow_nd=True, force_all_finite=False)
+        X_ = to_time_series_dataset(X)
+        X_ = check_dims(X_, X_fit_dims=self._X_fit_dims, extend=False)
+        mean_t = numpy.nanmean(X_, axis=1, keepdims=True)
+        std_t = numpy.nanstd(X_, axis=1, keepdims=True)
+        # retain the scaling relation cross the signals,
```
GillesVandewiele, line 340: Are these comments still relevant? Does it need to be stateful? I.e. should there be a fit() and transform() where in the fit() the self.std is stored and used in the transform() method?

tslearn/preprocessing/preprocessing.py
GillesVandewiele: And add space to right and left of `==`

GillesVandewiele:
Hi,

Thank you for this PR! I think this could indeed be useful for some use cases, but I will let @rtavenar confirm this.

I left 2 very minor remarks. I checked out the `check_methods_subset_invariance` from sklearn and it indeed makes sense that the results change depending on the sub-batch you provide to the method (as it uses the max std of that batch to normalise). However, we must ensure that only the relevant tests are skipped, and that as many as possible tests are performed. Are we sure that adding `_skip_test` as a tag does not skip more tags than needed?

GillesVandewiele:
One other thing that we could consider is, instead of making this part of the main codebase, to create an example/tutorial "defining a custom preprocessor" and then using this specific preprocessing technique. I can imagine that there are more alternatives to normalise out there, and they only have minimal differences. Including all of those into tslearn could quickly result in a bloated code-base.

rtavenar:
Hi @tonylee2016 
I think this is closely related to #280 and #285

The scalers we have implemented so far only rely on per-time series information. And we seem to have more and more requests like this one.

I would be OK to have such scalers in tslearn but I strongly believe that we should put effort in a clear documentation of the differences between scalers.

Also, maybe it would be worth it to have names that explicitly state that these scalers rely on global information, or have a subpackage dedicated to "global" scalers.

I am open to suggestions on these points and would like to clarify that before considering the add of these preprocessors in tslearn.

tonylee2016:
Hi, @GillesVandewiele and @rtavenar 

thanks for the comments and extra info. After reading #280 and #285, here are my thoughts on a more generic scaler framework

1. move the std and mu calculation to `fit()`. As #285 is already doing it.
2. add an optional scaler Infos gathered from `fit()`:  so that user know what data the scaler is based on 
3. better document and an example for `TimeSeriesScaleMeanMaxVariance()` to show the different effects

I can spend some time on 3 in this PR. But 2 and 1 may be handled by #285?

 @GillesVandewiele `_skip_test` will skip all estimator checks. However, I did not find a way to only skip `check_methods_subset_invariance`.

PR #355, opened 2021-05-13
Fixed bug in dtw_path when one input is all nans or with empty length

### LINKED ISSUE #354: tslearn.metrics.dtw_variants.dtw_path gets stuck when one input is all nans or an empty array
**Describe the bug**
The function in `tslearn.metrics.dtw_variants.dtw_path` gets stuck when one input is an empty array with length 0 or it is composed only of nans (because nans are removed and an empty array is left).

The real issue comes from the function `tslearn.metrics.dtw_variants._return_path` that with empty input enters in an infinite loop.

This issue has strong repercussions because a lot of other tslearn functionalities use this function (e.g., dtw based barycenter routines)

**To Reproduce**
This will never terminate
```
from tslearn.metrics.dtw_variants import dtw_path
s1 = np.zeros((3, 10))
s2_nan = np.full((3, 10), np.nan)
dtw_path(s1, s2_nan)
```

lkubin:
This pull request fixes #354 and raises a ValueError when one input of dtw_path is an empty sequence or is a only nans sequence. The root of the bug is the numba.njit decorated function `_return_path`, however the numba documentation states that `raise` construct is partially supported inside njit decorated function.

I also added a test to ensure that this issue would never occur.

### DIFF
--- a/docs/requirements_rtd.txt
+++ b/docs/requirements_rtd.txt
@@ -12,5 +12,5 @@ tensorflow>=2
 Pygments
 numba
 sphinx_bootstrap_theme
-git+git://github.com/numpy/numpydoc@master
+git+git://github.com/numpy/numpydoc@main
 matplotlib
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy", "Cython"]
+requires = ["setuptools", "wheel", "numpy<=1.19", "Cython"]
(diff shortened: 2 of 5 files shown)


rtavenar:
Hi @lkubin 

Thanks for spotting the bug, and for your suggested fix.
I believe the docs + build fails are unrelated (I will try to fix them in parallel).
My minor comment is: could you change the error from "One of the input time series contains only nans or has empty length." to "One of the input time series contains only nans or has zero length." (and do the same in the tests)?

rtavenar:
@lkubin you could just merge changes from branch `fix_deps` and that should make it for your failing tests.

lkubin:
@rtavenar thank you for your review and for your suggestion. I just updated the error message raised by the function `dtw_path` and I merged the branch `fix_deps`.

rtavenar:
Great! I'll merge a publish a minor tslearn version as soon as I find time!

Thanks for your contribution!

PR #356, opened 2021-05-17
Attempt to fix Cython errors in recent builds

rtavenar:
This implements:

* https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-832706637
* also numpydoc has renamed its `master` to `main`, so this has to be updated accordingly

Behavior discovered in #355


rtavenar:
This has been merged as part of #355

PR #360, opened 2021-08-16
Fix deprecation warning for np array

brk21:
/databricks/python/lib/python3.8/site-packages/tslearn/barycenters/softdtw.py:103: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  X_ = numpy.array([to_time_series(d, remove_nans=True) for d in X_])

Because this method often uses numpy arrays with different shapes, I think the approach is justified here.

### DIFF
--- a/tslearn/barycenters/softdtw.py
+++ b/tslearn/barycenters/softdtw.py
@@ -100,7 +100,7 @@ def softdtw_barycenter(X, gamma=1.0, weights=None, method="L-BFGS-B", tol=1e-3,
         barycenter = init
 
     if max_iter > 0:
-        X_ = numpy.array([to_time_series(d, remove_nans=True) for d in X_])
+        X_ = numpy.array([to_time_series(d, remove_nans=True) for d in X_],dtype=object)
 
         def f(Z):
             return _softdtw_func(Z, X_, weights, barycenter, gamma)


rtavenar:
Hi @brk21 

Thanks for pointing out this deprecation warning! It seems to me that it is not even necessary to cast the list into a numpy array, or is it?

Couldn't we do something like:

```python
X_ = [to_time_series(Xi, remove_nans=True) for Xi in X_]
```
?

brk21:
Hi @rtavenar,

I think you are right, but I also think we will generate the same warning in the `to_time_series` function here: 

https://github.com/tslearn-team/tslearn/blob/main/tslearn/utils/utils.py#L146

So you may end up adding the `object=true` parameter there to avoid the deprecation.

Please let me know if I can support in any way.

Ross

rtavenar:
I would like to avoid the `object=True` since our algorithms in `tslearn` are supposed to operate on datasets represented as 3d arrays, which is no longer the case when arrays of objects are used.
Regarding the `to_time_series` function, it is definitely unexpected behaviour to provide a series that does not have the same number of features for all timestamps, so maybe we should throw a `ValueError` in this case, what do you think?

rtavenar:
> @rtavenar I would be strongly opposed to that because one of the main values of `softdtw` is that it can operate on time series of different lengths! This `ValueError` would effectively eliminate our ability to make use of `softdtw` from `tslearn`.

I meant raising a `ValueError` for the case of the `to_time_series` function only, since this function only deals with one time series at a time, not a dataset of time series (for which `to_time_series_dataset` exists). Your workaround or the one I suggested earlier are fine for softDTW.

brk21:
Phew @rtavenar Apologies for the misunderstanding. 

Yes, I would be fine with that.

PR #361, opened 2021-09-02
Add license information to package metadata

ecederstrand:
This makes it easier for automatic license checkers to verify the license of this package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     description="A machine learning toolkit dedicated to time-series data",
     long_description=README,
     long_description_content_type='text/markdown',
-    license='2-clause BSD License',
+    license='BSD-2-Clause',
     include_dirs=[numpy.get_include()],
     ext_modules=cythonize("tslearn/metrics/*.pyx",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -28,6 +28,7 @@
     description="A machine learning toolkit dedicated to time-series data",
     long_description=README,
     long_description_content_type='text/markdown',
+    license='2-clause BSD License',
```
felixdivo, line 31: This would be the [correct SPDX identifier](https://spdx.org/licenses/BSD-2-Clause.html):

```suggestion
    license='BSD 2-Clause "Simplified" License',
```

ecederstrand: SPDX and OSI apparently disagree on the long name for this license (see https://opensource.org/licenses/BSD-2-Clause and https://spdx.org/licenses/BSD-2-Clause.html). I've switched to the short identifier which they do agree on.

felixdivo: Looks good 👍

rtavenar:
Thanks @ecederstrand 

This is no merge to `main` branch, will be included in the next release :)

Issue #271, opened 2020-06-29
DTW (and probably other metrics too) don't raise an error for heterogeneous time series


rtavenar:
**Describe the bug**
When fed with time series of different dimensionalities, DTW (and other methods that are not designed to handle heterogeneous time series (ie time series whose features do not lie in the same ambient space) should raise an error, which is not always the case.

**To Reproduce**

```python
>>> import numpy as np
>>> from tslearn.metrics import dtw

>>> 
>>> x = np.random.randn(10, 1)
>>> y = np.random.randn(10, 2)
>>> dtw(x, y)
4.794730857157181
>>> dtw(y, x)
inf
```

**Expected behavior**
Raise an error as soon as dimensionalities (number of features) do not match.

Thanks @lchapel for the notice ;)

PR #365, opened 2021-10-07
[MRG] Check time series shape and output meaningful error for DTW

kashee337:
I fixed issue #271 .
I implemented dimensions of timeseries check that raises ValueError('dimension mismatch') when the feature dimensions of the two input timeseries(s1,s2) are different.
Also, if I'm not mistaken, I think this process is also needed in the dtw_path function, so I added it there as well.

I would be happy to receive your feedback if this pull request is not appropriate.
Thanks.
(Sorry, the name of the branch was inappropriate, so I resubmit the PR again.)

### DIFF
--- a/tslearn/metrics/dtw_variants.py
+++ b/tslearn/metrics/dtw_variants.py
@@ -194,7 +194,7 @@ def dtw_path(s1, s2, global_constraint=None, sakoe_chiba_radius=None,
             "One of the input time series contains only nans or has zero length.")
 
     if s1.shape[1] != s2.shape[1]:
-        raise ValueError('dimension mismatch')
+        raise ValueError("All input time series must be the same shape.")
 
     mask = compute_mask(
         s1, s2, GLOBAL_CONSTRAINT_CODE[global_constraint],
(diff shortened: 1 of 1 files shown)


tslearn/metrics/dtw_variants.py
```diff
@@ -190,7 +190,11 @@ def dtw_path(s1, s2, global_constraint=None, sakoe_chiba_radius=None,
     s2 = to_time_series(s2, remove_nans=True)
 
     if len(s1) == 0 or len(s2) == 0:
-        raise ValueError("One of the input time series contains only nans or has zero length.")
+        raise ValueError(
+            "One of the input time series contains only nans or has zero length.")
+
+    if s1.shape[1] != s2.shape[1]:
+        raise ValueError('dimension mismatch')
```
rtavenar, line 197: Sorry that it took me so long to answer...

Could you provide an error message that would be more informative?

kashee337: Thank you for your reply.

Ok, I'll fix it.

tslearn/metrics/dtw_variants.py
```diff
@@ -194,7 +194,7 @@ def dtw_path(s1, s2, global_constraint=None, sakoe_chiba_radius=None,
             "One of the input time series contains only nans or has zero length.")
 
     if s1.shape[1] != s2.shape[1]:
-        raise ValueError('dimension mismatch')
+        raise ValueError("All input time series must be the same shape.")
```
rtavenar, line 197: Well, they should not be the same shape (they can be of different lengths), but they should share the same number of features

PR #379, opened 2021-12-20
[WIP] Upgrade check_estimator tests to sklearn 1.0

GillesVandewiele:
I updated the code such that the tests can now run with sklearn 1.0.

I made the following significant changes:
1) ~~keras.api import error because of using `tf.optimizers.Adam` instead of `keras.optimizers.Adam`~~ **(EDIT: this change has been reverted as tests failed.)**
2) Different yield_check functions from sklearn 1.0 only accept an estimator now, as opposed to (name, estimator) as before.
3) Added new `_xfail_checks` to our TimeSeriesSVC and TimeSeriesSVR. This is similar as to what the devs from sklearn add to their SVC and SVR. This new flag _xfail_checks specifies which exact tests to skip for a certain class. This could perhaps be used for other tests that we skip manually atm.
4) Patching in order to call `_maybe_skip` from sklearn to skip the necessary tests based on the `_xfail_checks` flag.
5) Patched the `_regression_dataset` from sklearn to our own `_create_large_ts_dataset` such that time series data is used.


GillesVandewiele:
If we want to support Python 3.6, we will have to use an older sklearn version for the tests there... Since the code is not fully backward compatible, that would also mean that we would have to use two versions of our `test_estimators` and `sklearn_patches` files

rtavenar:
> If we want to support Python 3.6, we will have to use an older sklearn version for the tests there... Since the code is not fully backward compatible, that would also mean that we would have to use two versions of our `test_estimators` and `sklearn_patches` files

Can you confirm that the only issue with backward compatibility is in the tests, not the core modules ?

GillesVandewiele:
Hi Romain,

I will double-check on this later. But for now, I can confirm that I only had to make changes to the testing code itself, and that afterwards all tests ran successfully (so no crashes due to no backward compatibility), so it seems that all our code is backward compatible!

GillesVandewiele:
Hi Romain,

Sorry for being inactive again for a few weeks. I have some time to look at it today/tomorrow. You were correct that there is probably more to change. AFAIK, all API Changes in sklearn still ensure backward compatibility for now, but it is probably a good idea to already refactor those changes here as well. An overview is here: https://scikit-learn.org/stable/whats_new/v1.0.html#changes-1-0

I'll go over it and see which changes are relevant to tslearn and modify accordingly.

GillesVandewiele:
Just went over the list, checked all the ones that were tagged with "API Change". The only relevant one is that the `load_boston` which was internally used for sklearn's tests is now deprecated due to ethical issues, but this was already fixed by my previous patches. All other "API Change" comments are not related to tslearn. Version `1.0.1` and `1.0.2` include no api changes.

Additionally, I could check the logs of our Azure Pipeline for any possible DeprecationWarnings as a final double-check? However, I cannot access the linux builds anymore (build not found). Could these be re-triggered?

rtavenar:
Hi,

I thought there was an attempt to for all estimators to have default values for all their hyper-parameters, am I mistaken?

Also, you are right, I cannot find the Azure runs anymore, not sure why, will investigate.

GillesVandewiele:
Isn't it already the case that they have default values for all their hyper-parameters? I didn't see it explicitly stated (except for the fact that some hyper-parameters change their default value) in the patch notes of v1.0.0 and I'm also not seeing it in v1.1.0 (the next release).

GillesVandewiele:
I did a quick grep for sklearn imports in our code-base. Below is a list of all estimators we internally use and their current 1.0.2 API and the previous 0.24.0 API. (filling in this list atm with edits).

# sklearn.linear_model.LinearRegression

0.24:
```
class sklearn.linear_model.LinearRegression(*, fit_intercept=True, normalize=False, copy_X=True, n_jobs=None, positive=False)
```
1.0:
```
class sklearn.linear_model.LinearRegression(*, fit_intercept=True, normalize='deprecated', copy_X=True, n_jobs=None, positive=False)
```

--> Important: the normalize keyword becomes deprecated.

GillesVandewiele:
Hi Romain,

Any updates regarding the azure pipelines? The logs could be useful to fix potential DeprecationWarnings. Also, I think we are safe regarding the hyper-parameters & changed defaults (see my list), but perhaps I have overlooked something?

GillesVandewiele:
Closed & re-opened this PR in order to re-trigger the azure pipelines (seem to have worked...)

rtavenar:
Integrated in #406 and merged to `main` branch

PR #411, opened 2022-07-01
[MRG] Fix numpy versions problems

YannCabanes:


YannCabanes:
Hello, there seems to be a problem related to numpy versions into tslearn's main branch.
For now this branch as no difference with tslearn's main branch.
I have no error message when I run the tests on my local computer using pytest.

Part of the error is related to the lines:

import numpy as np
cimport numpy as np
np.import_array()

in the file: 
https://github.com/tslearn-team/tslearn/blob/main/tslearn/metrics/soft_dtw_fast.pyx

Here is the error message:

__init__.pxd:942: in numpy.import_array
??? 
E RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xf

This error message looks to be related to different versions of numpy being installed:
https://github.com/freqtrade/freqtrade/issues/4281

The solution of this error message looks to be to upgrade numpy:
pip install numpy --upgrade

YannCabanes:
We should use numpy version <= 1.22 as I have the following error message:
E   ImportError: Numba needs NumPy 1.22 or less

YannCabanes:
Now we have the following error message when running the tests on Linux with Python 3.7:

+ python -m pip install numpy==1.22
ERROR: Ignored the following versions that require a different python version: 1.22.0 Requires-Python >=3.8

Python 3.7 requires NumPy version <= 1.21.6

YannCabanes:
Now we have the following error message:

tslearn/metrics/cysax.pyx:1: in init tslearn.metrics.cysax
    STUFF_cysax = "cysax"
E   ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject

which also seems to be related to numpy versions.

YannCabanes:
Now there is only one test being run: docs/readthedocs.org:tslearn (successful)
The other tests have not been performed.

YannCabanes:
Hello @rtavenar and @GillesVandewiele, 
The tests are not failing on my local computer, so I am trying to find the problem (probably related to numpy versions) using directly the Continuous Intergration of tslearn.
I am trying different things, I have not been very successful for now.
Do you have any ideas about how to solve this? Any suggestions is welcome.

GillesVandewiele:
Hi @YannCabanes 

I have been doing a bit of unsuccessful digging into these issues myself. My 2 cents I can already give straight away (I will look further in depth into this later) is that these `E ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject` errors are often related to `pycocotools` according to StackOverflow. But then again I would not know immediately which of our dependencies actually use pycocotools...

YannCabanes:
Now, I have the following error message:
E   ImportError: Numba needs NumPy 1.22 or less
numba 0.55.2 requires numpy<1.23,>=1.18, but you have numpy 1.23.0 which is incompatible.

I have previously tried to write: python -m pip install numpy==1.22
but then I have the error message for Python 3.7:
Numpy 1.22 needs Python >= 3.8

NimaSarajpoor:
> tslearn/metrics/cysax.pyx:1: in init tslearn.metrics.cysax

Would you mind trying this?
```
# in tslearn/metrics/cysax.pyx

import numpy
cimport numpy
numpy.import_array() # PLEASE ADD THIS RIGHT AFTER `cimport numpy`
```

Can you also add the same thing for `cycc.pyx`? 

----

btw, `soft_dtw_fast.pyx` is good already and has this line. 

----

And, then see if this issue can be resolved or if we get a new error or not.

YannCabanes:
Thank you @NimaSarajpoor! Yes, I will try!

YannCabanes:
Hello @NimaSarajpoor, 
I have tried to add numpy.import_array() after cimport numpy as you suggested. 
However, it seems that we still have the same kind of error message.
Any suggestion is welcome, so please tell me if there is anything else that you would like to try.

YannCabanes:
@rtavenar suggested me to try to replace the use of cython by a combination of numpy and numba, so I will try this method.

YannCabanes:
A link to test the usefulness of the option "parallel=True" of the @njit decorator of numba:
https://numba.readthedocs.io/en/stable/user/parallel.html#numba-parallel-diagnostics

NimaSarajpoor:
@YannCabanes 
Do you mind if I go through the changed files and review them, and share with you my thoughts? I think it would help me improve my "review" skill in my own work. (btw, please feel free to review mine. I can learn something!)

YannCabanes:
Hello @NimaSarajpoor,
Yes, you can review my codes. Any help is welcome.
After discussion with @rtavenar, we will try to replace the cython file by python files in which we will use numba.
Now I am trying to understand in which context the different options (nopython, parallel...) of the @jit decorator of numba can be used.
Then, I will run some tests to compare the execution time of cython and numba. Maybe that you can advise me for this part too.

NimaSarajpoor:
> Yes, you can review my codes. Any help is welcome.
Cool

> we will try to replace the cython file by python files in which we will use numba.
Yeah...if we can get the same performance, it would be great.

> Now I am trying to understand in which context the different options (nopython, parallel...) of the @jit decorator of numba can be used.
> Then, I will run some tests to compare the execution time of cython and numba. Maybe that you can advise me for this part too.

I have some experience with numba but not that much. If I notice something, I will let you know for sure.

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    di = 0
+    t0 = 0
+    seg_sz = original_size // sz
+    dataset_out = np.zeros((n_ts, original_size, d))
+
+    for i in prange(n_ts):
+        for t in range(sz):
+            t0 = t * seg_sz
+            for di in range(d):
+                dataset_out[i, t0 : t0 + seg_sz, di] = dataset_paa[i, t, di]
```
NimaSarajpoor, line 28: can we replace the third for-loop with:
```
dataset_out[i, t0 : t0 + seg_sz, :] = dataset_paa[i, t, :]
```
?
Also, maybe we can use `np.repeat` to remove the second for-loop as well. (I am not sure though)

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    return dataset_out
+
+
+@njit(parallel=True)
+def cydist_sax(sax1, sax2, breakpoints, original_size):
+    assert sax1.shape[0] == sax2.shape[0] and sax1.shape[1] == sax2.shape[1]
```
NimaSarajpoor, line 34: How about `assert sax1.shape == sax2.shape`?

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    s = 0.0
+    sz = sax1.shape[0]
+    d = sax1.shape[1]
```
NimaSarajpoor, line 37: sz, d = sax.shape

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    t = 0
+    di = 0
+    for t in prange(sz):
+        for di in range(d):
+            if np.abs(sax1[t, di] - sax2[t, di]) > 1:
+                max_symbol = max(sax1[t, di], sax2[t, di])
+                min_symbol = min(sax1[t, di], sax2[t, di])
+                s += (breakpoints[max_symbol - 1] - breakpoints[min_symbol]) ** 2
+    return np.sqrt(s * float(original_size) / sz)
```
NimaSarajpoor, line 46: Maybe add one blank line just before return statement?

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    s = 0.0
+    sz = sax1.shape[0]
+    d = sax1.shape[1]
+    t = 0
+    di = 0
+    for t in prange(sz):
+        for di in range(d):
+            if np.abs(sax1[t, di] - sax2[t, di]) > 1:
+                max_symbol = max(sax1[t, di], sax2[t, di])
+                min_symbol = min(sax1[t, di], sax2[t, di])
```
NimaSarajpoor, line 44: can we replace the inner for-loop with this:

```
mask = np.abs(sax1[t] - sax2[t]) > 1
max_sym = np.maximum(sax1[t, mask], sax2[t, mask])
min_sym = np.minimum(sax1[t, mask], sax2[t, mask])
s += np.square(breakpoints[max_symbol - 1] - - breakpoints[min_symbol])
```
Or, something like that?
if numba gives error due to advance indexing of `sax1[t, mask]`, then do: `sax1[t][mask]` (same for sax2)

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    seg_sz = original_size // sz
+    dataset_out = np.zeros((n_ts, original_size, d))
+
+    for i in prange(n_ts):
+        for t in range(sz):
+            t0 = t * seg_sz
+            for di in range(d):
+                dataset_out[i, t0 : t0 + seg_sz, di] = breakpoints_middle_[
+                    dataset_sax[i, t, di]
+                ]
```
NimaSarajpoor, line 67: I think you can replace third for-loop with `dataset_out[.., ..., :] = ...`. (similar to another comment I provided for another part of code)

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    return dataset_out
+
+
+@njit(parallel=True)
+def cyslopes(dataset, t0):
+    i = 0
+    d = 0
```
NimaSarajpoor, line 74: I think the lines 74 and 75 can be removed.

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    d = sax1.shape[1] // 2
+    t = 0
+    di = 0
+    t0 = 0
+    seg_sz = original_size // sz
+    t_middle = 0.0
+    slope1 = 0.0
+    slope2 = 0.0
+    avg1 = 0.0
+    avg2 = 0.0
```
NimaSarajpoor, line 105: I think we do not need to initialize the variables (except `s`). Maybe they are there because in C, we need to initialize variables. I think you can remove them.

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+                    avg1 + slope1 * (tt - t_middle) - (avg2 + slope2 * (tt - t_middle))
+                ) ** 2
+    return np.sqrt(s)
+
+
+@njit(parallel=True)
+def inv_transform_1d_sax(
+    dataset_sax, breakpoints_avg_middle_, breakpoints_slope_middle_, original_size
+):
+
```
NimaSarajpoor, line 126: It is really hard to get some idea about functions when there is no docstring. Other functions have no docstring as well. I think it really helps reader to understand  the inputs and their type (and size?) and also the output.

I know your focus is just the code...but I think it is worth it to put some time to add docstrings.

tslearn/metrics/cycc.py
```diff
@@ -0,0 +1,88 @@
+STUFF_cycc = "cycc"
+
+import numpy as np
+from numba import njit, prange
+
+__author__ = "Romain Tavenard romain.tavenard[at]univ-rennes2.fr"
+
+DTYPE = float
```
NimaSarajpoor: I think you may want to remove `DTYPE`, and clearly shows the word "float" in the code below. Because, it is not easy to go through the code and then see `DTYPE`. Then, you need to go back and check what `DTYPE` was. I mean...this may not be necessarily cleaner.

NimaSarajpoor (commented):
@YannCabanes 
I left some comments. If you disagree on any on of them, please feel free and let me know. It is good to polish my "review" skill :)

YannCabanes:
Hello @NimaSarajpoor,
Thank you for your review!
Well, I do not have all answers to your questions for now as I only translated the cython files:
- soft_dtw_fast.pyx
- cysax.pyx
- cycc.pyx
into python files and I am not the original author of these cython files.
I will try to answer as many questions as I can.

tslearn/metrics/soft_dtw_fast.py
```diff
@@ -0,0 +1,94 @@
...
+    m = D.shape[0]
+    n = D.shape[1]
+
+    # Initialization.
+    for i in prange(m + 1):
+        R[i, 0] = DBL_MAX
+
+    for j in prange(n + 1):
+        R[0, j] = DBL_MAX
+
```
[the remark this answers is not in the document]
YannCabanes: For the matrix R, I think that it only the first row (from index 1 to n) and the first column (from index 1 to m) should be equal to DBL_MAX and the coefficient of index (0, 0) should be equal to zero. The other coefficients are computed in the loops after.
About DBL_MAX, I chose to use of np.finfo('double').max because it has a dtype:
- np.finfo('double').max.dtype returns dtype('float64')
- np.inf.dtype returns AttributeError: 'float' object has no attribute 'dtype'

tslearn/metrics/soft_dtw_fast.py
```diff
@@ -0,0 +1,94 @@
...
+    return -gamma * (np.log(tmp) + max_val)
+
+
+@njit(parallel=True)
+def _soft_dtw(D, R, gamma):
+
+    m = D.shape[0]
+    n = D.shape[1]
+
+    # Initialization.
```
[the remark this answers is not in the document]
YannCabanes: I do not know yet, I guess that if the matrix R is larger than the matrix D, the elements or R with large indices are left unchanged.

tslearn/metrics/soft_dtw_fast.py
```diff
@@ -0,0 +1,94 @@
...
+            # D is indexed starting from 0.
+            R[i, j] = D[i - 1, j - 1] + _softmin3(
+                R[i - 1, j], R[i - 1, j - 1], R[i, j - 1], gamma
+            )
+
+
+@njit(parallel=True)
+def _soft_dtw_grad(D, R, E, gamma):
+
+    # We added an extra row and an extra column on the Python side.
```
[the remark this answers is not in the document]
YannCabanes: Ok, thank you.

tslearn/metrics/soft_dtw_fast.py
```diff
@@ -0,0 +1,94 @@
...
+    # Initialization.
+    for i in prange(1, m + 1):
+        # For D, indices start from 0 throughout.
+        D[i - 1, n] = 0
+        R[i, n + 1] = -DBL_MAX
+
+    for j in prange(1, n + 1):
+        D[m, j - 1] = 0
+        R[m + 1, j] = -DBL_MAX
+
```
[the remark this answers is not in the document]
YannCabanes: Ok, thank you.

tslearn/metrics/soft_dtw_fast.py
```diff
@@ -0,0 +1,94 @@
...
+    E[m + 1, n + 1] = 1
+    R[m + 1, n + 1] = R[m, n]
+    D[m, n] = 0
+
+    # DP recursion.
+    for k in prange(n):
+        j = n - k  # ranges from n to 1
+        for p in range(m):
+            i = m - p  # ranges from m to 1
```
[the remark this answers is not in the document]
YannCabanes: Ok, thank you. It should range from m to 1 so I have used range(m, 0, -1)

tslearn/metrics/soft_dtw_fast.py
```diff
@@ -0,0 +1,94 @@
...
+            b = np.exp((R[i, j + 1] - R[i, j] - D[i - 1, j]) / gamma)
+            c = np.exp((R[i + 1, j + 1] - R[i, j] - D[i, j]) / gamma)
+            E[i, j] = E[i + 1, j] * a + E[i, j + 1] * b + E[i + 1, j + 1] * c
+
+
+@njit(parallel=True)
+def _jacobian_product_sq_euc(X, Y, E, G):
+    m = X.shape[0]
+    n = Y.shape[0]
+    d = X.shape[1]
```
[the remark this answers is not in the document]
YannCabanes: To write m, d = X.shape, I need to be sure that X is a matrix of dimension 2. Are you sure of that?
I do not know yet about the shape of G.

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+DTYPE = float
+DTYPE_INT = int
+
+
+@njit(parallel=True)
+def inv_transform_paa(dataset_paa, original_size):
+    n_ts = dataset_paa.shape[0]
+    sz = dataset_paa.shape[1]
+    d = dataset_paa.shape[2]
```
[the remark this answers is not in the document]
YannCabanes: Ok, thank you.

tslearn/metrics/cysax.py
```diff
@@ -0,0 +1,150 @@
...
+    i = 0
+    t = 0
+    di = 0
+    t0 = 0
```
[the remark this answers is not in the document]
YannCabanes: I think that we can just delete them as they were defined in cython mainly to specify their types, but their values are redefined after in the code.

tslearn/metrics/cysax.py
YannCabanes: Thank you. I have change the order of the loops, now I only have:
    for t in prange(sz):
        t0 = t * seg_sz
        dataset_out[:, t0: t0 + seg_sz, :] = dataset_paa[:, t, :]

tslearn/metrics/cysax.py
YannCabanes: Are you sure that both matrices sax1 and sax2 are of dimension 2?

tslearn/metrics/cysax.py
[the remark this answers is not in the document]
NimaSarajpoor: Right! I missed that.

tslearn/metrics/cysax.py
YannCabanes: Maybe that I can write: `assert sax1.shape[:2] == sax2.shape[:2]`
However, according to the name of the variables, the matrices sax1 and sax2 should be of dimension 2

tslearn/metrics/soft_dtw_fast.py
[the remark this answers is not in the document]
NimaSarajpoor: Ahh...right.. so I should have written R[0, ...]

tslearn/metrics/cysax.py
YannCabanes: Thank you.

tslearn/metrics/cysax.py
YannCabanes: Ok.

tslearn/metrics/soft_dtw_fast.py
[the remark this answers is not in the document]
NimaSarajpoor: No. I am not sure. So it should be X.shape[:2].

tslearn/metrics/cysax.py
NimaSarajpoor: Ahh..I just assumed :) so, maybe we can do .shape[:2] if it works(?)

tslearn/metrics/cysax.py
YannCabanes: According to the previous cython file cysax.pyx, the input breakpoints_middle_ is a  numpy ndarray of dimension 1.
The input dataset_sax is a numpy ndarray of dimension 3 of integers.
I am not sure that we can simplify this loop using `dataset_out[.., ..., :] = ...`.

tslearn/metrics/cysax.py
YannCabanes: According to the previous cython file cysax.pyx, the inputs sax1 and sax2 should be of dimension 2.

tslearn/metrics/cysax.py
YannCabanes: According to the previous cython file cysax.pyx, the input breakpoints is a numpy ndarray of dimension 1.
I do not think that we can write `breakpoints[max_symbol - 1]` or `breakpoints[min_symbol])`.

tslearn/metrics/cysax.py
YannCabanes: Thank you.

tslearn/metrics/cysax.py
YannCabanes: Thank you.

tslearn/metrics/cycc.py
YannCabanes: Ok, thank you.

tslearn/metrics/cysax.py
NimaSarajpoor: > According to the previous cython file cysax.pyx, the inputs sax1 and sax2 should be of dimension 2.

I think regardless of their `ndim`, we can replace `sax1.shape[0] == sax2.shape[0] and sax1.shape[1] == sax2.shape[1]` . with `sax1.shape[:2] = sax2.shape[:2]`. Aren't they the same? what do you think?

tslearn/metrics/cysax.py
NimaSarajpoor: this works for me:
```
breakpoints = np.arange(100)
max_symbol = np.array([2, 10, 4])
min_symbol = np.array([1, 7, 3])
print(breakpoints[max_symbol - 1])
print(breakpoints[min_symbol])

```
max_symbol is definitely integer. Otherwise, you cannot use it as index. So, `max_symbol - 1` is a mask that is used to get the corresponding element from 1D array `breakpoints`

tslearn/metrics/cysax.py
NimaSarajpoor: I think `dataset_sax[i, t, di]` is an integer value as it is used as index of `breakpoints_middle_`. Hence, we 
should be able to do:
```
dataset_out[i, t0 : t0 + seg_sz] = breakpoints_middle_[dataset_sax[i, t]]
```

So, `dataset_sax[i, t]` is an array containing integers that are being used as indices of `breakpoints_middle_`. And, then, we put that array into `dataset_out[i, t0 : t0 + seg_sz, :]`

YannCabanes:
Hello @NimaSarajpoor, I have added docstrings to the files soft_dtw_fast.py, cysax.py and cycc.py. If you have any suggestions to solve the code errors, please do not hesitate to send me your ideas.

NimaSarajpoor:
@YannCabanes 
Great effort! Sure.

YannCabanes:
At this point, all the tests are passing.
I will now try to add the jit decorators (currently commented).
Then I will compare the execution speeds of the codes with the current numba version and the previous cython version for the files cycc.py, cysax.py and soft_dtw_fast.py.

YannCabanes:
Adding the jit decorator creates errors for Python 3.7.

YannCabanes:
During last continuous integration tests, only Linux Python37 failed (AssertionError: Arrays are not equal).

YannCabanes:
Here are the execution times of the functions previously coded in Cython.
The input values of the functions are simulated using numpy.random.randn
The results are presented with the following hierarchical structure:
1) Size of the input dataset (small then large)
2) Name python file
3) Name of the function tested in the python file
4) Modes (Python, Numba py_func, Numba and Cython)


Small time series
The execution time is divided by the number of repetitions to obtain the average time of a single execution.
N_REPETITIONS = 100
N_TS = 15   (number of time series)
SZ = 14       (size of the time series)
D = 13         (dimension of the time series)


Functions of file cycc.py

TEST_NORMALIZED_CC
Function type       Execution Time 
Python                 6.966590881347656e-05
Numba py_func   6.471872329711914e-05
Numba                6.30354881286621e-05
Cython                7.087945938110352e-05

TEST_CDIST_NORMALIZED_CC
Function type        Execution Time 
Python                  0.004678127765655518
Numba py_func    0.006372392177581787
Numba                  0.02055845022201538
Cython                  0.004590597152709961

TEST_Y_SHIFTED_SBD_VEC
Function type         Execution Time 
Python                   0.0007832765579223633
Numba py_func     0.001002175807952881
Numba                  0.0073595881462097164
Cython                  0.000695946216583252


Functions of file cysax.py

TEST_INV_TRANSFORM_PAA
Function type        Execution Time 
Python                  0.0001658177375793457
Numba py_func    0.00012455224990844726
Numba                  2.3276805877685546e-05
Cython                  0.00072235107421875

TEST_CYDIST_SAX
Function type         Execution Time 
Python                   0.0006324005126953125
Numba py_func     0.0006050252914428711
Numba                   5.793571472167969e-06
Cython                   0.0003102374076843262

TEST_INV_TRANSFORM_SAX
Function type          Execution Time 
Python                    0.0012830519676208497
Numba py_func      0.001281435489654541
Numba                    3.258943557739258e-05
Cython                    0.0007562804222106934

TEST_CYSLOPES
Function type         Execution Time 
Python                   0.03299321174621582
Numba py_func     0.03369884967803955
Numba                   0.04607600450515747
Cython                   0.03369905710220337

TEST_CYDIST_1D_SAX
Function type         Execution Time 
Python                   0.0009629201889038086
Numba py_func     0.0009295821189880371
Numba                   5.824565887451172e-06
Cython                   0.0003400826454162598

TEST_INV_TRANSFORM_1D_SAX
Function type         Execution Time 
Python                   0.010548267364501953
Numba py_func     0.009688065052032471
Numba                  1.7206668853759766e-05
Cython                   0.004239740371704101


Functions of file soft_dtw_fast.py

TEST_SOFTMIN3
Function type          Execution Time 
Python                    3.7169456481933593e-06
Numba py_func      3.7360191345214844e-06
Numba                    5.7220458984375e-07

TEST_SOFT_DTW
Function type         Execution Time 
Python                   0.0008032011985778808
Numba py_func     0.00016197919845581054
Numba                   1.3413429260253906e-05
Cython                   6.75201416015625e-06

TEST_SOFT_DTW_GRAD
Function type         Execution Time 
Python                   0.000658884048461914
Numba py_func     0.0006336688995361329
Numba                   4.945039749145508e-05
Cython                   6.663799285888672e-06

TEST_JACOBIAN_PRODUCT_SQ_EUC
Function type            Execution Time 
Python                      0.0015450143814086915
Numba py_func        0.0015096139907836913
Numba                      6.406307220458985e-06
Cython                      4.334449768066406e-06



Large time series
The execution time is divided by the number of repetitions to obtain the average time of a single execution.
N_REPETITIONS = 10 
N_TS = 150
SZ = 140
D = 130


Functions of file cycc.py

TEST_NORMALIZED_CC
Function type         Execution Time 
Python                   0.002593326568603516
Numba py_func     0.0023230791091918947
Numba                   0.001892685890197754
Cython                   0.0039809226989746095

TEST_CDIST_NORMALIZED_CC
Function type           Execution Time 
Python                     14.68750901222229
Numba py_func       24.232205820083617
Numba                     7.441932797431946
Cython                     15.007141089439392

TEST_Y_SHIFTED_SBD_VEC
Function type         Execution Time 
Python                   0.24766547679901124
Numba py_func     0.38209333419799807
Numba                   0.11369473934173584
Cython                   0.24488503932952882


Functions of file cysax.py

TEST_INV_TRANSFORM_PAA
Function type         Execution Time 
Python                   0.07514638900756836
Numba py_func     0.07673845291137696
Numba                   0.043023204803466795
Cython                   0.8301484823226929

TEST_CYDIST_SAX
Function type          Execution Time 
Python                    0.07396893501281739
Numba py_func      0.07371225357055664
Numba                    3.3354759216308595e-05
Cython                    0.029455232620239257

TEST_INV_TRANSFORM_SAX
Function type          Execution Time 
Python                    1.3193160057067872
Numba py_func      1.3029633522033692
Numba                    0.04689924716949463
Cython                    0.7892702579498291

TEST_CYSLOPES
Function type           Execution Time 
Python                     3.478916120529175
Numba py_func       3.4544804096221924
Numba                     4.405482006072998
Cython                     3.4761438608169555

TEST_CYDIST_1D_SAX
Function type          Execution Time 
Python                    0.09309067726135253
Numba py_func      0.0933596134185791
Numba                    4.3773651123046874e-05
Cython                    0.03114762306213379

TEST_INV_TRANSFORM_1D_SAX
Function type           Execution Time 
Python                     9.720394968986511
Numba py_func       9.639565801620483
Numba                     0.024070429801940917
Cython                     4.553128099441528


Functions of file soft_dtw_fast.py

TEST_SOFTMIN3
Function type         Execution Time 
Python                   4.100799560546875e-06
Numba py_func     3.838539123535157e-06
Numba                   6.67572021484375e-07

TEST_SOFT_DTW
Function type           Execution Time 
Python                     0.07861130237579346
Numba py_func       0.016379952430725098
Numba                     0.0007523536682128906
Cython                     0.0007349014282226563

TEST_SOFT_DTW_GRAD
Function type           Execution Time 
Python                     0.07179102897644044
Numba py_func       0.06839404106140137
Numba                     0.0008146047592163086
Cython                     0.000745081901550293

TEST_JACOBIAN_PRODUCT_SQ_EUC
Function type          Execution Time 
Python                    1.5018571853637694
Numba py_func      1.496809482574463
Numba                    0.0005124330520629883
Cython                    0.0023012399673461915

tslearn/metrics/softdtw_variants.py
```diff
@@ -308,12 +310,14 @@ def gamma_soft_dtw(dataset, n_samples=100, random_state=None):
     ----------
     .. [1] M. Cuturi, "Fast global alignment kernels," ICML 2011.
     """
-    return 2. * sigma_gak(dataset=dataset,
```
rtavenar, line 311: Here I find the original version easier to read...

tslearn/metrics/cysax.py
rtavenar: @YannCabanes have you tried this?

I suspect the last line should be 
```python
s += np.sum(np.square(breakpoints[max_symbol - 1] - breakpoints[min_symbol]))
```

but I might be wrong

tslearn/metrics/cysax.py
rtavenar: @YannCabanes have you tried this?

YannCabanes:
I have removed the signatures (input and output types) in jit decorators in the last commit.
It does not affect the execution time and will give more flexibility to the jit decorated functions with respect to the input types.
I have written the expected input and output types in the documentation of each function decorated with @jit.

Information about signatures in jit decorators can be found at:
https://numba.readthedocs.io/en/stable/reference/types.html#numba-types
or at:
https://numba.readthedocs.io/en/stable/reference/jit-compilation.html
It is said in the latest that the jit decorator has several modes of operation:

    If one or more signatures are given in signature, a specialization is compiled for each of them. Calling the decorated function will then try to choose the best matching signature, and raise a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) if no appropriate conversion is available for the function arguments. If converting succeeds, the compiled machine code is executed with the converted arguments and the return value is converted back according to the signature.

    If no signature is given, the decorated function implements lazy compilation. Each call to the decorated function will try to re-use an existing specialization if it exists (for example, a call with two integer arguments may re-use a specialization for argument types (numba.int64, numba.int64)). If no suitable specialization exists, a new specialization is compiled on-the-fly, stored for later use, and executed with the converted arguments.

YannCabanes:
The last continuous integration test gave the following error message:

=================================== FAILURES ===================================
___________ test_all_estimators[LearningShapelets-LearningShapelets] ___________

name = 'LearningShapelets'
Estimator = <class 'tslearn.shapelets.shapelets.LearningShapelets'>

    @pytest.mark.parametrize('name, Estimator', get_estimators('all'))
    def test_all_estimators(name, Estimator):
        """Test all the estimators in tslearn."""
        allow_nan = (hasattr(checks, 'ALLOW_NAN') and
                     Estimator().get_tags()["allow_nan"])
        if allow_nan:
            checks.ALLOW_NAN.append(name)
        if name in ["GlobalAlignmentKernelKMeans", "ShapeletModel",
                    "SerializableShapeletModel"]:
            # Deprecated models
            return
>       check_estimator(Estimator)

tslearn/tests/test_estimators.py:215: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tslearn/tests/test_estimators.py:197: in check_estimator
    check(estimator)
/opt/hostedtoolcache/Python/3.9.14/x64/lib/python3.9/site-packages/sklearn/utils/_testing.py:311: in wrapper
    return fn(*args, **kwargs)
tslearn/tests/sklearn_patches.py:558: in check_pipeline_consistency
    assert_allclose_dense_sparse(result, result_pipe)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = array([[3.7043095e-03],
       [6.7453969e-01],
       [6.3824987e-01],
       [1.2295246e-03],
       [2.0980835e-05]...4e-03],
       [8.6247969e-01],
       [1.4195442e-03],
       [5.0067902e-06],
       [9.4977307e-01]], dtype=float32)
y = array([[0.40121353],
       [0.06187719],
       [0.05123574],
       [0.21641088],
       [0.2602595 ],
       [0.076...      [0.25475943],
       [0.12683961],
       [0.27159142],
       [0.29283226],
       [0.16161257]], dtype=float32)
rtol = 1e-07, atol = 1e-09, err_msg = ''

    def assert_allclose_dense_sparse(x, y, rtol=1e-07, atol=1e-9, err_msg=""):
        """Assert allclose for sparse and dense data.
    
        Both x and y need to be either sparse or dense, they
        can't be mixed.
    
        Parameters
        ----------
        x : {array-like, sparse matrix}
            First array to compare.
    
        y : {array-like, sparse matrix}
            Second array to compare.
    
        rtol : float, default=1e-07
            relative tolerance; see numpy.allclose.
    
        atol : float, default=1e-9
            absolute tolerance; see numpy.allclose. Note that the default here is
            more tolerant than the default for numpy.testing.assert_allclose, where
            atol=0.
    
        err_msg : str, default=''
            Error message to raise.
        """
        if sp.sparse.issparse(x) and sp.sparse.issparse(y):
            x = x.tocsr()
            y = y.tocsr()
            x.sum_duplicates()
            y.sum_duplicates()
            assert_array_equal(x.indices, y.indices, err_msg=err_msg)
            assert_array_equal(x.indptr, y.indptr, err_msg=err_msg)
            assert_allclose(x.data, y.data, rtol=rtol, atol=atol, err_msg=err_msg)
        elif not sp.sparse.issparse(x) and not sp.sparse.issparse(y):
            # both dense
>           assert_allclose(x, y, rtol=rtol, atol=atol, err_msg=err_msg)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=1e-09
E           
E           Mismatched elements: 30 / 30 (100%)
E           Max absolute difference: 0.7881605
E           Max relative difference: 23.541649
E            x: array([[3.704309e-03],
E                  [6.745397e-01],
E                  [6.382499e-01],...
E            y: array([[0.401214],
E                  [0.061877],
E                  [0.051236],...

/opt/hostedtoolcache/Python/3.9.14/x64/lib/python3.9/site-packages/sklearn/utils/_testing.py:418: AssertionError

tslearn/metrics/softdtw_variants.py
YannCabanes: Hello @rtavenar, 
Are you talking about the code aesthetics of the file softdtw_variants.py?
I think that I only did a few useful modifications in this file, most of the "aesthetics" modifications have been performed by the "black" command. The command: "black + filename" in the terminal automatically corrects Python codes aesthetics to make sure that it corresponds to PEP8 conventions. It is very practical to make sure that the PEP8 conventions are respected, but sometimes it might not be the "aesthetics" version prefered by the users. 
Would you like make to come back to the previous presentation?
And more generally, want do you think of using the "black" command? 
Personally, I find "black" very convenient, but I wouldn't want it to come at the expense of aesthetics or code clarity.

tslearn/metrics/cysax.py
YannCabanes: I have tried to replace the last loop by:

`mask = np.abs(sax1[t] - sax2[t]) > 1
max_sym = np.maximum(sax1[t, mask], sax2[t, mask])
min_sym = np.minimum(sax1[t, mask], sax2[t, mask])
s += np.sum(np.square(breakpoints[max_symbol - 1] - breakpoints[min_symbol]))`

It works with standard Python, but does not work with Numba.
I have the following error message related to `sax1[t, mask]` and `sax1[t, mask]`:
NumbaNotImplementedError: only one advanced index supported

I think that the same problem is discussed here:
https://stackoverflow.com/questions/70957881/numbanotimplementederror-only-one-advanced-index-supported-how-to-rewrite-a
The solution proposed is to use a loop instead of array indexing.

YannCabanes:
There is still the same error message.

YannCabanes:
The tests are failing with Linux, they pass with MacOS and Windows.
The tests pass on my local computer, I am using Linux and Python 3.8.

YannCabanes:
There is still the same failing tests with the signatures in the jit decorators.
I will remove the signatures once again for more flexibility.

YannCabanes:
Hello @rtavenar, 

I have resolved all conversations except for one conversation with you about codes esthetics. 
Could you read my answer and give me your opinion?

Also, could you look at the failing tests and give me your opinion please?
The tests are failing with Linux or they pass with Windows and MacOS.
I use Linux and Python 3.8, and the tests pass on my local computer.
Do you think that the failing test is related to the current PR 411?

Once these questions are answered, I think that we will be able to merge this PR.

tslearn/metrics/softdtw_variants.py
rtavenar: I just think that, in this case, black made things worse instead of better-looking.

rtavenar:
> Hello @rtavenar,
> 
> I have resolved all conversations except for one conversation with you about codes esthetics. Could you read my answer and give me your opinion?
> 
> Also, could you look at the failing tests and give me your opinion please? The tests are failing with Linux or they pass with Windows and MacOS. I use Linux and Python 3.8, and the tests pass on my local computer. Do you think that the failing test is related to the current PR 411?
> 
> Once these questions are answered, I think that we will be able to merge this PR.

Hi,

It is unclear to me why some tests fail on Linux. These tests are checking that LearningShapelets model behave the same when used in a pipeline or alone, and to do so a seed is applied. In the model's fit, this seed is used to set the seed for both numpy and TF. I will try to find out why, but if you have any idea, it would be helpful.

tslearn/metrics/softdtw_variants.py
YannCabanes: Ok, I will restore the previous formulation.

YannCabanes:
On my local computer, the following message is related to the tests of the class `tslearn.shapelets.shapelets.LearningShapelets` in functions: `test_all_estimators` --> `check_estimator` --> `check_pipeline_consistency`:

`2022-09-30 16:03:48.019023: E tensorflow/stream_executor/cuda/cuda_driver.cc:271] failed call to cuInit: CUDA_ERROR_NO_DEVICE: no CUDA-capable device is detected
2022-09-30 16:03:48.019106: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:156] kernel driver does not appear to be running on this host (psylia): /proc/driver/nvidia/version does not exist
2022-09-30 16:03:48.019572: I tensorflow/core/platform/cpu_feature_guard.cc:151] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.`

YannCabanes:
It seems that this error message in not related to the Python files cycc.py, cysax.py, soft_dtw.py, soft_dtw_fast.py and softdtw_variants.py which are the only Python files related to this PR.
I have used the "print" function in each function of these files, and no function was called during the tests of the class tslearn.shapelets.shapelets.LearningShapelets by functions: `test_all_estimators` --> `check_estimator` --> `check_pipeline_consistency`

PR #416, opened 2022-07-25
[WIP] Experiment multiple backends

YannCabanes:


YannCabanes:
Hello @rtavenar, I have started a PR about multiple backends.
To begin with, we only focus on numpy and pytorch.
I have created a new file named tslearn/experiment_multiple_backends.py in which I have created several backend classes and a test classifier.
The objective for now is the create a backend structure and a classifier structure such that the classifiers (and other machine learning algorithms) of tslearn will adapt to the backend of the data provided by the user.
Once we have a better idea of the structure of what we want for the backends' structure, I plan to split the experiment_multiple_backends.py file into several python files for each backend and a test file.

PR #388, opened 2022-02-20
[WIP] Add calculation of the SBD metric #276

swifmaneum:
As discussed in this [issue ](https://github.com/tslearn-team/tslearn/issues/276): Adding calculation of the SBD metric according to the [original paper](http://web2.cs.columbia.edu/~gravano/Papers/2015/sigmod2015.pdf) and [this repo](https://github.com/johnpaparrizos/kshape).

Issue #385, opened 2022-02-15
KShape clustering loses first time point


uksa10:
**Describe the bug**
In order to visualize it better, I added 4 to the first time point. But I don't see any spike in the first index of the cluster centers array.
The whole array seems to be shifted 1 to the left
Additionally the last index of the cluster center array is filled up with a random number. 
![sinuscos](https://user-images.githubusercontent.com/56071069/154038276-108c4834-41ce-417e-800e-8fdd7102946d.png)
Sometimes there is a Spike in one of the clusters, but  then the entire array gets shifted.

![sinus](https://user-images.githubusercontent.com/56071069/154038692-62cfe2dc-9d60-4868-aeec-631568594955.png)

Is this expected behavior or a bug?

In other tests
**To Reproduce**
```
import numpy as np
import matplotlib.pyplot as plt
from tslearn.clustering import KShape
from tslearn.preprocessing import TimeSeriesScalerMeanVariance
x = np.arange(0, 10, 0.1)
sincos=np.zeros((20, 100), dtype=float)

for i in range (0,10):
 sincos[i] = np.sin(x)+np.random.normal(scale=0.1, size=len(x))
 sincos[i+10] = np.cos(x)+np.random.normal(scale=0.1, size=len(x))
 sincos[i,0]=sincos[i,0]+4
 sincos[10+i,0]=sincos[10+i,0]+4
sincos = TimeSeriesScalerMeanVariance().fit_transform(sincos)
ks = KShape(n_clusters=2,verbose=True)
y_pred = ks.fit_predict(sincos)
plt.figure(figsize=(13,5))
for i in range(2):
    plt.subplot(2, 1, 1 + i)
    for xx in sincos[y_pred == i]:
        plt.plot(xx[0:100].ravel(), "k-" ,alpha=.2)
    plt.plot(ks.cluster_centers_[i,0:100].ravel(), "r-")
    plt.xlim(0, 100)
    plt.ylim(-2, 4)
    plt.title(f"Cluster {str(i + 1)} has {str(sincos[y_pred == i].shape[0])} timeseries")
plt.tight_layout()
plt.show()
```

**Environment (please complete the following information):**
 - tslearn version 0.5.2

**Additional context**
Maybe it is because of the shifting operations in the KShape algorithm.

PR #417, opened 2022-07-26
[Bugfix] Fixes an off-by-one error in kshape. (Issue #385)

temcomp:
Problem pointed out in [issue #385](https://github.com/tslearn-team/tslearn/issues/385). The original [paper](https://dl.acm.org/doi/pdf/10.1145/2723372.2737793) uses MATLAB-style indexing, i.e. starting on 1, as opposed to Python-style indexing starting on 0. Therefore there was a mismatch between tslearn and line 6, Algorithm 1 of that paper.

It should be noted that this indexing issue seems to be accounted for in [this PR](https://github.com/tslearn-team/tslearn/pull/388).


charavelg:
I've ported it to current codebase in #596... Hope you don't mind and thanks for the fix!

PR #428, opened 2022-10-11
Fix url bibtex

yasirroni:
The old link in `http://jmlr.org/papers/v21/20-091.html` seems to be dead and moved to `https://www.jmlr.org/papers/v21/20-091.html`

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ If you use `tslearn` in a scientific publication, we would appreciate citations:
   volume  = {21},
   number  = {118},
   pages   = {1-6},
-  url     = {http://jmlr.org/papers/v21/20-091.html}
+  url     = {https://www.jmlr.org/papers/v21/20-091.html}
 }
 ```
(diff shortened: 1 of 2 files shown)


rtavenar:
On my side, both links work...

yasirroni:
Ah, I see.

I was previously tested the link in brave. Tested in chrome, the older link works.

I don't think brave strictly only allow HTTP because AFAIK, it will only warn user when going to HTTP.
_
Tested in brave, it return

```plaintext
Check your Internet connection
Check any cables and reboot any routers, modems or other network devices you may be using.
```
_
I don't know the cause. If you accept this PR, than it will support both browser. If this PR is rejected, I'm okay with it since the old link works on certain browser.

yasirroni:
I test brave to open HTTP (http://example.com/) and it works.

Then, it might not be because brave didn't support HTTP, but the old link might has other issue.

PR #452, opened 2023-05-27
Added method to compute dtw distance using custom metric without the path

sw241395:
Added in a method too compute the DTW distance between 2 time series using a custom metric but not generate the path. 

Overall computing DTW can be quite a slow process, so i though if you was only interested in the distance then you could help speed up the code buy not creating the path. I have tried to keep it consistent with the rest of the code but happy to made stylistic changes if needed. 

Overall still new to contributing to open source so happy to make any changes and feedback is always appreciated.


rtavenar:
Hi @sw241395 

Thanks for this suggestion. You are perfectly right that this should be offered as a direct function to the end-user.

Here is my quick feedback:
1. This looks great (almost ready for integration)
2. I would rather call the function `dtw_from_metric`

sw241395:
Hello @rtavenar 

I have made the change. Hope it all looks ok. Happy to make other changes if needed.

PR #453, opened 2023-05-31
Update locate returned data type from np.int to np.int_

fspinna:
This commit updates the data type from np.int to np.int_ in locate method of LearningShapelets. The change ensures compatibility with the NumPy library given that using the aliases of builtin types like np.int is deprecated.


YannCabanes:
Hello @fspinna, thank you for your PR!
Indeed, `np.int` is deprecated since NumPy 1.20.0:
https://numpy.org/devdocs/release/1.20.0-notes.html
It is replaced by `np.int_` or `int`.

PR #460, opened 2023-07-03
List of improvements expected for tslearn version 0.6.2

YannCabanes:
For the version 0.6.2 of tslearn, the following improvements are expected:
- [ ] Fix the test error with MacOS for Python 3.9 (`Segmentation fault`):
    * `test_symmetric_cdist` --> The failing rate of this test is about 3/4
    * `test_gamma_soft_dtw` --> The failing rate of this test is about 1/5
When these tests are failing, the error messages can be different from one fail to the other: the Segmentation fault can occur in different lines of the code. These two tests are currently skipped with MacOS for Python 3.9.
- [x] In the tests, do we have to specify the backend if our data is prepared using `be`?
We should test both cases: with backend inferred from data and with backend explicitly specified, even with data from another format.
     --> This problem has been dealt with in PR #479.
- [x] The documentation about the backends should be improved to explain how they work.
For example what does it mean for DTW to support the `PyTorch` backend, do we differentiate at fixed path?
     --> This problem has been dealt with in PR #482.
- [x] The documentation of the class `SoftDTWLossPyTorch` should be improved. 
      --> The documentation of the class `SoftDTWLossPyTorch` has been improved in the PR #467.
- [x] For the class `Backend` in `backend/backend.py`, should we overload `__getattr__` instead of listing all attributes? 
     --> This requirement has been dealt with in PR #471.


rtavenar:
If I may add something: at the moment, the tests install scikit-learn v1.0, which does not provide wheels for Python 3.10. As a result, for all tests running on Python 3.10, a large part of the test time is dedicated to compiling scikit-learn. It would be a good idea, if feasible, to switch to scikit-learn 1.3 for the tests.

YannCabanes:
The documentation of the class `SoftDTWLossPyTorch` has been improved in the PR #467.

rtavenar:
> If I may add something: at the moment, the tests install scikit-learn v1.0, which does not provide wheels for Python 3.10. As a result, for all tests running on Python 3.10, a large part of the test time is dedicated to compiling scikit-learn. It would be a good idea, if feasible, to switch to scikit-learn 1.3 for the tests.

We are not there (=1.3) yet, however we upgraded tests to support scikit-learn 1.2, which provides wheels for Python 3.8 to 3.11 it seems

charavelg:
0.6.4 is out, closing

PR #463, opened 2023-07-04
[MRG] Fix import error when torch is not installed.

eicca:
Init file in `tslearn/backend` imports torch backend, which imports torch. This causes an import error when torch is not installed. This change fixes this issue by removing extra import in `tslearn/backend/__init__.py`

Additionally, this change fixes failing tests when torch is not installed.

Fixes #462

I tried to run the tests locally, but some of them fail with and without pytorch being installed, so I can't say for sure if I fixed all test cases.

When this is fixed, I think some users will appreciate a new release, since the latest version is breaking for everybody who didn't install pytorch. To make things slightly worse, the latest version of pytorch is incompatible with the latest version of tensorflow: https://github.com/pytorch/pytorch/issues/97580, therefore the temporary fix of installing pytorch won't work in some cases.

PR #464, opened 2023-07-05
Fix import error when torch is not installed (bis).

YannCabanes:
This PR is related to the PR #463 (it has the same objective).


YannCabanes:
Only the `docs/readthedocs.org:tslearn` test is running.

YannCabanes:
Only the `docs/readthedocs.org:tslearn` test is running.

PR #467, opened 2023-07-13
Improve the docstring of the class SoftDTWLossPyTorch and rename the parameter dist_func into squared_dist_func

YannCabanes:
Improve the docstring of the class `SoftDTWLossPyTorch` and rename the parameter `dist_func` into `squared_dist_func`.

### DIFF
--- a/tslearn/metrics/soft_dtw_loss_pytorch.py
+++ b/tslearn/metrics/soft_dtw_loss_pytorch.py
@@ -73,32 +73,120 @@ else:
             return grad_output.view(-1, 1, 1).expand_as(E) * E, None, None
 
     class SoftDTWLossPyTorch(torch.nn.Module):
-        """Class for the Soft-DTW loss function in PyTorch.
+        r"""Define the Soft-DTW loss function in PyTorch.
+
+        Soft-DTW was originally presented in [1]_ and is
+        discussed in more details in our
+        :ref:`user-guide page on DTW and its variants<dtw>`.
+
+        Soft-DTW is computed as:
+
+        .. math::
+
+            \text{soft-DTW}_{\gamma}(X, Y) =
+                \min_{\pi}{}^\gamma \sum_{(i, j) \in \pi} d^{2} \left( X_i, Y_j \right)
+
+        where :math:`d^{2}` is a squared distance function supporting PyTorch automatic differentiation
+        and :math:`\min^\gamma` is the soft-min operator of parameter :math:`\gamma` defined as:
+
+        .. math::
+
+            \min^\gamma \left( a_{1}, ..., a_{n} \right) =
+                - \gamma \log \sum_{i=1}^{n} e^{- a_{i} / \gamma}
+
+        In the limit case :math:`\gamma = 0`, :math:`\min^\gamma` reduces to a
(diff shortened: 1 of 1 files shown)


YannCabanes:
Hello @rtavenar, could you review this PR please?
The failing tests are not related to this PR, I am trying to solve these issues in the PR #468.

tslearn/metrics/soft_dtw_loss_pytorch.py
```diff
@@ -73,32 +73,117 @@ def backward(ctx, grad_output):
             return grad_output.view(-1, 1, 1).expand_as(E) * E, None, None
 
     class SoftDTWLossPyTorch(torch.nn.Module):
-        """Class for the Soft-DTW loss function in PyTorch.
+        r"""Define the Soft-DTW loss function in PyTorch.
```
rtavenar, line 76: I would remove the "Define the"

tslearn/metrics/soft_dtw_loss_pytorch.py
```diff
@@ -73,32 +73,117 @@ def backward(ctx, grad_output):
...
             Regularization parameter.
+            It should be strictly positive.
             Lower is less smoothed (closer to true DTW).
         normalize : bool
             If True, the Soft-DTW divergence is used.
             The Soft-DTW divergence is always positive.
             Optional, default: False.
-        dist_func : callable
-            The distance function.
+        squared_dist_func : callable
```
rtavenar, line 129: I prefer the previous version for these 2 lines.

rtavenar:
This looks great. I only have two comments and we should be able to merge soon.

tslearn/metrics/soft_dtw_loss_pytorch.py
YannCabanes: Done.

tslearn/metrics/soft_dtw_loss_pytorch.py
YannCabanes: I am afraid that the variable name `dist_func` could be confusing for the users.
By default, when `dist_func=None`, `dist_func` is defined as the squared Euclidean distance (`_euclidean_squared_dist`) which is not a distance function, that's why I wanted to rename `dist_func` into `squared_dist_func`.
Moreover we request this callable function to be differentiable whereas distance functions have singular values in which they are not differentiable.
Do you still want to keep `dist_func` for the parameter name @rtavenar ?
If not `squared_dist_func`, could we use another parameter name (like `dissimilarity`)?

rtavenar: I understand your reasons but I still think it should not be called squared... Since one can pick another similarity measure (and having pointwise singularities is not an issue in fact).
Another option would be to call it metric, but since we have released a version in which we called it dist_fun I would rather not break backward compatibility for an issue that seems minor to me. You can maybe edit the docs so that they say "Distance function (or dissimilarity)"

YannCabanes: Ok, done.

YannCabanes:
All tests using Python 3.10 have failed.

YannCabanes:
All tests using Python 3.10 have failed.

PR #468, opened 2023-07-13
Fix deprecated np.bool, np.int and np.float

YannCabanes:
The continuous integration test `docs/readthedocs.org:tslearn` is failing due to `np.bool` being a deprecated alias for the builtin `bool` in NumPy 1.20.

Here is one of the error messages:

```
Extension error:
Here is a summary of the problems encountered when running the examples

Unexpected failing examples:
/home/docs/checkouts/readthedocs.org/user_builds/tslearn/checkouts/467/docs/examples/misc/plot_distance_and_matrix_profile.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/tslearn/checkouts/467/docs/examples/misc/plot_distance_and_matrix_profile.py", line 47, in <module>
    mp = matrix_profiler.fit_transform([ts]).flatten()
  File "/home/docs/checkouts/readthedocs.org/user_builds/tslearn/envs/467/lib/python3.8/site-packages/sklearn/utils/_set_output.py", line 140, in wrapped
    data_to_wrap = f(self, X, *args, **kwargs)
  File "/home/docs/checkouts/readthedocs.org/user_builds/tslearn/envs/467/lib/python3.8/site-packages/tslearn-0.6.1-py3.8.egg/tslearn/matrix_profile/matrix_profile.py", line 250, in fit_transform
    return self._fit(X)._transform(X)
  File "/home/docs/checkouts/readthedocs.org/user_builds/tslearn/envs/467/lib/python3.8/site-packages/tslearn-0.6.1-py3.8.egg/tslearn/matrix_profile/matrix_profile.py", line 195, in _transform
    band_width, dtype=np.bool) &
  File "/home/docs/checkouts/readthedocs.org/user_builds/tslearn/envs/467/lib/python3.8/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

PR #469, opened 2023-07-17
Fix errors deprecated np.bool, np.int and np.float and array of inhomogeneous shape

YannCabanes:
This PR has the same objective than the PR #468: to fix errors related to deprecated `np.bool`, `np.int` and `np.float` and arrays of inhomogeneous shape.
The package `black` has been used in the PR #468 on the modified files to reformat the codes, which makes the PR difficult to review.
This PR aims to be cleaner, avoiding the use of the package black on the modified files.

### DIFF
--- a/tslearn/barycenters/softdtw.py
+++ b/tslearn/barycenters/softdtw.py
@@ -100,7 +100,7 @@ def softdtw_barycenter(X, gamma=1.0, weights=None, method="L-BFGS-B", tol=1e-3,
         barycenter = init
 
     if max_iter > 0:
-        X_ = numpy.array([to_time_series(d, remove_nans=True) for d in X_])
+        X_ = [to_time_series(d, remove_nans=True) for d in X_]
 
         def f(Z):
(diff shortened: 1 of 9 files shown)


tslearn/barycenters/softdtw.py
```diff
@@ -100,7 +100,7 @@ def softdtw_barycenter(X, gamma=1.0, weights=None, method="L-BFGS-B", tol=1e-3,
         barycenter = init
 
     if max_iter > 0:
-        X_ = numpy.array([to_time_series(d, remove_nans=True) for d in X_])
+        X_ = [to_time_series(d, remove_nans=True) for d in X_]
```
rtavenar, line 103: Here, do we have a test with time series of unequal length somewhere, to make sure that this formulation actually works?

rtavenar: OK, so the doctest actually checks this, so if all tests pass, that means this implementation is OK

PR #471, opened 2023-07-24
Simplify backend

YannCabanes:
Simplify the way the attributes and methods are defined in the class `Backend`.


YannCabanes:
All checks have passed.

PR #476, opened 2023-08-21
[WIP] Fix cdist_soft_dtw_normalized when two different datasets are provided

### LINKED ISSUE #475: [BUG] `cdist_soft_dtw_normalized` fails unexpectedly when time series panels have different number of instances
**Describe the bug**
`cdist_soft_dtw_normalized` fails with an exception when time series panels with different number of instances are provided to it.

This behaviour is different from `cdist_soft_dtw` where the same code does not fail, so I assume it is a but rather than an interface limitation.

**To Reproduce**

```python
from tslearn.metrics import cdist_soft_dtw_normalized

import numpy as np

arr = np.array([[1, 2], [2, 3], [3, 4]])

X = [arr, arr]
X2 = [arr, arr, arr]

cdist_soft_dtw_normalized(X, X2)
```

**Expected behavior**
The above should return a 2 x 3 `np.array` of `float`-s, just as the same call with `cdist_soft_dtw`

**Environment (please complete the following information):**
 - OS: windows, python 3.11
 - tslearn version 0.6.1

**Additional context**
Surfaced through systematic testing in a draft for an `sktime` adapter here: https://github.com/sktime/sktime/pull/5039

rtavenar:
Fixes #475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,14 @@ Changelogs for this project are recorded in this file since v0.2.0.
 
 
 
+## [Towards v0.7]
+
 ## [Towards v0.6.2]
+
+### Fixed
+
+* Fixed an incorrect calculation of the normalization term for `cdist_soft_dtw_normalized` when `dataset2` is provided.
+* Fixed UCR/UEA datasets download link
 
 ## [v0.6.1]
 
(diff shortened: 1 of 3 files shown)


rtavenar:
OK, so the fix seems to be efficient, yet all tests are failing at the moment due to a problem with UEA/UCR datasets.

Seems the form of the links there have changed. For example, the link to the Adiac dataset is expected to be:

https://www.timeseriesclassification.com/ClassificationDownloads/Adiac.zip

But it is now:

https://timeseriesclassification.com/aeon-toolkit/Adiac.zip

@TonyBagnall can you confirm? Can we expect that the new form for the links will be somewhat stable?

TonyBagnall:
hi @rtavenar yes sorry about that, it will now be stable

rtavenar:
OK, great, then I'll make the change in this PR. Thanks @TonyBagnall for the prompt reply!

Issue #473, opened 2023-08-01
Compute SoftDTWLossPyTorch with normalization option and time series of different lengths


denisbeslic:
**Describe the bug**
Hello, I want to use the SoftDTWLoss with normalization (Soft-DTW divergence) for comparing predicted time series of different length. It seems to throw this error.
```
  File ".../lib/python3.10/site-packages/tslearn/metrics/soft_dtw_loss_pytorch.py", line 146, in forward
    xxy = torch.cat([x, x, y])

RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 99 but got size 100 for tensor number 2 in the list.
```
Is it possible to generate a SoftDTWLoss for time series with unequal lengths or is this a bug?

**To Reproduce**
```
targets = torch.tensor(np.random.randn(16, 100, 1))
prediction = torch.tensor(np.random.randn(16, 99, 1))
loss_func = SoftDTWLossPyTorch(gamma=1.0, normalize=True, dist_func=None)
loss = loss_func(prediction, targets)
print(loss.shape)
print(loss)
```

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04.5 LTS
 - tslearn version 0.6.1

**Additional context**
-

YannCabanes:
Hello @denisbeslic, thanks a lot for reporting this bug!
I am dealing with this issue in PR https://github.com/tslearn-team/tslearn/pull/478.

PR #478, opened 2023-09-07
 Fix bug related to SoftDTWLossPytorch with option normalize=True when used on inputs of different lengths

YannCabanes:
This PR is related to the issue https://github.com/tslearn-team/tslearn/issues/473 reporting an error message when `SoftDTWLossPytorch`  with option `normalize=True` is used on inputs of different lengths.
The bug is coming from the use of `torch.cat` (https://pytorch.org/docs/stable/generated/torch.cat.html) which impose the lengths of the inputs to be equal. (The shape of the inputs is `[batch_size, length, dim]`.)
The use of `torch.cat` is removed in this PR.

### DIFF
--- a/tslearn/metrics/soft_dtw_loss_pytorch.py
+++ b/tslearn/metrics/soft_dtw_loss_pytorch.py
@@ -230,13 +230,13 @@ else:
             by, ly, dy = y.shape
             assert bx == by
             assert dx == dy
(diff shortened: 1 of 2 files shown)


YannCabanes:
All checks have passed.

YannCabanes:
All checks have passed.

PR #487, opened 2023-10-16
Fix readthedocs fail test changing build.image into build.os

YannCabanes:
The `readthedocs` test is failing on the main branch with the following error message:
`The configuration key "build.image" is deprecated. Use "build.os" instead to continue building your project. Read more at https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os`

Indeed, `build.image` is deprecated since Monday, October 16, 2023 and is now replaced by `build.os`:
https://blog.readthedocs.com/use-build-os-config/

Specific details to write the `.readthedocs.yml` file can be found at:
https://docs.readthedocs.io/en/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,13 @@
 # Required
 version: 2
 
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-latest
+  tools:
+    python:
+      version: "3.8"
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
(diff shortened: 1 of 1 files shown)


YannCabanes:
We have the following error related to the Ubuntu version:
`Problem in your project's configuration. Invalid configuration option "build.os": expected one of (ubuntu-20.04, ubuntu-22.04), got ubuntu-latest`

Here is a discussion about `ubuntu-latest`:
https://github.com/readthedocs/readthedocs.org/issues/8861

Issue #509, opened 2024-02-28
lcss similarity is returns unity for all timeseries


NAThompson:
**Describe the bug**

In the [documentation](https://tslearn.readthedocs.io/en/stable/user_guide/lcss.html#lcss) of LCSS, it says that the pseudo-metric obeys the property ∀x LCSS(x, x) = 0.

However, I have found an x such that LCSS(x, x) = 1.0. In addition, I have built a family of randomly perturbed waveforms such that they all have the same score.

**To Reproduce**

```python
from math import pi as π
import random
import numpy

def test_reproduce():
    f0 = 20e9
    period = 1/f0
    ω0 = 2*π*f0
    waveforms = numpy.empty(shape=(101, 512))
    times = numpy.linspace(-period/2, period/2, waveforms.shape[1])
    for i in range(waveforms.shape[0]):
        φ = random.gauss(0.0, 0.5)
        k = 3 + random.uniform(-0.5, 0.5)
        values = 0.5*(numpy.tanh(k*ω0*times + φ) + 1) + random.uniform(-0.05, 0.05)
        waveforms[i, :] = values

    typical_values = 0.5*(numpy.tanh(3*ω0*times) + 1)
    scores = numpy.empty(waveforms.shape[0])
    for i in range(waveforms.shape[0]):
        scores[i] = lcss(typical_values, waveforms[i, :])

    # Prints an array of 1s:
    print(scores)
    # prints 1:
    print(lcss(waveforms[0, :], waveforms[0, :]))
```

The output is:

```
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1.]
1.0
```

**Expected behavior**

I expect a randomly perturbed family of waveforms to all have different scores, and the property LCSS(x,x) = 0 to hold.

[...]

PR #513, opened 2024-03-19
Fix LCSS documentation

YannCabanes:
Fix a few mistakes in the LCSS documentation:
- Fix a typo in `epsilon` and `delta` mathematical symbols
- Fix an error in `epsilon` and `delta` data types
- In the pseudocode of the function `lcss`, normalize the returned result by dividing by `min(n, m)` 
- Precise that the values returned by LCSS range from 0 to 1, the value 1 being taken when the two time series completely match. It was written that for all `x`, `lcss(x, x) = 0` instead of `lcss(x, x) = 1`. This mistake has been noticed in Issue https://github.com/tslearn-team/tslearn/issues/509
- Remove a wrong sentence saying that the clustering algorithm `TimeSeriesKMeans` can be used with the `lcss` metric

### DIFF
--- a/docs/user_guide/lcss.rst
+++ b/docs/user_guide/lcss.rst
@@ -28,8 +28,8 @@ This is the algorithm at stake when invoking
 Problem
 --------------------
 
-The similarity S between :math:`x` and :math:`y`, given an integer :math:`\epsilon` and
-a real number :math:`\delta`, is formulated as follows:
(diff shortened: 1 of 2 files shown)


rtavenar:
Thanks @YannCabanes !

It seems the $\delta$ parameter is not implemented in practice (cf docs for the LCSS function) since it plays the same role as the Sakoe-Chiba band. Could you remove all references to this $\delta$ parameter in the docs page you edited?

rtavenar:
LGTM!

PR #521, opened 2024-07-01
Update tslearn's codes to be compatible with the version 0.16.0 and above of the package sphinx-gallery

YannCabanes:
Update `tslearn`'s codes to be compatible with the version 0.16.0 and above of the package `sphinx-gallery`.


charavelg:
Close after #556

PR #561, opened 2025-07-22
Example amendment

AnnaBobasheva:
Merging main from origimal repository


charavelg:
Thanks for contributing!!!

Unfortunately, it seems that UCR UEA datasets are not currently available, hence the failing CI. Not sure why though.

AnnaBobasheva:
Maybe the  CI cannot access the outside sources?

PR #593, opened 2025-09-10
Random init and nan in subset of dimensions

michelegentili93:
The main problems that I found are:

1) when setting init='random', pick a random set of ts as cluster centroids; it may take the same ts multiple times (because  in np.random.choice,  replace is set to True). Therefore the "duplicated" clusters stays empty, no centroid is computed and then you get an error.
2) When computing the distance between two ts with multiple dimensions, if there is only a nan in one or more of the dimensions (but not all), the distance will be inf. It expects that if there are nans should be on all the dimensions observed in the time series. EX: a ts with 3 timepoints and 2 dimensions =[[1,2,3],[2,3,na]] is not supported. 
3) In the Barycenter computation again it is assumed that either a ts has all nans in all dimensions. 
These 3 problems should be addressed:
when init is random, we should make sure that each cluster is initialized on a different ts. Done
when we have a nan in a subset of the dimension of the ts, don't use the ts only for that dimension for the computation of the distance and the barycenter. Done

### DIFF
--- a/tslearn/barycenters/dba.py
+++ b/tslearn/barycenters/dba.py
@@ -341,7 +341,7 @@ def _mm_update_barycenter(X, diag_sum_v_k, list_w_k):
     sum_w_x = numpy.zeros((barycenter_size, d))
     for k, (w_k, x_k) in enumerate(zip(list_w_k, X)):
         # because w_k is the matching bertween ts timestamps and centroid timestamps, and is the same across dimension, we gotta discount the diag_sum_v_k (which is how many timestamp from the ts have been assigned to that centroid timestamp) of the timestamps that have a nan in that specific dimension.
-        x_k_na0 = np.nan_to_num(x_k, nan=0)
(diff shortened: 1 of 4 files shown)


charavelg:
Samples with nans on all dimensions are skipped only when they are trailing samples because that's the way tslearn represents variable length dataset.

Dropping nan dimensions in the local_squared_distance computation will affect all metrics using it as an inner distance. Although it might be fitting your needs, in the particular case of DTW with KMeans, I don't think that's what we want as the default behavior in tslearn. We should instead probably make it clear that missing values (as opposed to empty trailing samples) are not supported in this particular setup.

You might want to look at [the imputer preprocessor](https://tslearn.readthedocs.io/en/latest/gen_modules/preprocessing/tslearn.preprocessing.TimeSeriesImputer.html#tslearn.preprocessing.TimeSeriesImputer) to deal with missing values.

PR #595, opened 2025-09-16
Add frechet metric

charavelg:


tslearn/neighbors/neighbors.py
```diff
@@ -1,16 +1,28 @@
...
+)
+from tslearn.piecewise import SymbolicAggregateApproximation
+from tslearn.utils import (
+    to_time_series_dataset,
+    to_sklearn_dataset,
+    check_array,
+    check_dims
+)
 
 neighbors.VALID_METRICS['brute'].extend(['dtw', 'softdtw', 'sax', 'ctw'])
```
rtavenar, line 27: Don't we need to add Fréchet here too? Sorry I don't remember why it was important to populate this `VALID_METRICS['brute']` stuff :/

tslearn/neighbors/neighbors.py
```diff
@@ -60,17 +72,30 @@ def _precompute_cross_dist(self, X, other_X=None):
         X = to_time_series_dataset(X)
 
         if self._ts_metric == "dtw":
```
rtavenar, line 74: Here, it feels weird that we have different calls depending on the metrics, whereas it should be very similar for all, something around the lines of:

```python
cdist_XXX(
  X,
  other_X,
  n_jobs=self.n_jobs,
  **metric_params
)
```

if we had something more homogeneous like that, we could just pick the `cdist_XXX` based on a dict mapping distance names to functions and there we go, or is it a bad idea?

tslearn/metrics/frechet.py
```diff
@@ -0,0 +1,685 @@
...
+        See Also
+        --------
+        frechet_path : Get both the matching path and the similarity score
+        frechet_path_from_metric : Compute similarity score and path
+            using a user-defined distance metric
+        cdist_frechet : Cross similarity matrix between time series datasets
+
+        References
+        ----------
+        .. [1] H. Sakoe, S. Chiba, "Dynamic programming algorithm optimization for
```
rtavenar, line 129: Update this to the correct paper for Fréchet distance.

I guess the paper is this one:

FRÉCHET, M. (1906): Sur quelques points du calcul fonctionnel. Rendiconti del Circolo Mathematico di Palermo, 22, 1–74

tslearn/metrics/frechet.py
```diff
@@ -0,0 +1,685 @@
...
+           spoken word recognition," IEEE Transactions on Acoustics, Speech and
+           Signal Processing, vol. 26(1), pp. 43--49, 1978.
+
+    """  # noqa: E501
+    return _frechet(
```
rtavenar, line 134: One slight problem with this is that we compute the path even if the used does not want it, which results in degraded performance. Shouldn't we avoid it?

tslearn/neighbors/neighbors.py
charavelg: Currently, softdtw does not accept the n_jobs parameter.

tslearn/metrics/frechet.py
charavelg: Done

PR #632, opened 2026-01-20
Fix deprecation warning in piecewise module

felixdivo:
This warning occurred 72 times in the test suite:

```
  tslearn/piecewise/piecewise.py:685: DeprecationWarning: Conversion of an array with ndim > 0
  to a scalar is deprecated, and will error in future. Ensure you extract a single element from your 
  array before performing this operation. (Deprecated NumPy 1.25.)
  
    X_slopes[:, i_seg, :] = cyslopes(X[:, start:end, :], start)
```


felixdivo:
Related to #625 (but fixes different warnings)

PR #679, opened 2026-06-09
ENH allow changing data root

tomMoral:
Proposal for dataset management:
- Allow changing the root_dir on `UCR_UEA_datasets`
- Move from `os.path` to `pathlib`
- Respect the `XDG_DATA_HOME` convention for cache directory (I could also remove this if you are not convinced by it but many projects follow it :)

Also add some dedicated dataset test (not only doctests).

Happy to change anything that you don't like, the most interesting change for me is the `root_dir` override to allow caching at different locations depending on the needs.

### DIFF
--- a/tslearn/datasets/cached.py
+++ b/tslearn/datasets/cached.py
@@ -20,7 +20,7 @@ class CachedDatasets:
        Series Classification Repository, www.timeseriesclassification.com
     """
     def __init__(self):
-        # Local test datseta are stored in tslearn/.cached_datasets
+        # Local test datasets are stored in tslearn/.cached_datasets
         self.path = Path(__file__).parents[1] / ".cached_datasets"
 
     def list_datasets(self):
--- a/tslearn/datasets/cached.py
+++ b/tslearn/datasets/cached.py
@@ -1,5 +1,6 @@
 import numpy
-import os
+from pathlib import Path
+
 
 class CachedDatasets:
     """A convenience class to access cached time series datasets.
@@ -19,18 +20,16 @@ class CachedDatasets:
        Series Classification Repository, www.timeseriesclassification.com
     """
     def __init__(self):
-        self.path = os.path.join(os.path.dirname(__file__),
-                                 "..",
-                                 ".cached_datasets")
+        # Local test datseta are stored in tslearn/.cached_datasets
+        self.path = Path(__file__).parents[1] / ".cached_datasets"
 
     def list_datasets(self):
         """List cached datasets.
 
         Examples
         --------
-        >>> from tslearn.datasets import UCR_UEA_datasets
-        >>> _ = UCR_UEA_datasets().load_dataset("Trace")
-        >>> cached = UCR_UEA_datasets().list_cached_datasets()
+        >>> from tslearn.datasets import CachedDatasets
+        >>> cached = CachedDatasets().list_datasets()
         >>> "Trace" in cached
         True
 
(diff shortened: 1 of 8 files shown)


tslearn/datasets/cached.py
```diff
@@ -19,30 +20,27 @@ class CachedDatasets:
        Series Classification Repository, www.timeseriesclassification.com
     """
     def __init__(self):
-        self.path = os.path.join(os.path.dirname(__file__),
-                                 "..",
-                                 ".cached_datasets")
+        # Local test datseta are stored in tslearn/.cached_datasets
```
charavelg: dataset*

felixdivo (approved):

charavelg:
Hi and thanks for the PR !! 

Regarding tests, we want to decouple datasets tests that fetch data from remote location so that the unavailability of the remote infrastructure does not block the test suite. You might want to have a look at [conftest](https://github.com/tslearn-team/tslearn/blob/main/conftest.py)  and [dedicated github action](https://github.com/tslearn-team/tslearn/blob/main/.github/workflows/test_UCR_UEA_datasets.yml). I'd recommend you add non-CachedDatasets tests to the list of tests potentially skipped and make sure they are routed to the dedicated github action.

That being said, both  `root_dir` selection and `XDG_DATA_HOME` support seems like valuable addition. And pathlib is for sure relevant here !!!  Thanks for the effort!

I'd be happy to help with the changes if needed.

charavelg (commented):
Typo

tslearn/datasets/cached.py
felixdivo:
```suggestion
        # Local test datasets are stored in tslearn/.cached_datasets
```

tomMoral:
Thanks for the quick feedback.
Just added a fixture to skip tests when UCR is not available.
Something that could help with availability: set `XDG_DATA_HOME` in the CI to a `./data` folder and use https://github.com/actions/cache to cache it, so you download the UCR data only once in a while to avoid rate limits and use cache version.
I can set this up if you want (I have done it for multiple repo).

tomMoral:
One thing I did not do: document the change.
Not sure where/how it should be documented, happy to take pointers here.

tslearn/datasets/ucr_uea.py
```diff
@@ -259,25 +288,25 @@ def load_dataset(self, dataset_name):
         (7494, 8, 2)
         >>> assert (None, None, None, None) == data_loader.load_dataset(
         ...         "DatasetThatDoesNotExist")
+        >>>
```
charavelg: Needed?

charavelg:
The changelog is indeed the right place to document the change. 

Regarding the fixture to skip tests when UCR is not available, my personal preference would be to have a unified way of handling this. I believe the fixture is computed for every test it is used in. Do you have strong feelings against adding the tests in the skipped list c.f. [conftest](https://github.com/tomMoral/tslearn/blob/30a164a00111adcffd7f6e2f271daaca7536cdc0/conftest.py#L96)?

Not sure about the `set XDG_DATA_HOME in the CI to a ./data`, but it probably belongs to an other PR anyway.

tomMoral:
The scope makes the fixture computed only once for the entire session.

It is the recommended way to skip tests in the pytest doc and I prefer not to touch to pytest internals. 
Also personally I dislike list patterns like this as it is easy to make typo/dont realise you include smthg that does not exist.

That being said, I agree that having only one way to skip is better and for doctest, I don't know how to do fixture (never looked into it).

Happy to follow along the way you prefer in tslearn

charavelg:
You made your point. Let's go with the fixture.

Issue #702, opened 2026-08-18
silhouette_samples: precomputed distance matrix silently re-interpreted as raw data


SomSamantray:
## Problem
In the new `tslearn.clustering.silhouette_samples`, a square 2D input passed with the default metric (`None`, i.e. dtw) is silently treated as n time series of length n, feeding each row to DTW and producing plausible-but-wrong silhouettes. Users migrating from sklearn's `silhouette_samples` (which defaults to `metric="euclidean"` on 2D data) are the likely victims.

## Suggested fix
When `metric` is `None` (or `"dtw"`) and `X.ndim == 2` with `X.shape[0] == X.shape[1]`, emit a warning that X looks like a precomputed distance matrix and `metric="precomputed"` should be used.

## Source
- Branch: `feat/silhouette-samples` (head `1ed546c`)
- File: `tslearn/clustering/utils.py` (line ~335)

## Metadata
Severity: P3
Confidence: 100
Reviewer(s): adversarial
Finding ID: utils.py-335-precomputed-distance-matrix-passed-without-metric-precomputed-is-silently-re-interpreted-as-raw-time-series-data

SomSamantray:
This finding is carried by PR #703 (feat/silhouette-samples).

charavelg:
Issue on an open PR, invalid. Probably better fix the PR then.

Issue #701, opened 2026-08-18
silhouette_samples: sample_size/random_state guard is bypassable via metric_params


SomSamantray:
## Problem
The rejection guard in the new `tslearn.clustering.silhouette_samples` only checks `kwds` for `sample_size`/`random_state`. A user can smuggle these keys through `metric_params`, which is merged into `metric_params_` and silently ignored by the precomputed/euclidean/callable branches, producing full-length results instead of the documented `TypeError`.

## Suggested fix
After the `metric_params_` merge, apply the same rejection check on `metric_params_` keys (`sample_size`, `random_state`) with the same clear `TypeError`, making the documented rejection contract uniform across all six branches.

## Source
- Branch: `feat/silhouette-samples` (head `1ed546c`)
- File: `tslearn/clustering/utils.py` (line ~317)

## Metadata
Severity: P3
Confidence: 100
Reviewer(s): adversarial
Finding ID: utils.py-317-sample-size-random-state-rejection-guard-is-bypassable-via-metric-params

SomSamantray:
This finding is carried by PR #703 (feat/silhouette-samples).

charavelg:
Issue on an open PR, invalid. Probably better fix the PR then.

PR #703, opened 2026-08-18
[MRG] Add silhouette_samples to tslearn.clustering

### LINKED ISSUE #451: Support for  silhouette_samples() functionalities like sklearn.
Is there any support for silhouette_samples() functionalities as sklearn has.

SomSamantray:
## Summary

Time-series clustering users can now get per-sample silhouette coefficients with time-series metrics. Previously only the scalar `silhouette_score` was available; per-sample values (needed for silhouette plots and identifying misassigned or outlier series) required a manual two-step workaround of computing a cross-distance matrix and calling sklearn directly.

`tslearn.clustering.silhouette_samples()` accepts time-series data or a precomputed distance matrix, computes the cross-distance matrix with the requested metric (`dtw`, `softdtw`, `euclidean`, a callable, or `"precomputed"`), and delegates the per-sample computation to sklearn's `silhouette_samples` with `metric="precomputed"` — mirroring the existing `silhouette_score` wrapper. The function deliberately omits `sample_size`/`random_state` (sklearn's `silhouette_samples` has no subsampling), rejecting non-`None` values with a `TypeError`.

## What changed

- New `tslearn.clustering.silhouette_samples()` in `tslearn/clustering/utils.py`, exported from `tslearn.clustering`.
- Per-sample values are `(n_ts,)` and satisfy the documented invariant `mean(silhouette_samples(...)) == silhouette_score(...)` for the same inputs.
- Callable metrics receive `metric_params`/`**kwds` and return backend-normalized values, so a callable behaves like the equivalent string metric under both numpy and pytorch backends.
- Docstring with doctests, API reference entry, CHANGELOG entry, and a silhouette-plot example (`docs/examples/clustering/plot_silhouette.py`).

## Testing

- `test_silhouette_samples()` covers shape, value range, mean-vs-score equality per metric, `TypeError` rejection of `sample_size`/`random_state` across all metric branches, `None` no-op semantics, and `metric_params` forwarding to callable metrics.
- `pytest tests/test_clustering.py --doctest-modules tslearn/clustering/utils.py`: 11 passed.
- Example verified headless (matplotlib Agg): mean silhouette 0.71 on the Trace dataset.

Related: #451

### DIFF
--- a/docs/examples/clustering/plot_silhouette.py
+++ b/docs/examples/clustering/plot_silhouette.py
@@ -1,0 +1,77 @@
+# -*- coding: utf-8 -*-
+"""
+Silhouette analysis
+===================
+
+This example computes, for each time series of a dataset clustered with
+:math:`k`-means, its individual silhouette coefficient [1]_. While the mean
+silhouette coefficient gives a single scalar summarizing cluster quality, the
+per-sample values reveal *which* time series are well-embedded in their
+cluster and which ones are likely misassigned or outliers.
+
+In the figure below, each bar corresponds to one time series, colored by its
+cluster membership. Bars close to +1 indicate time series that are far from
+neighboring clusters, bars near 0 indicate time series close to the decision
+boundary between two clusters, and negative bars indicate time series that
+may have been assigned to the wrong cluster.
+
+.. [1] Peter J. Rousseeuw. "Silhouettes: a Graphical Aid to the
+  Interpretation and Validation of Cluster Analysis". Computational and
+  Applied Mathematics 20: 53-65, 1987.
+"""
+
+# Author: Romain Tavenard
+# License: BSD 3 clause
+
+import numpy
+import matplotlib.pyplot as plt
+
+from tslearn.clustering import TimeSeriesKMeans, silhouette_score, \
+    silhouette_samples
+from tslearn.datasets import CachedDatasets
+from tslearn.preprocessing import TimeSeriesScalerMeanVariance, \
+    TimeSeriesResampler
+
+seed = 0
+numpy.random.seed(seed)
+X_train, y_train, X_test, y_test = CachedDatasets().load_dataset("Trace")
+X_train = X_train[y_train < 4]  # Keep first 3 classes
+numpy.random.shuffle(X_train)
+# Keep only 50 time series
+X_train = TimeSeriesScalerMeanVariance().fit_transform(X_train[:50])
+# Make time series shorter
+X_train = TimeSeriesResampler(sz=40).fit_transform(X_train)
+
+# k-means clustering with DTW
+km = TimeSeriesKMeans(n_clusters=3, metric="dtw", verbose=True,
+                      random_state=seed)
+y_pred = km.fit_predict(X_train)
+
+# Per-sample and mean silhouette coefficients
+samples = silhouette_samples(X_train, y_pred, metric="dtw")
+score = silhouette_score(X_train, y_pred, metric="dtw")
+
+print("Mean silhouette score: {:.4f}".format(score))
+
+plt.figure()
+y_lower = 10
+for yi in range(3):
+    cluster_samples = samples[y_pred == yi]
+    cluster_samples.sort()
+    y_upper = y_lower + cluster_samples.shape[0]
+    plt.fill_betweenx(numpy.arange(y_lower, y_upper),
+                      0., cluster_samples,
+                      alpha=.7,
+                      label="Cluster %d" % (yi + 1))
+    plt.text(-0.1, y_lower + 0.5 * cluster_samples.shape[0],
+             "Cluster %d" % (yi + 1))
+    y_lower = y_upper + 10
+
+plt.axvline(x=score, color="red", linestyle="--",
+            label="Mean silhouette score")
+plt.xlabel("Silhouette coefficient values")
+plt.ylabel("Time series (ordered by cluster)")
+plt.title("Silhouette plot for $k$-means (DTW) on the Trace dataset")
+plt.legend(loc="best")
+plt.tight_layout()
+plt.show()
--- a/docs/examples/clustering/plot_silhouette.py
+++ b/docs/examples/clustering/plot_silhouette.py
@@ -43,8 +43,7 @@ X_train = TimeSeriesScalerMeanVariance().fit_transform(X_train[:50])
 X_train = TimeSeriesResampler(sz=40).fit_transform(X_train)
 
 # k-means clustering with DTW
-km = TimeSeriesKMeans(n_clusters=3, metric="dtw", verbose=True,
-                      random_state=seed)
+km = TimeSeriesKMeans(n_clusters=3, metric="dtw", random_state=seed)
 y_pred = km.fit_predict(X_train)
 
 # Per-sample and mean silhouette coefficients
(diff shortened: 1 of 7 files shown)


tslearn/clustering/utils.py
```diff
@@ -215,6 +217,162 @@ def sklearn_metric(x, y):
...
+        sklearn_X = _cdist_dtw(
+            X,
+            n_jobs=n_jobs,
+            verbose=verbose,
+            be=be,
+            **metric_params_
+        )
+    elif metric == "softdtw":
+        X = to_time_series_dataset(X, be=be)
+        sklearn_X = _cdist_soft_dtw_normalized(X, be=be, **metric_params_)
```
charavelg: `_cdist_soft_dtw_normalized` now have `n_jobs` and `verbose` args (It should also be updated in silhouette_score but that's another matter).
```suggestion
        sklearn_X = _cdist_soft_dtw_normalized(X, be=be, n_jobs=n_jobs, verbose=verbose, **metric_params_)
```

tslearn/clustering/utils.py
```diff
@@ -215,6 +217,162 @@ def sklearn_metric(x, y):
...
+    verbose : int (default: 0)
+        If nonzero, joblib progress messages are printed during the
+        cross-distance matrix computation (only used by the dtw branch).
+
+    **kwds : optional keyword parameters
+        Any further parameters are passed directly to the distance function,
+        just as for the `metric_params` parameter.
+        Note that, unlike :func:`tslearn.clustering.silhouette_score`,
+        ``sample_size`` and ``random_state`` are not supported (they are
+        rejected with a ``TypeError`` if provided).
```
charavelg: Why is it necessary to reject `sample_size` and `random_state` specifically? Aren't 'kwds' and 'metric_params' arguments redundant? I understand this mimics silhouette_score definition but this seems irrelevant.

tslearn/clustering/utils.py
```diff
@@ -215,6 +217,162 @@ def sklearn_metric(x, y):
...
+    """
+    if kwds.get("sample_size") is not None or \
+            kwds.get("random_state") is not None:
+        raise TypeError(
+            "silhouette_samples does not support the 'sample_size' or "
+            "'random_state' parameters (see sklearn.metrics.silhouette_samples)"
+        )
+    # Explicit None for these keys is a no-op, matching sklearn semantics
+    kwds = {k: v for k, v in kwds.items()
+            if k not in ("sample_size", "random_state")}
```
charavelg: Same as above

docs/residual-review-findings/feat-silhouette-samples.md
charavelg, line 1: Should not be commited

docs/examples/clustering/plot_silhouette.py
```diff
@@ -0,0 +1,77 @@
...
+neighboring clusters, bars near 0 indicate time series close to the decision
+boundary between two clusters, and negative bars indicate time series that
+may have been assigned to the wrong cluster.
+
+.. [1] Peter J. Rousseeuw. "Silhouettes: a Graphical Aid to the
+  Interpretation and Validation of Cluster Analysis". Computational and
+  Applied Mathematics 20: 53-65, 1987.
+"""
+
+# Author: Romain Tavenard
```
charavelg: Probably not the right author :)

docs/examples/clustering/plot_silhouette.py
```diff
@@ -0,0 +1,77 @@
...
+X_train = X_train[y_train < 4]  # Keep first 3 classes
+numpy.random.shuffle(X_train)
+# Keep only 50 time series
+X_train = TimeSeriesScalerMeanVariance().fit_transform(X_train[:50])
+# Make time series shorter
+X_train = TimeSeriesResampler(sz=40).fit_transform(X_train)
+
+# k-means clustering with DTW
+km = TimeSeriesKMeans(n_clusters=3, metric="dtw", verbose=True,
+                      random_state=seed)
```
charavelg:
```suggestion
km = TimeSeriesKMeans(n_clusters=3, metric="dtw", random_state=seed)
```
See https://tslearn--703.org.readthedocs.build/en/703/auto_examples/clustering/plot_silhouette.html for generated documentation

charavelg:
Hi @SomSamantray , thanks for the PR, that would be indeed a nice addition to tslearn!!!

Would you mind  checking the review so we can make your PR cross the finish line ? I'm of course available to help you in any way if needed.

SomSamantray:
Hi @charavelg — thanks for the "Cleaner figure" fix, pulled it in and it looks good.

Checked the coverage concern as well: the Codecov patch-coverage gap (2 uncovered lines in tslearn/clustering/utils.py) was closed in e32c94f before this review — current checks show codecov/patch at 100.00% of diff hit and codecov/project at 95.20% (+0.05%), no annotations. All CI is green and the two residual P3 items are already tracked separately as #701 and #702 (non-blocking).

This should be good to merge whenever you get a chance — let me know if you'd like anything else addressed first!

charavelg:
Really sorry I started the review but forgot to submit it ... You should now be able to see it...

tslearn/clustering/utils.py
SomSamantray: Good question — the rejection guard exists because `silhouette_score` accepts `sample_size`/`random_state`, so users naturally try to pass them to the per-sample variant by muscle memory. sklearn's per-sample `silhouette_samples` has no subsampling at all, so a silent no-op would quietly produce results from a subset (or, worse, mislead users into thinking subsampling applied). Rejecting with a `TypeError` makes the unsupported parameter loud.

On `kwds` vs `metric_params`: they are intentionally redundant. `**kwds` mirrors `silhouette_score`'s signature so callers can pass metric kwargs positionally (`silhouette_samples(X, y, metric="dtw", global_constraint="sakoe_chiba")`), while `metric_params` is the canonical dict form. Both feed the same merged dict.

One caveat I'll note: the guard is bypassable by passing `sample_size` inside `metric_params` (it's forwarded to the metric as a kwarg). That's tracked as #701.

tslearn/clustering/utils.py
SomSamantray: Same rationale as the other thread: the `TypeError` guard makes the unsupported `sample_size`/`random_state` (which `silhouette_score` accepts but sklearn's per-sample variant does not support) fail loudly instead of silently no-op'ing. `kwds` and `metric_params` are intentionally redundant convenience channels into the same merged dict.

tslearn/clustering/utils.py
SomSamantray: > `_cdist_soft_dtw_normalized` now have `n_jobs` and `verbose` args

Fixed — the softdtw branch now threads `n_jobs` and `verbose` through to `cdist_soft_dtw_normalized` (the public wrapper, which forwards them), in commits cb36ac1 and 7046fdd. The `verbose` docstring now reflects that it applies to the dtw and softdtw branches.

docs/residual-review-findings/feat-silhouette-samples.md
SomSamantray: > Should not be commited

Fixed — the internal `docs/residual-review-findings/feat-silhouette-samples.md` artifact was removed from the branch in commit 90c6797 and no longer appears in the PR diff.

tslearn/clustering/utils.py
charavelg: On kwds vs metric_params, I do agree with your analysis and we have the same implementation on the tslearn silhouette_score, so let's go that way!

Regarding explicit rejection guard, I don't see this kind of checks within sklearn implementation for example which has the same silhouette_score / silhouette_samples couple. Plus, removing it it would make #701 irrelevant.  

>  a silent no-op would quietly produce results from a subset
Could you explain ? What's the code path that would lead to such a result?

Definitely in favor of removing the explicit guard part so far

SomSamantray: On the removal: you're right, once I actually tried to construct the silent-no-op scenario I couldn't find one that the guard was protecting against — for dtw/softdtw/callable the value gets forwarded to the underlying cdist call regardless (erroring naturally if unsupported), and for euclidean/precomputed metric_params was never even consulted. So the guard's only power was rejecting *itself*, which isn't a real protection. Removed in `14f930a`, along with the corresponding TypeError tests in test_clustering.py updated to reflect the new (sklearn-matching) behavior.

charavelg: nice!

tslearn/clustering/utils.py
charavelg: Gone with guard removal

charavelg:
@SomSamantray, thanks for your time and effort making this PR go through!!! Nice addition to tslearn code base!!!

Issue #445, opened 2023-03-26
Difference in results between tslearn and DBA.py implementation


filehelp:
I've been trying to use the average DBA function provided by tslearn to compute the average of a set of time series. However, I've noticed that the results obtained with tslearn's implementation and with the original DBA.py module by Francois Petitjean (https://github.com/fpetitjean/DBA/blob/master/DBA.py) seem to differ. I've attached graphs that illustrate the discrepancies. I've used the exact same data for both implementations.

I'm unsure whether this is due to a bug in the tslearn code or if I might have missed some modification that's required for the tslearn implementation. Could someone help me investigate this issue, please?
Thank you.
![image](https://user-images.githubusercontent.com/128979961/227788734-0dfe3f2f-d830-425e-b522-9c6ff6bf3a74.png)

PR #704, opened 2026-08-20
[MRG] Document DBA initialization differences

kiwoongyoon:
Title: [MRG] Document DBA initialization differences

## Summary

- Document that `dtw_barycenter_averaging` starts from the pointwise mean by
  default.
- Clarify that Petitjean's standalone `DBA.py` starts from an approximate
  medoid and that comparisons should provide the same `init_barycenter`.

The issue snapshot is unassigned and shows no related PR or duplicate label.

## Test evidence

- `python3 -c 'import ast, pathlib; tree = ast.parse(pathlib.Path("tslearn/barycenters/dba.py").read_text()); doc = next(ast.get_docstring(node) for node in tree.body if isinstance(node, ast.FunctionDef) and node.name == "dtw_barycenter_averaging"); doc = " ".join(doc.split()); assert "starts from an approximate medoid" in doc and "set ``init_barycenter`` explicitly" in doc'`
- `git diff --check`

Both commands passed. Runtime doctests were not available in the Builder
environment because pytest and joblib are not installed.

## Risks or notes for maintainers

The reporter did not provide the requested reproduction data, so this patch
does not claim that initialization is the only source of the plotted
difference. It changes documentation only and intentionally leaves #445 open
for a concrete reproduction.

Refs #445

### DIFF
--- a/tslearn/barycenters/dba.py
+++ b/tslearn/barycenters/dba.py
@@ -500,6 +500,9 @@ def dtw_barycenter_averaging(
     DBA was originally presented in [1]_.
     This implementation is based on an idea from [2]_ (Majorize-Minimize Mean
     Algorithm).
+    By default, the initial barycenter is the pointwise mean of ``X``.
+    Petitjean's standalone ``DBA.py`` instead starts from an approximate
+    medoid, so set ``init_barycenter`` explicitly when comparing results.
 
     Parameters
     ----------


charavelg:
Hi @kiwoongyoon and thanks for the interest! 
I don' think including reference to outside code in docstrings specific to comparison result is the right think to do though. Yet the analysis  about potential root cause of #445 is relevant. Would you mind commenting on #445 with your findings? 

Moreover, explanations on default None argument for `init_barycenter` is indeed lacking and taking care of in #706

Issue #526, opened 2024-10-14
Why doesn't the tslearn lcss mention the delta parameter?


camilo-unipd:
Although the delta parameter from the original LCSS paper (Discovering Similar Multidimensional Trajectories) is mentioned in the [user guide page](https://tslearn.readthedocs.io/en/stable/user_guide/lcss.html#lcss) and in the [function documentation's body](url), it is not being considered as an actual function parameter. I have been looking in al the tslearn resources I found, but I could not find an explicit explanation for this.

Also, I was checking the source code (a snippet is shown below) and the delta parameter does not appear either (in contrast to the algorithm shown in the user guide). What am I missing in this case? Is this controlled by the other parameters?

```
def lcss_accumulated_matrix(s1, s2, eps, mask, be=None):
    """Compute the longest common subsequence similarity score between
    two time series.

    Parameters
    ----------
    s1 : array-like, shape=(sz1, d) or (sz1,)
        First time series. If shape is (sz1,), the time series is assumed to be univariate.
    s2 : array-like, shape=(sz2, d) or (sz2,)
        Second time series. If shape is (sz2,), the time series is assumed to be univariate.
    eps : float
        Matching threshold.
    mask : array-like, shape=(sz1, sz2)
        Mask. Unconsidered cells must have infinite values.
    be : Backend object or string or None
        Backend. If `be` is an instance of the class `NumPyBackend` or the string `"numpy"`,
        the NumPy backend is used.
        If `be` is an instance of the class `PyTorchBackend` or the string `"pytorch"`,
        the PyTorch backend is used.
        If `be` is `None`, the backend is determined by the input arrays.
        See our :ref:`dedicated user-guide page <backend>` for more information.

    Returns
    -------
    acc_cost_mat : array-like, shape=(sz1 + 1, sz2 + 1)
        Accumulated cost matrix.
    """
    be = instantiate_backend(be, s1, s2)
    s1 = be.array(s1)
    s2 = be.array(s2)
    s1 = to_time_series(s1, remove_nans=True, be=be)
    s2 = to_time_series(s2, remove_nans=True, be=be)
    l1 = be.shape(s1)[0]
    l2 = be.shape(s2)[0]
    acc_cost_mat = be.full((l1 + 1, l2 + 1), 0)

    for i in range(1, l1 + 1):
        for j in range(1, l2 + 1):
            if be.isfinite(mask[i - 1, j - 1]):
                if be.is_numpy:
[...]

PR #705, opened 2026-08-21
Fix LCSS ignoring the global_constraint / Sakoe-Chiba band

AlejandroCoronadoN:
`lcss` and `lcss_path_from_metric` accept `global_constraint` / `sakoe_chiba_radius` / `itakura_max_slope`, but the constraint is silently ignored: the numba LCSS score functions test the band with `numpy.isfinite(mask[i - 1, j - 1])`. The mask returned by `compute_mask` is a boolean array, and `numpy.isfinite` is `True` for both `True` and `False`, so the test is always true and no pairing is ever excluded.

Repro:

```python
import tslearn.metrics as M
s1 = [2.0, 0.0, 2.0, 0.0]
s2 = [0.0, 0.0, 1.0, 1.0]
print(M.lcss(s1, s2, eps=0))  # 0.5
print(M.lcss(s1, s2, eps=0, global_constraint="sakoe_chiba", sakoe_chiba_radius=1))  # 0.5, should be 0.25
```

The two matches that give `0.5` sit two steps off the diagonal, so a radius-1 Sakoe-Chiba band forbids one of them and the similarity should drop to `0.25` (confirmed with a brute-force LCSS restricted to the band).

The fix tests the boolean mask directly (`if mask[i - 1, j - 1]:`), which is the same convention already used by `_dtw.py` and the non-numba LCSS backend. Only the two score functions are changed. The two path-backtracking functions keep `isfinite` on purpose: their `while i > 0 and j > 0` loop only decrements inside the `if` block, so honoring the mask there could loop forever on a masked cell; constrained `lcss_path` is left as a separate follow-up.

Added a regression test that checks the unconstrained result is unchanged (`0.5`) and the radius-1 result is now `0.25`. Relates to #526.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Changelogs for this project are recorded in this file since v0.2.0.
 
 * soft-dtw API is now consistent with other metrics regarding parameters names and backend management.
 
+### Fixed
+
+* `lcss` and `lcss_path_from_metric` now respect the `global_constraint` / Sakoe-Chiba / Itakura band, which was previously ignored. ([#526](https://github.com/tslearn-team/tslearn/issues/526))
+
 ## [v0.9.0]
 
 ### Added
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -231,6 +231,22 @@ def test_lcss():
             np.testing.assert_equal(sim, 1.0)
 
 
+def test_lcss_sakoe_chiba_constraint():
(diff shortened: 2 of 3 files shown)


charavelg:
Hi @AlejandroCoronadoN and thanks for the clean PR, nice catch.

AlejandroCoronadoN:
Thanks, @charavelg! Glad it was a clean catch. Happy to keep contributing.

charavelg:
> Happy to keep contributing.

Any help will be welcomed!!!

Issue #510, opened 2024-02-28
Global alignment kernel returns NaN for all timeseries


NAThompson:
**Describe the bug**

`gak(x,y)` returns `nan` for all `x,y`.

**To Reproduce**

```python
import random
import numpy
from math import pi as π
from tslearn.metrics import sigma_gak, gak

def test_reproduce():
    f0 = 20e9
    period = 1/f0
    ω0 = 2*π*f0
    # works at length 405; dies at 410:
    waveforms = numpy.empty(shape=(5, 410))
    times = numpy.linspace(-period/2, period/2, waveforms.shape[1])
    for i in range(waveforms.shape[0]):
        φ = random.gauss(0.0, 0.5)
        k = 3 + random.uniform(-0.5, 0.5)
        waveforms[i, :] = 0.5*(numpy.tanh(k*ω0*times + φ) + 1) + random.uniform(-0.05, 0.05)

    typical_values = 0.5*(numpy.tanh(3*ω0*times) + 1)
    σ = sigma_gak(typical_values)
    scores = numpy.empty(waveforms.shape[0])
    for i in range(waveforms.shape[0]):
        scores[i] = gak(typical_values, waveforms[i, :], sigma=σ)

    print(scores)
    print(gak(waveforms[0, :], waveforms[0, :], sigma=σ))
```

**Expected behavior**

The computation should not return NaN; maybe it needs to be stabilized with the log-sum-exp method?

**Environment (please complete the following information):**
 - OS: ARM MacOS
 - tslearn version: 0.6.3

YannCabanes:
Hello @NAThompson,
Thank you for your issue.
Indeed, there is an overflow which can occur if the function `gak` is run on long time series which have very few variations.
We have: `gak(s1, s2) = unnormalized_gak(s1, s2) / sqrt(unnormalized_gak(s1, s1) * unnormalized_gak(s2, s2))`
For example, if `s1` is a constant time series a size `n`, the function `unnormalized_gak(s1, s1)` will return the number of possible warpings to match `s1` with itself, each warping being optimal. 
This number is equal to the number of paths in a grid of `n` rows and `n` columns to go from the top left cell to the bottom right cell when the only possible moves are going right, down or diagonally (right and down).
When `s1` is a constant time series, `unnormalized_gak(s1, s1) = inf` on my local computer when `n >= 406`.
Everything is working fine when `n <= 405`.
[...]

PR #715, opened 2026-08-22
[MRG] Fix GAK returning NaN for time series longer than ~405 samples (#450)

### LINKED ISSUE #450: Regression not working for arrays longer than 405 elements (overflow in matrix)
I am trying to do some regression on time series on different length.
But I'm having issues with lengths longer than 406 elements 

The error message is:
ValueError: Input X contains NaN.
SVR does not accept missing values encoded as NaN natively.

Function: "njit_gak"  has an overflow issue

**To Reproduce**
####################################################################
import numpy as np
from tslearn.utils import to_time_series_dataset
from tslearn.svm import TimeSeriesSVR
from numpy import random
from tslearn.preprocessing import TimeSeriesScalerMinMax

def fun():
    x = np.arange(500)   # Length of array here
    y = x**2*random.rand()/2000+ np.sin(x) + np.cos(x)

    return abs(y)

input = []
output = []
for reps in range(3):

    y = fun()
    input.append(list(y))
    output.append(np.sqrt(np.min(y)))

X = to_time_series_dataset(input)
X1 = TimeSeriesScalerMinMax().fit_transform(X)

clf = TimeSeriesSVR(C=1.0, kernel="gak")

y_reg = output
clf.fit(X1, y_reg)
###################################################################
Does anyone have the same issue?

Thank you very much!

samim-reza:
## Summary

`gak` and `cdist_gak` returned `NaN` for every pair of time series longer than
about 405 samples. This normalizes the Global Alignment Kernel in log space, so
the result stays accurate whatever the length of the inputs.

Fixes #450 (and the duplicates #510, #188 that were closed in its favour).

## Problem

The GAK recursion sums over every alignment path, so the unnormalized kernel
grows like the central Delannoy numbers, roughly $(3 + 2\sqrt{2})^{sz}$. It
therefore leaves the range of a 64-bit float at `sz ≈ 406`:

```python
import numpy as np
from tslearn.metrics import gak, cdist_gak

x = np.sin(2 * np.pi * np.linspace(0, 1, 500)).reshape(-1, 1)
gak(x, x, sigma=5.0)          # nan   -- and gak(x, x) is 1 by construction
cdist_gak(np.stack([x, x]))   # all nan
```

`gak` normalized by dividing those already-overflowed values, so
`inf / inf` turned every result into `NaN`. Because `TimeSeriesSVC`,
`TimeSeriesSVR` and `KernelKMeans` build their kernel matrix with
`cdist_gak`, they failed outright on long series:

```python
from tslearn.generators import random_walk_blobs
from tslearn.svm import TimeSeriesSVC

X, y = random_walk_blobs(n_ts_per_blob=5, sz=500, d=1, n_blobs=2, random_state=0)
TimeSeriesSVC(kernel="gak").fit(X, y)
# ValueError: Input X contains NaN.
```

That is the symptom originally reported in #188 and #450, and #514 raised the
threshold from ~204 to ~405 without removing it.

## Solution

The quantity users ask for is the *normalized* kernel, which is always in
`(0, 1]` and perfectly well defined at any length — only the intermediate
values overflow. So the normalization is now done in log space:

```
gak(x, y) = exp( log k(x, y) - ½·log k(x, x) - ½·log k(y, y) )
```

and `cdist_gak` normalizes its matrix the same way instead of multiplying by
diagonal matrices of `1/sqrt(k(x, x))`.

To obtain `log k(x, y)`, `_log_unnormalized_gak` runs the existing linear
recursion first and only re-runs a new log-space recursion (log-sum-exp over
the three predecessors) when the linear value has overflowed to `inf` or
underflowed to `0`. Series that already worked therefore keep both their
current values and their current speed; the extra cost is paid only in the
range that used to return `NaN`.

The log-space recursion is added through the same `__make_..._from_gram_matrix`
factory as the existing one, so the numba and PyTorch backends stay in sync.

`unnormalized_gak` is unchanged and still returns `inf` past ~405 samples —
the exact value really is outside float64 range — which its docstring and the
user guide now state explicitly.

## Testing

- `tests/test_metrics.py::test_gak_long_time_series` — new non-regression test:
  `gak(x, x) == 1`, `cdist_gak` finite with unit diagonal, and the
  one/two-dataset paths agree, at `sz=500`.
- `tests/test_svm.py::test_gak_svm_long_time_series` — new non-regression test
  covering the originally reported `TimeSeriesSVC` / `TimeSeriesSVR` failure.
- `tests/test_metrics.py::test_gak_log_and_linear_recursions_agree` — new test
  that the two recursions give the same answer where both are valid.

Both non-regression tests fail on `main` (`NaN`, and `ValueError: Input X
contains NaN`) and pass here.

Full suite on Linux / Python 3.12 with `all_features`, before and after:

```
main:  1063 passed, 31 skipped, 13 xfailed, 34 xpassed
here:  1070 passed, 31 skipped, 13 xfailed, 34 xpassed
```

(the difference is the 3 new tests plus 4 doctests of `tslearn/metrics/_gak.py`;
no other test changed status). `flake8` reports no new warnings on the touched
files. The PyTorch backend was exercised for `gak`, `unnormalized_gak` and both
`cdist_gak` paths, and its log-space recursion was checked against the numba one.

Timings for `gak` on two random walks (numpy backend, best of 5):

| `sz` | `main` | this PR |
|---|---|---|
| 300 | 6.0 ms | 5.2 ms |
| 600 | 34.7 ms | 56.3 ms |

`sz=300` uses the linear recursion, as before. `sz=600` pays for the log-space
pass, and previously returned `NaN`.

### DIFF
--- a/tslearn/metrics/_gak.py
+++ b/tslearn/metrics/_gak.py
@@ -166,13 +166,16 @@ def gak(s1, s2, sigma=1.0, be=None):
     s1 = to_time_series(s1, remove_nans=True, be=be)
     s2 = to_time_series(s2, remove_nans=True, be=be)
 
-    denom = be.sqrt(
-        _unnormalized_gak(s1, s1, sigma=sigma, backend=be)
-    ) * be.sqrt(
-        _unnormalized_gak(s2, s2, sigma=sigma, backend=be)
+    # Normalizing in log space keeps the ratio representable even when the
+    # unnormalized values themselves overflow (see issue #450).
+    log_denom = 0.5 * (
+        _log_unnormalized_gak(s1, s1, sigma=sigma, backend=be)
+        + _log_unnormalized_gak(s2, s2, sigma=sigma, backend=be)
     )
 
-    return _unnormalized_gak(s1, s2, sigma=sigma, backend=be) / denom
+    return be.exp(
+        _log_unnormalized_gak(s1, s2, sigma=sigma, backend=be) - log_denom
+    )
 
 
 def unnormalized_gak(s1, s2, sigma=1.0, be=None):
@@ -213,6 +216,14 @@ def unnormalized_gak(s1, s2, sigma=1.0, be=None):
     float
         Kernel value
 
+    Notes
+    -----
+    The unnormalized kernel sums over every possible alignment between the two
+    series, so it grows extremely fast with their length and leaves the range
+    of a 64-bit float for time series longer than about 405 samples, in which
+    case `inf` is returned. Use :func:`gak`, whose normalization is computed in
+    log space, when a value is needed for long time series.
+
     Examples
     --------
     >>> unnormalized_gak([1, 2, 3],
@@ -244,14 +255,44 @@ def unnormalized_gak(s1, s2, sigma=1.0, be=None):
     return _unnormalized_gak(s1, s2, sigma, be)
 
 
+def _log_gram_matrix(s1, s2, sigma, backend):
+    log_gram = -backend.cdist(s1, s2, "sqeuclidean") / (2 * sigma * sigma)
+    log_gram -= backend.log(2 - backend.exp(log_gram))
+    return log_gram
+
+
 def _unnormalized_gak(s1, s2, sigma, backend):
-    gram = -backend.cdist(s1, s2, "sqeuclidean") / (2 * sigma * sigma)
-    gram -= backend.log(2 - backend.exp(gram))
-    gram = backend.exp(gram)
+    gram = backend.exp(_log_gram_matrix(s1, s2, sigma, backend))
     if backend.is_numpy:
         return _njit_gak_from_gram_matrix(gram)
     else:
         return _gak_from_gram_matrix(gram)
+
+
+def _log_unnormalized_gak(s1, s2, sigma, backend):
+    """Compute the natural logarithm of the unnormalized GAK value.
+
+    The unnormalized kernel sums over every alignment path, so it grows like
+    the central Delannoy numbers (:math:`\\approx (3 + 2\\sqrt{2})^{sz}`) and
+    leaves the range of a 64-bit float for time series longer than about 405
+    samples. The recursion is therefore evaluated in linear space first, which
+    is cheaper, and only re-run in log space when that result is not usable.
+    """
+    log_gram = _log_gram_matrix(s1, s2, sigma, backend)
+    if backend.is_numpy:
+        gak_from_gram = _njit_gak_from_gram_matrix
+        log_gak_from_gram = _njit_log_gak_from_gram_matrix
+    else:
+        gak_from_gram = _gak_from_gram_matrix
+        log_gak_from_gram = _log_gak_from_gram_matrix
+
+    value = gak_from_gram(backend.exp(log_gram))
+    if 0.0 < value < backend.inf:
+        return backend.log(value)
+
+    # The value has overflowed to `inf` (or underflowed to 0): redo the
+    # recursion in log space, where it stays representable.
+    return log_gak_from_gram(log_gram)
 
 
 def __make_gak_from_gram_matrix(backend):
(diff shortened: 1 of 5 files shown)


tslearn/metrics/_gak.py
```diff
@@ -244,16 +255,46 @@ def unnormalized_gak(s1, s2, sigma=1.0, be=None):
...
+        gak_from_gram = _gak_from_gram_matrix
+        log_gak_from_gram = _log_gak_from_gram_matrix
+
+    value = gak_from_gram(backend.exp(log_gram))
+    if 0.0 < value < backend.inf:
+        return backend.log(value)
+
+    # The value has overflowed to `inf` (or underflowed to 0): redo the
+    # recursion in log space, where it stays representable.
+    return log_gak_from_gram(log_gram)
```
rtavenar, line 295: Hi!

Thanks for this PR, it is of great interest for our lib! I just have one naive question: why not directly perform the computation in log space, to save time? In other words, why do you need the call to `gak_from_gram` in the first place?

samim-reza: Hi @rtavenar, thanks for the quick look!

Not naive at all — it's a deliberate trade-off, but the numbers are the real answer.

The linear pass is there because the log-space recursion is **much more expensive per cell**: the inner loop trades two additions and a multiply for three `exp` and a `log`. Timing the numba kernels alone (square gram, best of 7):

| `sz` | linear | log-space | |
|---|---|---|---|
| 200 | 0.08 ms | 1.08 ms | 14x |
| 400 | 0.35 ms | 4.52 ms | 13x |
| 800 | 1.57 ms | 18.4 ms | 12x |

So it isn't a free swap: going straight to log space would slow down every user whose series are short enough to work today. End-to-end on `gak` (numpy backend, best of 5):

| `sz` | this PR | always log-space |
|---|---|---|
| 100 | 0.76 ms | 1.49 ms |
| 300 | 4.67 ms | 11.4 ms |
| 400 | 12.5 ms | 24.4 ms |
| 600 | 51.9 ms | 52.7 ms |
| 800 | 116 ms | 95 ms |

Below the threshold the linear pass is ~2x faster; above it you're right that the speculative pass is wasted, and it costs ~20%.

### A way to get both, if you want it

The overflow point depends only on the **lengths**, so it can be decided up front instead of by trial. Every gram entry is `u / (2 - u)` with `u = exp(-d / 2σ²) ∈ (0, 1]`, hence `≤ 1`, so the unnormalized value is bounded by the number of alignment paths — the Delannoy number `D(sz1-1, sz2-1)`.

That bound turns out to be tight exactly where users hit the bug:

```
log(DBL_MAX)      = 709.78
log D(404, 404)   = 708.59   -> sz = 405 cannot overflow
log D(405, 405)   = 710.35   -> sz = 406 can
```

which is precisely the 405/406 cutoff reported in #450 / #510 / #188. So `_log_unnormalized_gak` could pick the recursion from `sz1, sz2` in O(1) and never run a speculative pass — keeping the linear speed below the threshold and dropping the ~20% overhead above it.

Happy to push that, or to **drop the linear path entirely** if you'd rather keep the code simple and accept the ~2x on short series. Your call — just say which and I'll update.

Two smaller notes while I'm here:

- I checked the concern that a gram entry underflowing to `0` could make the linear pass silently drop paths: over 400 randomised cases where 4–13% of gram entries underflowed *and* the guard accepted the linear result, it agreed with the log-space value to 4e-16. A path through a cell with `gram < 5e-324` carries weight `≤ 5e-324`, so it only matters when essentially every path is killed — and then the total underflows to `0`, which the `0.0 < value` guard already rejects.
- The PyTorch log-space recursion is currently only verified locally, not in CI (a 500-sample pure-Python DP would be far too slow there). I can add a cheap test that exercises it directly on a small gram matrix against the numba one — happy to include it in whichever change you prefer.

charavelg: The current implementation in this PR seems legit: no overcost for existing working code is good IMHO.

charavelg:
@samim-reza, thanks for the contribution !!!

PR #717, opened 2026-08-26
Fix TypeError when be is passed to silhouette_score/silhouette_samples

### LINKED ISSUE #713: silhouette_samples/silhouette_score: metric_params collides with pinned arguments
## Problem

Passing `metric_params={"be": ...}` (or `**kwds` with `be`) to `tslearn.clustering.silhouette_samples` raises `TypeError: got multiple values for keyword argument 'be'`, because the kwds are merged into `metric_params_` and splatted into a call that already pins `be=be`.

## Suggested fix

Treat `be` like `n_jobs`/`verbose`: strip it from `metric_params_` (the backend is already pinned by `instantiate_backend(X)`), or explicitly reject it with a clear message. At minimum, document that `be` cannot be forwarded.

## Metadata

Severity: P2
Confidence: 100
Reviewer(s): adversarial (ce-code-review run 20260821-221628-4e205155 on PR #703)
Finding ID: tslearn/clustering/utils.py:347 metric_params={'be': ...} collides with the backend argument

Dark-Knight499:
Closes #713.

## Problem
`silhouette_score` / `silhouette_samples` pin `be=be` (backend fixed via `instantiate_backend(X)`) in their internal cross-distance calls, but kwargs merged from `metric_params`/`**kwds` were still forwarded — so `metric_params={"be": "numpy"}` raised:

```
TypeError: _cdist_dtw() got multiple values for keyword argument 'be'
```

## Fix
Treat `be` exactly like the already-handled `n_jobs`/`verbose`: strip it from `metric_params_` before the splat, in **both** functions. Also added `verbose` stripping to `silhouette_score`, which pins `verbose=verbose` in its calls but only stripped `n_jobs` — same collision class (`samples()` already stripped both).

## Verification
- Reproduced the exact `TypeError` on master for both functions before touching anything.
- After fix: both calls succeed; `silhouette_score(X, labels, metric="dtw", metric_params={"be": "numpy"})` equals `np.mean(silhouette_samples(...))` as expected.
- Added regression test `test_silhouette_pinned_kwargs_not_forwarded` covering: `be` via `metric_params`, and `be` via `**kwds`.
- Full `tests/test_clustering.py`: **10 passed** (includes all pre-existing silhouette tests unchanged).

### DIFF
--- a/tests/test_clustering.py
+++ b/tests/test_clustering.py
@@ -393,6 +393,10 @@ def test_silhouette_pinned_kwargs_not_forwarded():
                         rel_tol=1e-07)
     assert samples.shape == (12,)
     assert silhouette_score(X, labels, metric="dtw", be="numpy") is not None
+    # verbose is pinned by silhouette_score's call sites too; passing it
+    # through metric_params must be stripped, not forwarded
+    assert silhouette_score(X, labels, metric="dtw",
+                            metric_params={"verbose": 1}) is not None
 
 
 def test_silhouette_samples_softdtw_njobs_verbose(monkeypatch):
--- a/tests/test_clustering.py
+++ b/tests/test_clustering.py
@@ -375,6 +375,26 @@ def test_silhouette_samples():
     assert not np.allclose(constrained, unconstrained)
 
 
+def test_silhouette_pinned_kwargs_not_forwarded():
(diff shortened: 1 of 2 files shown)


charavelg:
Rejected in favor of #718 as it misses the documentation update

PR #729, opened 2026-09-09
Fix documented defaults that disagree with the signatures

VenishPaneliya:
Three documented defaults disagree with their signatures. Docs only, no behaviour change.

### `ctw_path` and `ctw` — `verbose`

Both are declared with `verbose=False`:

```python
def ctw_path(
    s1, s2, max_iter=100, n_components=None, global_constraint=None,
    sakoe_chiba_radius=None, itakura_max_slope=None,
    verbose=False,
    be=None,
):
```

but documented as:

```
    verbose : bool (default: True)
        If True, scores are printed at each iteration of the algorithm.
```

So the docs promise per-iteration score output from a plain `ctw(s1, s2)` call, which doesn't happen. `cdist_ctw`, in the same module, documents its own `verbose=0` correctly as `verbose : int, optional (default=0)` — so these two are outliers rather than a module-wide convention.

### `SoftDTWLossPyTorch.forward` — `global_constraint`

```python
def forward(ctx, D, gamma, global_constraint=None, sakoe_chiba_radius=None, itakura_max_slope=None):
```

```
    global_constraint : {0, 1, 2} (default: 0)
        Global constraint to restrict admissible paths for DTW:
        - "itakura" if 1
        - "sakoe_chiba" if 2
        - no constraint otherwise
```

Two problems: the default is `None`, not `0`, and `None` isn't a member of `{0, 1, 2}`, so the type was wrong too. The entry's own last line ("no constraint otherwise") plus the `sakoe_chiba_radius` text below it, which talks about `global_constraint` being unset, both describe `None` as the real default. I wrote it as `{0, 1, 2} or None (default: None)` to match the `... or None (default: None)` form used for `global_constraint` throughout `ctw.py` and `dtw_variants.py`.

Checked with `flake8`: identical output with and without this change (34 lines either way) — the warnings in `soft_dtw_loss_pytorch.py` are on untouched lines and are present on `main` too.

### DIFF
--- a/tslearn/metrics/ctw.py
+++ b/tslearn/metrics/ctw.py
@@ -99,7 +99,7 @@ def ctw_path(
         two. In this case, if `global_constraint` corresponds to no global
         constraint, a `RuntimeWarning` is raised and no global constraint is
         used.
-    verbose : bool (default: True)
+    verbose : bool (default: False)
         If True, scores are printed at each iteration of the algorithm.
     be : Backend object or string or None
         Backend. If `be` is an instance of the class `NumPyBackend` or the string `"numpy"`,
@@ -252,7 +252,7 @@ def ctw(
         two. In this case, if `global_constraint` corresponds to no global
         constraint, a `RuntimeWarning` is raised and no global constraint is
         used.
-    verbose : bool (default: True)
+    verbose : bool (default: False)
         If True, scores are printed at each iteration of the algorithm.
     be : Backend object or string or None
         Backend. If `be` is an instance of the class `NumPyBackend` or the string `"numpy"`,
--- a/tslearn/metrics/soft_dtw_loss_pytorch.py
+++ b/tslearn/metrics/soft_dtw_loss_pytorch.py
@@ -51,7 +51,7 @@ else:
             gamma : float
                 Regularization parameter.
                 Lower is less smoothed (closer to true DTW).
-            global_constraint : {0, 1, 2} (default: 0)
+            global_constraint : {0, 1, 2} or None (default: None)
                 Global constraint to restrict admissible paths for DTW:
                 - "itakura" if 1
                 - "sakoe_chiba" if 2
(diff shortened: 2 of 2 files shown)


charavelg:
Regarding _SoftDTWLossPyTorch.forward — global_constraint, I think the docstring is OK and the default value should be updated in the function parameters to 0 rather than None. 

`def forward(ctx, D, gamma, global_constraint=None, sakoe_chiba_radius=None, itakura_max_slope=None):`

_SoftDTWLossPyTorch is an internal API where global constraint historically is {0, 1, 2} (default: 0) (whereas public API tend to use {None, "sakoe_chiba", "itakura"})

VenishPaneliya:
Agreed — you're right, and I've pushed it that way: docstring restored to `{0, 1, 2} (default: 0)`, signature changed to `global_constraint=0`.

I checked the change is behaviour-preserving before making it, since it touches code rather than a docstring:

- `GLOBAL_CONSTRAINT_CODE` maps `{None: 0, "": 0, "itakura": 1, "sakoe_chiba": 2}`, so the public path already converts `None` to `0` before it reaches `forward`
- `compute_mask`, which this delegates to, itself defaults to `global_constraint=0`
- and directly: `_njit_compute_mask(6, 4, 0, None, None)` and `_njit_compute_mask(6, 4, None, None, None)` return identical all-True masks, so anyone calling `forward` directly with `None` today gets exactly what `0` now gives them

`tests/test_metrics.py -k "soft or dtw"` passes (13 tests), and flake8 output on the file is unchanged.

The two `ctw`/`ctw_path` `verbose` fixes in the first commit are untouched — those are the public API and genuinely default to `False`.

One thing I'm now unsure about: with the internal `{0, 1, 2}` convention in mind, is the `or None` in the *public* `SoftDTWLossPyTorch.__init__` docstring (`{"itakura", "sakoe_chiba"} or None`) the intended split — strings plus None at the public boundary, ints internally? If so I'll leave that side alone entirely.

charavelg:
{"itakura", "sakoe_chiba"} or None) is indeed the intended split for public API. Current PR is ok!

